PR #11, opened 2022-07-07
feat(libraries): add groups, multivar FF, continuity support for posthog-android

alexkim205:
Adds:

- [ ] Add groups support
- [ ] Add feature flags (new PostHogFeatureFlag.class)
- [ ] Add decide endpoints
- [ ] Add `send_feature_flags` option to capture
- [ ] Add persistence

Testing:

- [ ] All FF tests
- [ ] `send_feature_flags` with capture test
- [ ] E2E tests for sample apps

Not in scope
- [ ] Implementing override flags

### DIFF
--- a/posthog/src/main/java/com/posthog/android/PostHogFeatureFlags.java
+++ b/posthog/src/main/java/com/posthog/android/PostHogFeatureFlags.java
@@ -1,7 +1,7 @@
 package com.posthog.android;
 
 import static com.posthog.android.Persistence.ENABLED_FEATURE_FLAGS_KEY;
-import static com.posthog.android.internal.Utils.assertNotNull;
+import static com.posthog.android.internal.Utils.DEFAULT_FLAG_RELOAD_DEBOUNCE_INTERVAL;
 import static com.posthog.android.internal.Utils.closeQuietly;
 
 import android.support.annotation.NonNull;
@@ -18,7 +18,6 @@ import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.net.HttpURLConnection;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -115,7 +114,7 @@ public class PostHogFeatureFlags {
         if (this.reloadFeatureFlagsQueued && !this.reloadFeatureFlagsInAction) {
             new Thread(() -> {
                 try {
-                    Thread.sleep(5);
+                    Thread.sleep(DEFAULT_FLAG_RELOAD_DEBOUNCE_INTERVAL);
                     this.reloadFeatureFlagsQueued = false;
                     this.reloadFeatureFlagsRequest();
                 } catch (Exception e) {
@@ -125,20 +124,8 @@ public class PostHogFeatureFlags {
         }
     }
 
-    public Boolean isPostHogNull() {
-        return this.posthog == null;
-    }
-
-    public void putPostHog(PostHog posthog) {
-        this.posthog = assertNotNull(posthog, "posthog");
-    }
-
     private void setReloadingPaused(Boolean isPaused) {
         this.reloadFeatureFlagsInAction = isPaused;
-    }
-
-    private void resetRequestQueue() {
-        this.reloadFeatureFlagsQueued = false;
     }
 
     private void receivedFeatureFlags(HashMap response) {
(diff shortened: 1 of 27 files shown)


posthog/src/main/java/com/posthog/android/PostHogFeatureFlags.java
```diff
@@ -0,0 +1,228 @@
...
+        Persistence persistence = this.posthog.persistenceCache.get();
+
+        if (flags != null) {
+            persistence.putEnabledFeatureFlags(flags);
+            persistence.putActiveFeatureFlags(new ArrayList(flags.keySet()));
+        } else {
+            persistence.put(ACTIVE_FEATURE_FLAGS_KEY, null);
+            persistence.put(ENABLED_FEATURE_FLAGS_KEY, null);
+        }
+    }
```
benjackwhite, line 135: I'm a bit confused why we have "active" and "enabled" feature flags?

In this PR, `activeFeatureFlags` is unused as far as I can see and in the [iOS PR](https://github.com/PostHog/posthog-ios/pull/28/files#diff-e019e5b4b296ff0290e7a2a62105d48e408f9b5c62ea296954e3991269db6570R228) it is the opposite and `enabledFeatureFlags` is  unused.

As far as I understand the `enabled` idea only exists on the frontend where you can override the values to turn them off. If that is the funtionality we might want I would vote for removing one of these settings and maybe just stick with `activefeatureFlags` allowing us to later add `overriddenFeatueFlags` or something containing a map of any values that were explicitly overridden by the caller.

Quite possible I've misunderstood something so, so please let me know :D

posthog/src/main/java/com/posthog/android/PostHogFeatureFlags.java
```diff
@@ -0,0 +1,228 @@
...
+            this.reloadFeatureFlagsQueued = true;
+            this.startReloadTimer();
+        }
+    }
+
+    private void startReloadTimer() {
+        if (this.reloadFeatureFlagsQueued && !this.reloadFeatureFlagsInAction) {
+            new Thread(() -> {
+                try {
+                    Thread.sleep(5);
```
benjackwhite, line 106: Is 5 seconds the right default? and should we perhaps make this a config option

posthog/src/main/java/com/posthog/android/PostHogFeatureFlags.java
alexkim205: That makes sense! I would suggest that we keep `enabledFeatureFlags` over `activeFeatureFlags` since the former carries multivariate value data that is lost in `activeFeatureFlags`. Also I dont there'll be a need to also store `activeFeatureFlags` (when we end up implementing overridden flags) since it can be deduced from `enabledFeatureFlags` or via the `getFlags()` method.

posthog/src/main/java/com/posthog/android/PostHogFeatureFlags.java
alexkim205: for some context, this was copied over from the implementation in posthog-js: https://github.com/PostHog/posthog-js/blob/master/src/posthog-featureflags.js#L112

alexkim205: Looking at this again, I don't think this should be configurable option that is exposed to the user. We use it internally to debounce decide requests when polling feature flags. I'll abstract it to the Utils file though—

benjackwhite: I think I misunderstood what this does. It is just for debouncing multiple requests so 👍

posthog/src/main/java/com/posthog/android/PostHogFeatureFlags.java
```diff
@@ -0,0 +1,246 @@
...
+     * 3. Called once when building PostHogFeatureFlags instance (see build method)
+     */
+    public void reloadFeatureFlags() {
+        if (!this.reloadFeatureFlagsQueued) {
+            this.reloadFeatureFlagsQueued = true;
+            this.startReloadTimer();
+        }
+    }
+
+    private void startReloadTimer() {
```
benjackwhite, line 115: I find this method slightly misleading tbh. I thought it was responsible for periodically refreshing feature flags but its actually debouncing using a timer. I wonder if we couldn't call it `reloadFeatureFlagsWithDebounce` or something to clearly differentiate it.

benjackwhite (approved):
Overall looks great with one final comment of something that confused me but can be ignored if you disagree 😊

posthog/src/main/java/com/posthog/android/PostHogFeatureFlags.java
alexkim205: Thanks for the quick review here! It's a name I borrowed from posthog-js but this is so much better. Will change—

PR #13, opened 2022-07-22
Add session ID

liyiy:
checklist:

- [x] Every event sent from these clients should have a $session_id property as a uuid on it.
- [x] The same $session_id should be appended to all events until there's a 30 minute gap in activity. At which point, the $session_id should rotate
- [x] The 30 minute gap used above should be configurable. The name should be close to session_expiration_time_seconds (accounting for platform naming conventions)
- [x] A function should be exposed that resets the session_id (and expiration timer). On posthog-js, this is done with posthog.sessionManager.resetSessionId()
- [x] The $session_id should persist across app restarts (meaning, not stored in memory)
- [x ] Tested live on app.posthog

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,7 +44,7 @@ ext {
   versionCode = Integer.parseInt(VERSION_CODE)
   versionName = VERSION_NAME
 
-  minSdkVersion = 14
+  minSdkVersion = 26
   minSdkVersionWear = 21
   targetSdkVersion = 27
   compileSdkVersion = 27
--- a/posthog/src/main/java/com/posthog/android/PostHog.java
+++ b/posthog/src/main/java/com/posthog/android/PostHog.java
@@ -494,7 +494,8 @@ public class PostHog {
               // Add all feature flag keys to $active_feature_flags key
               finalProperties.putActiveFeatureFlags(activeFlags);
             }
-
+            getSessionId();
+            finalProperties.putSessionId(propertiesCache.get().sessionId());
             CapturePayload.Builder builder =
                 new CapturePayload.Builder().event(event).properties(finalProperties);
             fillAndEnqueue(builder, finalOptions);
@@ -546,7 +547,8 @@ public class PostHog {
             } else {
               finalProperties = properties;
             }
-
+            getSessionId();
+            finalProperties.putSessionId(propertiesCache.get().sessionId());
             //noinspection deprecation
             ScreenPayload.Builder builder =
                 new ScreenPayload.Builder()
@@ -790,7 +792,6 @@ public class PostHog {
     if (!isNullOrEmpty(distinctId)) {
       builder.distinctId(distinctId);
     }
-    getSessionId();
     enqueue(builder.build());
   }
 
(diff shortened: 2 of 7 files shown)


build.gradle
```diff
@@ -44,7 +44,7 @@ ext {
   versionCode = Integer.parseInt(VERSION_CODE)
   versionName = VERSION_NAME
 
-  minSdkVersion = 14
+  minSdkVersion = 26
```
liyiy, line 47: 🤔 is this gon be okay

EDsCODE: Why was this change needed?

posthog/src/main/java/com/posthog/android/PostHog.java
```diff
@@ -797,6 +803,18 @@ void enqueue(BasePayload payload) {
...
+  void getSessionId() {
+    Properties properties = propertiesCache.get();
+    Persistence persistence = persistenceCache.get();
+    String sessionId = properties.sessionId();
+    Instant sessionLastTimestamp = persistence.sessionLastTimestamp();
+    if (sessionId == null || (Duration.between(sessionLastTimestamp, Instant.now()).getSeconds() > this.sessionExpirationTimeSeconds)) {
+      String newSessionId = UUID.randomUUID().toString();
+      properties.putSessionId(newSessionId);
+    }
+    persistence.putSessionLastTimestamp(Instant.now());
```
EDsCODE, line 815: Does Properties persist the sessionId? I'm a bit unclear how this works

liyiy: bump the min sdk version to 26 to use java's time utils. We already made the change to use java 8 in the previous PR

liyiy: Yes there's a Properties and a Persistence for caching. I think properties is meant for values that get saved into properties that's why it holds sessionId but not sessionTimestamp

build.gradle
liyiy: i think it's fine? @alexkim205 wyt . Needed it for java time util

posthog/src/test/java/com/posthog/android/PostHogTest.java
```diff
@@ -397,6 +400,37 @@ protected boolean matchesSafely(GroupPayload payload) {
...
+  public void createsSessionIdIfNotSet() {
+    assertThat(properties.sessionId()).isNullOrEmpty();
+    assertThat(persistence.sessionLastTimestamp()).isNull();
+    posthog.capture("test");
+    assertThat(persistence.sessionLastTimestamp()).isInstanceOf(Instant.class);
+    assertThat(properties.sessionId()).isInstanceOf(String.class);
+  }
+
+  @Test
+  public void generatesNewSessionIdIfExpired() {
```
rcmarron, line 413: Should we add one more test that's the same as this but with a custom sessionExpiration time?

rcmarron: It's probably worth having a test that verified `properties.sessionId()` is assigned the value of the session id

rcmarron (commented):
Generally looks great! Just a few comments. The big one is that I think we shouldn't be changing the `minSDKVersion` - I'm not sure how big of a hassle that's going to be...

marandaneto:
Closed in favor of https://github.com/PostHog/posthog-android/pull/69

PR #17, opened 2023-01-18
Make sure defaultOptions are configured on capture calls

EDsCODE:
- Capture calls were incorrectly handling defaultOptions object
- This PR ensures that sending feature flag option will be handled correctly

### DIFF
--- a/posthog/src/main/java/com/posthog/android/PostHog.java
+++ b/posthog/src/main/java/com/posthog/android/PostHog.java
@@ -399,12 +399,7 @@ public class PostHog {
               finalOptions = options;
             }
 
-            final Properties finalUserProperties;
-            if (userProperties == null) {
-              finalUserProperties = EMPTY_PROPERTIES;
-            } else {
-              finalUserProperties = userProperties;
-            }
+            final Properties finalUserProperties = this.prepareEventProperties(userProperties, finalOptions)
 
             IdentifyPayload.Builder builder =
                 new IdentifyPayload.Builder().userProperties(finalUserProperties);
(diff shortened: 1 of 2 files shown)


posthog/src/test/java/com/posthog/android/PostHogTest.java
```diff
@@ -276,6 +276,32 @@ protected boolean matchesSafely(CapturePayload payload) {
                             }));
   }
 
+  @Test
+  public void captureWithSendFeatureFlagsByDefault() {
+    posthog.capture("capture with flags", new Properties().putValue("url", "github.com"));
+    verify(integration)
+            .capture(
```
neilkakkar, line 283: quick flyby that this should work for `screen()` calls too right?

neilkakkar (approved):
🚀

posthog/src/main/java/com/posthog/android/PostHog.java
```diff
@@ -754,6 +711,39 @@ public void reloadFeatureFlags() {
...
+    }
+
+    // Send feature flags with capture call
+    boolean shouldSendFeatureFlags = true;
+    if (
+            !options.context().isEmpty() &&
+                    options.context().get(SEND_FEATURE_FLAGS_KEY) instanceof Boolean &&
+                    (Boolean) options.context().get(SEND_FEATURE_FLAGS_KEY)
+    ) {
+      shouldSendFeatureFlags = (Boolean) options.context().get(SEND_FEATURE_FLAGS_KEY);
```
EDsCODE, line 729: Bug here: this line is never reached if key val is false

PR #20, opened 2023-02-02
Fix feature flags tests with missing distinct ID

liyiy:
Feature flag tests were failing because there was no distinct ID being set

### DIFF
--- a/posthog/src/main/java/com/posthog/android/PostHogFeatureFlags.java
+++ b/posthog/src/main/java/com/posthog/android/PostHogFeatureFlags.java
@@ -152,14 +152,15 @@ public class PostHogFeatureFlags {
             connection = client.decide();
             HttpURLConnection con = connection.connection;
 
-            if (properties.distinctId() == null) {
-                throw new IllegalArgumentException("Calling decide endpoint requires user to be identified before.");
-            }
 
             // Construct payload
             JSONObject payload = new JSONObject();
             payload.put("token", this.posthog.apiKey);
             payload.put("distinct_id", properties.distinctId());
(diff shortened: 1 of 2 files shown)


EDsCODE (approved):

neilkakkar:
there should be anon IDs though which are automatically generated? 🤔

EDsCODE:
Seems like the identify is triggering a refresh that otherwise isn't happening..

EDsCODE:
Here's the weird gotcha: https://github.com/PostHog/posthog-android/blob/master/posthog/src/main/java/com/posthog/android/PostHogFeatureFlags.java#L156

Decide not hit unless identify happens first. Should this be the pattern for posthog-android?

liyiy:
> there should be anon IDs though which are automatically generated? 🤔

There is an anon ID

liyiy:
> Here's the weird gotcha: https://github.com/PostHog/posthog-android/blob/master/posthog/src/main/java/com/posthog/android/PostHogFeatureFlags.java#L156
> 
> Decide not hit unless identify happens first. Should this be the pattern for posthog-android? Otherwise the fix would be to remove the line above and change it so that the decide payload will pass anonId as distinct_id if there is no distinct_id set

```remove the line above and change it so that the decide payload will pass anonId as distinct_id if there is no distinct_id set```

Right, just making sure ^ is the standard, idk what the standard is between all the libraries sometimes

neilkakkar:
Some apps don't have logins/ are privacy protected, where the implementors will never call `identify()`, so we should support flags without identify in android too :) - it's the common standard across all libs, great job catching this issue!

PR #22, opened 2023-02-06
Feature flags: decide endpoint resiliency and switch to v3

liyiy:


marandaneto:
We're currently reworking the Android SDK and it already uses the latest version of the decide endpoint (v3).

liyiy:
> We're currently reworking the Android SDK and it already uses the latest version of the decide endpoint (v3).

Awesome! Just making sure that that handles the flag resiliency work too. I’ll close this PR then

PR #24, opened 2023-02-22
chore: send feature flags by default

EDsCODE:
https://github.com/PostHog/posthog-android/pull/17 didn't work in all cases. Couldn't identify why but reenabling it again using as small of a change as possible


liyiy (approved):
🤞

PR #30, opened 2023-09-25
Next major Android SDK v3

marandaneto:


.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: |
+          echo "Tag name from github.ref_name: ${{ github.event.release.target_commitish }}"
+          ./scripts/commit-code.sh ${{ github.event.release.target_commitish }}
+
+      # Let's be sure that everything works before we release    
+      - name: Dry release
+        run: make dryRelease
+  
+      # - name: Release
+      #   run: make release
```
marandaneto, line 53: Remove comments when publishing the 1st alpha, this will upload artifacts to sonatype.

posthog-android/src/main/java/com/posthog/android/internal/PostHogAppInstallIntegration.kt
```diff
@@ -0,0 +1,46 @@
...
+                }
+                props["version"] = versionName
+                props["build"] = versionCode
+
+                preferences.setValue("version", versionName)
+                preferences.setValue("build", versionCode)
+
+                // TODO: do we need to send an event every time as an update? we need to compare the Ids maybe?
+                // maybe it didnt change
+                // fix me
```
marandaneto, line 41: TODO

posthog/src/main/java/com/posthog/PostHogConfig.kt
```diff
@@ -0,0 +1,50 @@
...
+    // apiKey and host are immutable due to offline caching
+    public val apiKey: String,
+    public val host: String = "https://app.posthog.com",
+    public var debug: Boolean = false,
+    @Volatile
+    public var optOut: Boolean = false,
+    public var sendFeatureFlagEvent: Boolean = true,
+    public var preloadFeatureFlags: Boolean = true,
+    // min. allowed is 1
+    public var flushAt: Int = 20, // TODO: remove this one
```
marandaneto, line 16: TODO

posthog/src/main/java/com/posthog/internal/PostHogFeatureFlags.kt
```diff
@@ -0,0 +1,115 @@
+package com.posthog.internal
+
+import com.posthog.PostHogConfig
+import java.util.concurrent.Executors
+import java.util.concurrent.atomic.AtomicBoolean
+
+internal class PostHogFeatureFlags(private val config: PostHogConfig, private val api: PostHogApi) {
+    // TODO: do we need the onFeatureFlags callback?
```
marandaneto, line 8: TODO

posthog/src/main/java/com/posthog/internal/SendCachedEventsIntegration.kt
```diff
@@ -0,0 +1,145 @@
...
+                return@execute
+            }
+
+            flushLegacyEvents()
+            flushEvents()
+        }
+        executor.shutdown()
+    }
+
+    // TODO: respect maxBatchSize
```
marandaneto, line 27: TODO

buildSrc/src/main/java/PostHogPublishConfig.kt
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+    organization {
+        name.set("PostHog")
+        url.set("https://posthog.com")
+    }
+    developers {
+        developer {
+            name.set("PostHog")
+            email.set("manoel@posthog.com")
```
marandaneto, line 86: Is there an Engineering email maybe?

.github/workflows/build.yml
```diff
@@ -0,0 +1,50 @@
+name: 'Build'
+on:
+  push:
+    branches:
+      - v3
```
marandaneto, line 5: TODO: remove once this gets merged

marandaneto:
@benjackwhite would love a 1st pass here, see install/setup instructions: https://github.com/PostHog/posthog-android/blob/v3/MIGRATION.md

marandaneto:
> @benjackwhite would love a 1st pass here, see install/setup instructions: [`v3`/MIGRATION.md](https://github.com/PostHog/posthog-android/blob/v3/MIGRATION.md?rgh-link-date=2023-09-26T14%3A15%3A19Z)

@pauldambra would you have some time to do a 1st pass?

posthog-android/src/test/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegrationTest.kt
```diff
@@ -0,0 +1,134 @@
...
+        assertEquals("ABCDEFG", fake.properties?.get("barcode"))
+        assertEquals("fasf", fake.properties?.get("Reference"))
+    }
+
+    @Test
+    fun `onActivityCreated captures deep link even if hierarchical url`() {
+        val url = "mailto:nobody@google.com"
+        val fake = executeDeepLinkTest(url)
+
+        assertEquals(url, fake.properties?.get("url"))
```
pauldambra, line 93: this makes me wonder if I could breakdown deeplinks by type "how many are mail vs web" for example.

Should we/do we capture that?

posthog/src/main/java/com/posthog/PostHogInterface.kt
```diff
@@ -0,0 +1,52 @@
+package com.posthog
+
+public interface PostHogInterface {
+    public fun <T : PostHogConfig> setup(config: T)
+
+    public fun close()
+
+    public fun capture(
+        event: String,
+        distinctId: String? = null, // TODO: should distinctId be here or just in identify?
```
pauldambra, line 10: * for the _java_ sdk. Yep, this should have distinctId
* for the _android_ sdk. I'd expect it not to have distinct id, but to read the current distinct id (based on whether identify or reset has been called)

pauldambra (commented):
a quick pass through with some thinking out loud... mostly to get all of the files marked as viewed so I can see other changes as they arrive

Lovely clear tests!

posthog-android/src/test/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegrationTest.kt
marandaneto: It'd capture because there's no validation for `Uri.parse(uri)` and we have no control over it, but the properties will be empty since `queryParameterNames` will not return anything useful.
I don't think we should validate the Uri since it's given by the client, so I'd keep it as it is.

posthog/src/main/java/com/posthog/internal/PostHogApi.kt
```diff
@@ -0,0 +1,76 @@
...
+    fun batch(events: List<PostHogEvent>) {
+        val batch = PostHogBatchEvent(config.apiKey, events)
+
+        val request = makeRequest("${config.host}/batch") {
+            batch.sentAt = Date()
+            serializer.serializeBatchApi(batch, it.bufferedWriter())
+        }
+
+        client.newCall(request).execute().use {
+            if (!it.isSuccessful) throw PostHogApiError(it.code, it.message, body = it.body)
```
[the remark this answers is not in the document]
marandaneto: Right now it does what the JS SDK does, if it's < 400, we do retry next time (with a backoff), if it's >= 400, we drop the event.

posthog/src/main/java/com/posthog/PostHogEvent.kt
```diff
@@ -0,0 +1,14 @@
...
+import java.util.Date
+import java.util.UUID
+
+public data class PostHogEvent(
+    val event: String,
+    @SerializedName("distinct_id")
+    val distinctId: String,
+    val properties: Map<String, Any>? = null,
+    val timestamp: Date = Date(),
+    val uuid: UUID = UUID.randomUUID(),
```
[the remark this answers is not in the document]
marandaneto: Right now it uses the standard UUID from the Java SDK, https://www.ietf.org/rfc/rfc4122.txt
I think it's v4 by default, IIRC v7 is still a draft, not approved, right?

posthog/src/main/java/com/posthog/PostHogInterface.kt
marandaneto: So I will keep it since it's an optional parameter, which behaves exactly like that, if it's given, it uses the given value, or it uses the cached value.
Kotlin has no overloads so I guess this is fine.

posthog/src/main/java/com/posthog/PostHogEvent.kt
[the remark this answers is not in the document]
pauldambra: yeah it's draft but yolo 🤣

posthog/src/main/java/com/posthog/PostHogInterface.kt
```diff
@@ -0,0 +1,52 @@
...
+        userPropertiesSetOnce: Map<String, Any>? = null,
+        groupProperties: Map<String, Any>? = null,
+    )
+
+    public fun identify(
+        distinctId: String,
+        properties: Map<String, Any>? = null,
+        userProperties: Map<String, Any>? = null,
+        userPropertiesSetOnce: Map<String, Any>? = null,
+        // TODO: should we have groupProperties here?
```
marandaneto, line 22: @pauldambra do you know about that? if the identify also accepts group properties? I found this mixed up in SDKs

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -0,0 +1,567 @@
...
+    }
+
+    // test: $merge_dangerously
+    public override fun capture(
+        event: String,
+        distinctId: String?,
+        properties: Map<String, Any>?,
+        userProperties: Map<String, Any>?,
+        userPropertiesSetOnce: Map<String, Any>?,
+        groupProperties: Map<String, Any>?,
```
marandaneto, line 224: I wonder if it'd be better to have a data class here that would have an optional ctor param for each property, so the method param. list is smaller, I am fine either way

posthog/src/main/java/com/posthog/internal/PostHogApi.kt
```diff
@@ -11,12 +11,14 @@ import java.io.IOException
 import java.io.OutputStream
 import java.util.Date
 
+/**
+ * The class that calls the PostHog API
+ * @property config the Config
```
pauldambra, line 15: I wouldn't have placeholder comments like these... 

I think the comments here should be to help the busy person who has just added PostHog and doesn't know what to do

"this is the class PostHog uses to blah, it shouldn't be used directly, if you want to wotsit, you need a clinkexpander instead"

(although _maybe_ not that exact text 🤣)

posthog/src/test/java/com/posthog/internal/PostHogApiTest.kt
```diff
@@ -0,0 +1,111 @@
...
+        mock.enqueue(response)
+        return mock
+    }
+
+    @Test
+    fun `batch returns successful response`() {
+        val http = mockHttp()
+        val url = http.url("/")
+
+        val sut = getSut(host = url.toString())
```
pauldambra, line 30: As a style thing I'd have

`val api = getPostHogAPI(bla)` here
or
`val sut = getPostHogAPI(bla)`

I _know_ it's the system under test from the context and find myself having to consciously translate it

It's only nit picking though

pauldambra (approved):
I'm team progress over perfection so if we can merge this and keep the release as Alpha that'd be my preference

I think then we need the

* here's where you look to get started
* here's where you look to migrate from v2 to v3

to be clear 

And we message in the community slack / twitter to ask for alpha testers

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -0,0 +1,648 @@
...
+
+    public override fun reset() {
+        if (!isEnabled()) {
+            return
+        }
+
+        memoryPreferences.clear()
+        // only remove properties, preserve BUILD and VERSION keys in order to to fix over-sending
+        // of 'Application Installed' events and under-sending of 'Application Updated' events
+        config?.cachePreferences?.clear(listOf("build", "version"))
```
pauldambra, line 481:
```suggestion
        config?.cachePreferences?.clear(except = listOf("build", "version"))
```

I'd keep the parameter named here, on first scan I didn't read the comment and assumed this was _only_ clearing build and version.

posthog/src/main/java/com/posthog/PostHogConfig.kt
```diff
@@ -0,0 +1,149 @@
...
+     * Defaults to false
+     */
+    public var debug: Boolean = false,
+
+    /**
+     * This flag prevents capturing any data if enabled
+     * Defaults to false
+     */
+    @Volatile
+    public var optOut: Boolean = false,
```
pauldambra, line 36: thinking myself into implementer mindset... "Can I change opt out or is it fixed on startup? So could I use this to enable a user aware tracking flow?"

posthog/src/main/java/com/posthog/PostHogConfig.kt
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * Send a $feature_flag_called event when a feature flag is used automatically
+     * Defaults to true
+     */
+    public var sendFeatureFlagEvent: Boolean = true,
```
pauldambra, line 42: imaginary user here too... why would I set this? Is there posthog documentation I could read

posthog/src/main/java/com/posthog/internal/PostHogApi.kt
marandaneto: Yeah totally agree with you but those classes are all internal, they are not visible to clients at all, so its mostly for us.

posthog/src/main/java/com/posthog/PostHogIntegration.kt
```diff
@@ -0,0 +1,17 @@
+package com.posthog
+
+/**
+ * Integration interface for capturing events automatically or adding plugins to the PostHog SDK
+ */
+public interface PostHogIntegration {
```
pauldambra, line 6: this is cool... can an SDK user add integrations too?

Should we have hooks in this interface that an integration has to consider?

posthog/src/main/java/com/posthog/PostHog.kt
marandaneto: Good point

posthog/src/main/java/com/posthog/PostHogConfig.kt
marandaneto: You can change during init or using `PostHog.optIn()`/`PostHog.optOut()`

posthog/src/main/java/com/posthog/PostHogIntegration.kt
marandaneto: Yep, they can add integrations as well that get automatically installed/uninstalled during the SDK's lifecycle.
All the events that are automatically captured are already integrations, its a plugin system.

posthog/src/main/java/com/posthog/PostHogEncryption.kt
```diff
@@ -0,0 +1,21 @@
+package com.posthog
+
+import java.io.InputStream
+import java.io.OutputStream
+
+/**
+ * Interface for encrypt and decrypt events, by default there's no encrypt since device is sandboxed
+ */
+public interface PostHogEncryption {
```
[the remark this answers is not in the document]
marandaneto: yes, will add a link to https://source.android.com/docs/security/app-sandbox as well

posthog/src/main/java/com/posthog/PostHogConfig.kt
marandaneto: I have not found any, just thru code snippets, eg https://github.com/PostHog/posthog.com/pull/6861

posthog/src/main/java/com/posthog/PostHogInterface.kt
pauldambra: I'd expect posthog-js to be the canonical implementation right now.

There when you call `posthog.group` we register super properties that are sent with every subsequent event. Confusingly (to me) events have groups but people don't.

Since Android library is stateful then I think it would work the same way.

I guess Java library will be stateless and so we'd either have to register super properties by distinct id or insist that groups are always passed by the caller

posthog/src/main/java/com/posthog/PostHogInterface.kt
```diff
@@ -0,0 +1,138 @@
...
+    )
+
+    /**
+     * Reloads the feature flags
+     * @param onFeatureFlags the callback to get notified once feature flags is ready to use
+     */
+    public fun reloadFeatureFlags(onFeatureFlags: PostHogOnFeatureFlags? = null)
```
pauldambra, line 55: The JS SDK interface has a nice feature to register callbacks that are called when feature flags change. https://github.com/PostHog/posthog-js/blob/master/src/posthog-core.ts#L1172

Might be a stretch to include it now but would be nice to have...

posthog/src/main/java/com/posthog/PostHogOnFeatureFlags.kt
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Interface for being notified once feature flags has been loaded
+ */
+public fun interface PostHogOnFeatureFlags {
+    /**
+     * The method that is called when feature flags are loaded
+     * @param featureFlags the Feature flags property
+     */
+    public fun notify(featureFlags: Map<String, Any>?)
```
pauldambra, line 11: Ah, it's here 🤣

pauldambra (commented):
I guess we need a companion PR here https://posthog.com/docs/libraries/android

posthog/src/main/java/com/posthog/PostHogConfig.kt
pauldambra: That's the kind of info I think fits really well in comments like this that will surface in the IDE when a user if poking around to understand the SDK.

It _might_ be better just in the documentation with a link out though so you're not editing in multiple places

posthog/src/main/java/com/posthog/internal/PostHogApi.kt
pauldambra: > so its mostly for us.
Yep, so I guess the useful comments then are when someone over confident like me thinks "Manoel is on holiday - I'll do a PR for this". What does the internal implementer really need to know here?

There's always the risk of a comment I genuinely saw in a previous job (pseudo code for easiness of typing)

```
// the person's name
// string
public getAge int
```

posthog/src/main/java/com/posthog/PostHogConfig.kt
pauldambra: @neilkakkar is sending feature flag events only for experiments?

Manoel has tricked me into nitpicking mode and I'm wondering what the comment should be here to help an implementer. Or if there's a canonical documentation link

posthog/src/main/java/com/posthog/PostHogIntegration.kt
pauldambra: That's really cool. Playing the role of stupid user - how do I find out more?

posthog/src/main/java/com/posthog/PostHogInterface.kt
marandaneto: thanks, removed the comment, I didn't find any code path for posthog-js

neilkakkar: Yeah this is right, `identify()` will never set group properties, even in stateless SDKs because the group properties have nothing to do with the person identity.

Only `posthog.group()` calls should have the parameter to set group properties :)

posthog/src/main/java/com/posthog/PostHogConfig.kt
marandaneto: Good idea, I will add to the dokka comment a well.

posthog/src/main/java/com/posthog/PostHogIntegration.kt
marandaneto: Right now there are no docs since it's a new approach, once we get closer to a stable version, I will add a new section to https://posthog.com/docs/libraries/android but would like to focus on the existing features for now.

Issue #26, opened 2023-04-11
Crash with recordScreenViews()


ferinagy:
Version used: 2.0.3

When enabling the screen view recording, we encountered the following crash:

```
FATAL EXCEPTION: main
Process: at.beeone.george.dev, PID: 2802
java.lang.IllegalArgumentException: name must be provided.
	at com.posthog.android.PostHog.screen(PostHog.java:523)
	at com.posthog.android.PostHog.screen(PostHog.java:501)
	at com.posthog.android.PostHog.recordScreenViews(PostHog.java:316)
	at com.posthog.android.PostHogActivityLifecycleCallbacks.onActivityStarted(PostHogActivityLifecycleCallbacks.java:106)
	at android.app.Application.dispatchActivityStarted(Application.java:421)
	at android.app.Activity.dispatchActivityStarted(Activity.java:1407)
	at android.app.Activity.onStart(Activity.java:1948)
	at androidx.fragment.app.FragmentActivity.onStart(FragmentActivity.java:345)
	at androidx.appcompat.app.AppCompatActivity.onStart(AppCompatActivity.java:251)
        ...
```
We have some activities that deliberately set `android:label=""` to avoid being read by screen reader that are probably causing this. The sdk should not crash. Either not tracking the event or tracking the activity class name instead would be better.

marandaneto:
We're currently reworking the Android SDK and this will be taken care of, will move this to WIP, so stay tuned.

marandaneto:
https://github.com/PostHog/posthog-android/pull/32 should address this use case.

PR #32, opened 2023-10-04
Does not report screen events if there's no title

marandaneto:
Closes https://github.com/PostHog/posthog-android/issues/26

## :pencil: Checklist

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
@@ -42,8 +42,8 @@ internal class PostHogActivityLifecycleCallbackIntegration(
     override fun onActivityStarted(activity: Activity) {
         if (config.captureScreenViews) {
             val activityLabel = activity.activityLabel(config)
-            activityLabel?.let {
-                PostHog.screen(it)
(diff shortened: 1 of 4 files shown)


posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
```diff
@@ -42,8 +42,8 @@ internal class PostHogActivityLifecycleCallbackIntegration(
     override fun onActivityStarted(activity: Activity) {
         if (config.captureScreenViews) {
             val activityLabel = activity.activityLabel(config)
-            activityLabel?.let {
-                PostHog.screen(it)
+            if (!activityLabel.isNullOrEmpty()) {
+                PostHog.screen(activityLabel)
```
pauldambra, line 46: I'd be tempted to report "Unlabelled activity"

But then I can imagine the next request would be "How do I turn off reporting unlabelled activities?" so 🤷

pauldambra (approved):

posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
marandaneto: Yeah, I'd say it's not useful because you won't know which screens they are anyway.

PR #38, opened 2023-10-05
Upsert flags when loading feature flags with computed errors

marandaneto:
https://github.com/PostHog/posthog-js-lite/blob/4a707fdc660785c187e0bfd99ec2df1887afcf5c/posthog-core/src/index.ts#L931
I did the wrong check.
Refactor for reading the JSON from resources instead of strings, its easier to maintain and read due to identation.

Unit tests

## :pencil: Checklist


posthog/src/test/java/com/posthog/internal/PostHogFeatureFlagsTest.kt
```diff
@@ -131,7 +134,7 @@ internal class PostHogFeatureFlagsTest {
     }
 
     @Test
-    fun `merge feature flags if no errors`() {
+    fun `merge feature flags if there are errors`() {
```
pauldambra, line 137: 🤯 😊

posthog/src/test/resources/json/basic-decide-no-errors.json
pauldambra, line 1: I'd find these two easier to think about if they were generated by a method. So I could see the difference is the payload and the `errorsWhileComputingFlags`

But it's stylistic and easy to change so non-blocking

pauldambra (approved):
👍

posthog/src/test/resources/json/basic-decide-no-errors.json
marandaneto: yeah, I agree, that's why I named the files this way, it's a bit more obvious but there's no syntax highlighting for JSONs represented as Strings so it's also not ideal.

PR #41, opened 2023-10-06
active_feature_flags event should filter non active flags

marandaneto:
## :bulb: Motivation and Context
Closes https://github.com/PostHog/posthog-android/issues/40

## :pencil: Checklist

- [X] I reviewed the submitted code.
- [X] I added tests to verify the changes.
- [ ] I updated the docs if needed.
- [X] No breaking change or entry added to the changelog.

PR #42, opened 2023-10-08
isFeatureEnabled now returns true if multivariant flag

marandaneto:
By doing https://github.com/PostHog/posthog-android/pull/41 I noticed that `isFeatureEnabled` is also wrong.

I also noticed that the iOS SDK never returns `false` if a flag has a `false` value.
https://github.com/PostHog/posthog-ios/blob/9298825fe26899a58b976d400254e0b050daa578/PostHog/Classes/PHGPostHog.m#L317
Is it a bug on the iOS SDK I believe, right?

## :pencil: Checklist

### DIFF
--- a/posthog/build.gradle.kts
+++ b/posthog/build.gradle.kts
@@ -25,7 +25,7 @@ configure<JavaPluginExtension> {
 
 buildConfig {
     useKotlinOutput()
-    packageName("com.posthog.internal")
+    packageName("com.posthog")
     buildConfigField("String", "VERSION_NAME", "\"${project.version}\"")
 }
 
--- a/posthog/src/main/java/com/posthog/internal/PostHogFeatureFlags.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogFeatureFlags.kt
@@ -51,8 +51,10 @@ internal class PostHogFeatureFlags(
                     synchronized(featureFlagsLock) {
                         if (response.errorsWhileComputingFlags) {
                             // if not all flags were computed, we upsert flags instead of replacing them
-                            this.featureFlags = (this.featureFlags ?: mapOf()) + (response.featureFlags ?: mapOf())
-                            this.featureFlagPayloads = (this.featureFlagPayloads ?: mapOf()) + (response.featureFlagPayloads ?: mapOf())
+                            this.featureFlags =
+                                (this.featureFlags ?: mapOf()) + (response.featureFlags ?: mapOf())
+                            this.featureFlagPayloads = (this.featureFlagPayloads
+                                ?: mapOf()) + (response.featureFlagPayloads ?: mapOf())
                         } else {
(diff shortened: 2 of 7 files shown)


posthog/src/main/java/com/posthog/internal/PostHogFeatureFlags.kt
```diff
@@ -51,8 +51,10 @@ internal class PostHogFeatureFlags(
                     synchronized(featureFlagsLock) {
                         if (response.errorsWhileComputingFlags) {
                             // if not all flags were computed, we upsert flags instead of replacing them
-                            this.featureFlags = (this.featureFlags ?: mapOf()) + (response.featureFlags ?: mapOf())
-                            this.featureFlagPayloads = (this.featureFlagPayloads ?: mapOf()) + (response.featureFlagPayloads ?: mapOf())
+                            this.featureFlags =
+                                (this.featureFlags ?: mapOf()) + (response.featureFlags ?: mapOf())
+                            this.featureFlagPayloads = (this.featureFlagPayloads
+                                ?: mapOf()) + (response.featureFlagPayloads ?: mapOf())
```
marandaneto, line 57: just identation

marandaneto:
Actually, I am a bit confused, the `decide` API does not return the disabled feature flags.

By disabled I mean this:
Screenshot 2023-10-08 at 08 42 54

so if the flag is within the response payload, it is enabled, always, then the [iOS](https://github.com/PostHog/posthog-ios/blob/9298825fe26899a58b976d400254e0b050daa578/PostHog/Classes/PHGPostHog.m#L317) implementation is correct.

Should `isFeatureEnabled` return `false` if the flag is within the payload but its value is `false`? on iOS this is never the case, it will always return `true`.
The older version of the [Android SDK](https://github.com/PostHog/posthog-android/blob/b8866efac427a08646c0f17c57c2b6c2df1b17ee/posthog/src/main/java/com/posthog/android/PostHogFeatureFlags.java#L90-L94) does what this PR is aiming to do.

@neilkakkar mind explaining?

neilkakkar:
Decide returns all enabled flags, correct.

But just because a flag is enabled doesn't mean that the API will always return true. For example, if you have a filter on Country = X on the flag, the API will return false if you're outside country X.

Same for, say, if the rollout % is not 100%.

neilkakkar:
> Should isFeatureEnabled return false if the flag is within the payload but its value is false?

Correct, and nice finding!

posthog/build.gradle.kts
```diff
@@ -25,7 +25,7 @@ configure<JavaPluginExtension> {
 
 buildConfig {
     useKotlinOutput()
-    packageName("com.posthog.internal")
```
marandaneto, line 28: To avoid https://github.com/pinterest/ktlint/issues/1874

benjackwhite (approved):

PR #47, opened 2023-10-11
SDK only sends the  event once per flag

marandaneto:
https://github.com/PostHog/posthog-ios-recorder/pull/6#discussion_r1354591017

## :pencil: Checklist

### DIFF
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -37,7 +37,7 @@ public class PostHog private constructor(
     private var featureFlags: PostHogFeatureFlags? = null
     private var queue: PostHogQueue? = null
     private var memoryPreferences = PostHogMemoryPreferences()
-    private val featureFlagsCalled = mutableSetOf<String>()
+    private val featureFlagsCalled = mutableMapOf<String, MutableList<Any?>>()
 
     public override fun <T : PostHogConfig> setup(config: T) {
         synchronized(lockSetup) {
@@ -455,10 +455,12 @@ public class PostHog private constructor(
 
         var shouldSendFeatureFlagEvent = true
         synchronized(featureFlagsCalledLock) {
-            if (featureFlagsCalled.contains(key)) {
+            val values = featureFlagsCalled[key] ?: mutableListOf()
+            if (values.contains(flag)) {
                 shouldSendFeatureFlagEvent = false
             } else {
-                featureFlagsCalled.add(key)
+                values.add(key)
+                featureFlagsCalled[key] = values
             }
         }
 
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -30,12 +30,14 @@ public class PostHog private constructor(
     private val lockSetup = Any()
     private val lockOptOut = Any()
     private val anonymousLock = Any()
+    private val featureFlagsSentLock = Any()
 
     private var config: PostHogConfig? = null
 
     private var featureFlags: PostHogFeatureFlags? = null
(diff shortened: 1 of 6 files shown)


marandaneto:
@neilkakkar I am tagging you on feature flags PR but feel free to just "this is ok and makes sense" rather than line by line.
Since SDKs do different things sometimes, its always good to have a quick look before either introducing a new behaviour or the wrong one =D

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -449,7 +453,16 @@ public class PostHog private constructor(
         }
         val flag = featureFlags?.getFeatureFlag(key, defaultValue) ?: defaultValue
 
-        if (config?.sendFeatureFlagEvent == true) {
+        var shouldSendFeatureFlagEvent = true
+        synchronized(featureFlagsSentLock) {
+            if (featureFlagsSent.contains(key)) {
+                shouldSendFeatureFlagEvent = false
+            } else {
+                featureFlagsSent.add(key)
```
marandaneto, line 461: One thing to consider is that this only tracks if the feature flag X has been called, but not really if the response was exactly the same.
Eg
`featureFlag: true`, but later its `featureFlag: false`, should we reset the `shouldSendFeatureFlagEvent` for this use case? (no other SDKs do this so I didn't do either).

neilkakkar: I'm pretty sure the other SDKs do, i.e. one event is sent per different response type

Example: https://github.com/PostHog/posthog-js/blob/master/src/posthog-featureflags.ts#L222 - depends on the flagReportedValue. Same is true for most server SDKs

neilkakkar (approved):
one comment to fix, otherwise LGTM!

posthog/src/main/java/com/posthog/PostHog.kt
marandaneto: @neilkakkar https://github.com/PostHog/posthog-js-lite/blob/9ac7090d676dc2a6e5688cbe2b0ec001e2a23088/posthog-core/src/index.ts#L987-L988
I looked at Android (old version), iOS (old and new one WIP) and js-lite, but all of them didn't track their value.

I wonder if that is really needed? at the end, what really matters is the feature flag usage or?
Also, we are not checking the `featureFlagPayload` of each flag anyway, so technically even the `js` SDK isn't doing 100% correctly.

The problem is that Mobile has limited memory, I don't think feature flags and their values alone are a problem, but it adds up.

marandaneto: I see, the `node` bits even consider the `distinctId` and have a `maxCacheSize` config but because its server side I guess, maybe it'd make sense if the user calls `identify` or `reset` though, also for frontend SDKs.

> What do you mean by checking the featureFlagPayload

My example was right, if the value of the feature flag is different, it'd send an event.
My example was supposed to be with the `featureFlagPayload`, eg.

Request 1 sends the called event:
```
{
  "featureFlags": {
    "4535-funnel-bar-viz": true
  },
  "featureFlagPayloads": {
    "4535-funnel-bar-viz": "aaa"
  }
}
```

Request 2 does not send the called event:

```
{
  "featureFlags": {
    "4535-funnel-bar-viz": true
  },
  "featureFlagPayloads": {
    "4535-funnel-bar-viz": "bbb"
  }
}
```

But the `featureFlagPayloads` payload has changed for the feature flag, I believe this is ok?

neilkakkar: yes indeed, payloads changing are ok

marandaneto: Alright, pushed a fix.

PR #49, opened 2023-10-12
`distinct_id` is always a top level property only

marandaneto:
No need to duplicate the `distinct_id` in the properties as well.
the Backend already checks for this field as top level or within the properties, but the top level is preferred.
_#skip-changelog_ since its just internal

## :pencil: Checklist


pauldambra (approved):
👍

Issue #50, opened 2023-10-17
`register` should cache properties in the disk


marandaneto:
### Description

See https://github.com/PostHog/posthog-ios-recorder/pull/8#discussion_r1361624547 for context

PR #51, opened 2023-10-17
Cache registered properties

marandaneto:
Closes https://github.com/PostHog/posthog-android/issues/50

## :pencil: Checklist

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/internal/PostHogSharedPreferences.kt
+++ b/posthog-android/src/main/java/com/posthog/android/internal/PostHogSharedPreferences.kt
@@ -5,6 +5,7 @@ import android.content.Context.MODE_PRIVATE
 import android.content.SharedPreferences
 import com.posthog.android.PostHogAndroidConfig
 import com.posthog.internal.PostHogPreferences
+import com.posthog.internal.PostHogPreferences.Companion.ALL_INTERNAL_KEYS
 import com.posthog.internal.PostHogPreferences.Companion.GROUPS
 
 /**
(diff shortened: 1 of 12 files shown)


marandaneto:
@pauldambra let me know if this change is ok, willing to a cut a `beta` this time :)

posthog-android/src/main/java/com/posthog/android/internal/PostHogSharedPreferences.kt
```diff
@@ -108,10 +118,35 @@ internal class PostHogSharedPreferences(
         }
     }
 
+    private fun addToStringifiedKeys(key: String, editor: SharedPreferences.Editor) {
+        val stringifiedKeys = mutableSetOf<String>()
```
pauldambra, line 122: in nit-picker mode I wonder if you could do this without allocating the mutable set since (pseudo code) really we just want `editor.putStringSet(STRINGIFIED_KEYS, getStringifiedKeys() + key)` 

but I don't remember Kotlin well enough to know if I'm just moving allocations around rather than reducing them so feel free to treat this as me thinking out loud and move on 😊

pauldambra (approved):
:lgtm:

posthog-android/src/main/java/com/posthog/android/internal/PostHogSharedPreferences.kt
marandaneto: The allocation is the same, but it is definitely less LoC, updated it to `getStringifiedKeys() + key`.

PR #52, opened 2023-10-19
Ship proguard rules for proguard/r8 full mode

marandaneto:
If proguard/R8 is enabled, the tooling can decide to remove classes that are called at runtime but are not referenced directly (eg via reflection), or rename fields that would break serialization that depends on reflection eg Gson.

## :pencil: Checklist


pauldambra (approved):
very low context :approval stamp: 😊

PR #53, opened 2023-10-19
In App Video Recording hackaton

marandaneto:
## :pencil: Checklist


posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
```diff
@@ -138,6 +139,7 @@ internal class PostHogQueue(
 
     private fun executeBatch() {
         if (!isConnected()) {
+            isFlushing.set(false)
```
marandaneto, line 142: I've found a bug here while dogfooding.

PR #60, opened 2023-11-08
Expose and allow to enable and disable the debug mode at runtime

marandaneto:
Unblock migration https://github.com/djamoapp/posthog-flutter/blob/ae8b77bb776e4d32882b2f61cdebf9fe45dbae34/lib/src/posthog.dart#L81-L84

## :pencil: Checklist

### DIFF
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -527,6 +527,13 @@ public class PostHog private constructor(
         return distinctId
     }
 
+    override fun debug(enable: Boolean) {
+        if (!isEnabled()) {
+            return
+        }
+        config?.debug = enable
+    }
+
     public companion object : PostHogInterface {
         private var shared: PostHogInterface = PostHog()
         private var defaultSharedInstance = shared
@@ -662,5 +669,8 @@ public class PostHog private constructor(
         }
 
         override fun distinctId(): String = shared.distinctId()
+        override fun debug(enable: Boolean) {
+            shared.debug(enable)
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -527,6 +527,13 @@ public class PostHog private constructor(
         return distinctId
     }
 
+    override fun debug(enable: Boolean) {
+        if (!isEnabled()) {
+            return
+        }
```
pauldambra, line 533: Ah, I think I see... we have the `isEnabled` check so that a developer gets feedback when they're trying to use the SDK incorrectly?

I guess over time it'd be nice to make that behaviour less repetitive in the code but total tangent and not blocking here

pauldambra (approved):
👍

posthog/src/main/java/com/posthog/PostHog.kt
marandaneto: The check is that you cannot flip any flag or set anything in case the SDK is disabled because at this point there's no active configuration.

So every method calls `isEnabled` first thing and early returns if the SDK was not enabled yet.

I don't know what you mean by less repetitive in the code, it's a guard check that has to be within every single public API.

pauldambra: A common C#/Java style for something like this is have an interface, an implementing class, and a wrapping class.

The wrapping class applies the check (or logs the info or whatever the repeated activity is) and then delegates through to the implementing class.

So, the implementing class doesn't have the repetition and the interface protects you from not having the check in the wrapping class. 

It's differently noisy though so I'm not saying we should do this now, but if this keeps growing then it might be less noisy to move the isEnabled check to a wrapper

marandaneto: Ahh now I got what you meant, 100%, yeah at some point this is a good idea, but not all the time since we have to watch out for bundle size, the compiler usually inlines stuff like that (I hope), but more classes/abstractions usually increase much more, will pay attention if I have to do it again.

PR #62, opened 2023-11-13
Pick up consumer proguard rules correctly

marandaneto:

> java.lang.RuntimeException: Failed to invoke constructor 'com.posthog.PostHogEvent()' with no args.

If Proguard/R8 removed the classes/methods/ctor not referenced directly.
https://github.com/google/gson/blob/2658aca66a1d0f097794a5ad5eb33c095a8f8d5d/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java#L273C39-L273C67

* Consumer file was not picked up correctly since it was set twice/and a typo.
* Issue reported by a user.

Running the sample with R8 enabled.

## :pencil: Checklist

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -17,15 +17,9 @@ org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8
 android.useAndroidX=true
 # Kotlin code style for this project: "official" or "obsolete":
 kotlin.code.style=official
-# Enables namespacing of each library's R class so that its R class includes only the
-# resources declared in the library itself and none from the library's dependencies,
-# thereby reducing the size of the R class for that library
-android.nonTransitiveRClass=true
 
 android.defaults.buildfeatures.buildconfig=true
-android.defaults.buildfeatures.renderscript=false
 android.defaults.buildfeatures.shaders=false
-android.defaults.buildfeatures.aidl=false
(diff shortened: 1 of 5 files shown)


gradle.properties
```diff
@@ -17,15 +17,9 @@ org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8
 android.useAndroidX=true
 # Kotlin code style for this project: "official" or "obsolete":
 kotlin.code.style=official
-# Enables namespacing of each library's R class so that its R class includes only the
```
marandaneto, line 20: Those are default values from >= AGP 8, so removing them

posthog-android/build.gradle.kts
```diff
@@ -32,7 +31,7 @@ android {
 
     buildTypes {
         release {
-            consumerProguardFiles("proguard-rules.pro")
+            consumerProguardFiles("consumer-rules.pro")
```
marandaneto, line 34: The file name didn't match, probably a typo since its correct in L24.

benjackwhite (approved):
Nothing looks wrong so assuming you know what you're doing 😅

marandaneto:
> Nothing looks wrong so assuming you know what you're doing 😅

Hopefully :D
I could not reproduce the issue locally but there was room for improvement anyway.

PR #63, opened 2023-11-14
Fix Release as a non-Draft and manually create a tag before release

marandaneto:
The version update was done via CI.
The tag was automatically created by the GH Release.
The tag was pointing to the wrong commit since the tag was created before the version update was done.
There was a script to fix the tag commit hash after the version update was done.
Because the tag commit hash changed, the GH release was moved to Draft and forced us to Publish it again.
After publishing it again, the Release process was triggered again, then it failed because the release already existed.

Now we create the tag manually (via a script), so the release is created with the right tag and this problem won't happen again.

_#skip-changelog_

This will be done for iOS as well, I've not seen that the CI was failing after the 2nd time :(

## :pencil: Checklist


pauldambra (approved):
👍

benjackwhite (approved):

PR #64, opened 2023-11-14
Properties sanitizer

marandaneto:
Some SDKs have a `sanitize_properties` method, others `property_denylist`.
This is similar to `sanitize_properties` but as a callback that is run before the event is cached or sent over the wire. 
The user can redact, remove, or add any properties.
The downside is that the user can mess up with properties that may cause a bug on the backend or something.

## :pencil: Checklist


benjackwhite (approved):
Personally I like it. I think sanitize properties is the right terminology (certainly as good as any other)

marandaneto:
> Personally I like it. I think sanitize properties is the right terminology (certainly as good as any other)

Alright, added tests and updated the usage markdown, will cut a version :)

marandaneto:
For context, there was a workaround before:
`PostHog.capture("testEvent", properties = mapOf("\$device_name" to "removed"))`

This would work for manually captured events, but it'd not work for automatically captured events nor events that don't contain the `properties` param eg `identify`.

PR #69, opened 2023-11-28
PoC Session Replay - Android with wireframes

marandaneto:
* Only works on Android API >= 26 (Android 8)

TODO:
- [x] ProGuard rules for RREvents
- [ ] Validate mutations (requires BE changes)
- [x] Network events
- [x] Logcat events
- [x] Validate touch clicks
- [x] Test and add support to more standard widgets (right now only button, text, divs...)
- [x] Test and add support to more styling
- [ ] Add support to custom mappers (from custom widget to wireframe)
- [x] Add support to masking
- [ ] Test Jetpack Compose
- [ ] Change docs for Mobile Sessions if different than [Web](https://posthog.com/docs/data/sessions)

Probably FE/BE work:
- [ ] If [Authorized Domains for Replay](https://app.posthog.com/settings/project-replay#replay-authorized-domains) is enabled, the flag always return false
Added a comment that this has to be disabled for now.
- [ ] Should we reuse the `Capture console logs` for Logcat?
Introduced an SDK config called `captureLogcat` for now, not resuing the feature flags for now.
- [ ] Should we reuse the `Capture network performance` for OkHttp requests?
Introduced an SDK config called `captureNetworkTelemetry` for now, not resuing the feature flags for now.
- [x] HTML/body is still not center aligning correctly, [Draft](https://github.com/PostHog/posthog/pull/19036)
- [ ] Filtering Mobile replays, [Draft](https://github.com/PostHog/posthog/pull/19005)

## :pencil: Checklist


buildSrc/src/main/java/PostHogPublishConfig.kt
```diff
@@ -105,7 +105,7 @@ fun MavenPublication.postHogConfig(projectName: String, properties: Map<String,
 fun SigningExtension.postHogConfig(variantName: String, publishingExtension: PublishingExtension) {
     val privateKey = System.getenv("GPG_PRIVATE_KEY")
     val password = System.getenv("GPG_PASSPHRASE")
-    isRequired = true
+    isRequired = false
```
marandaneto, line 108: Rollback, just for testing it locally

buildSrc/src/main/java/PosthogBuildConfig.kt
```diff
@@ -11,9 +11,7 @@ object PosthogBuildConfig {
 
     object Android {
         val COMPILE_SDK = 34
-
-        // when changing this, remember to check the net.sf.androidscents.signature
-        val MIN_SDK = 21
+        val MIN_SDK = 24 // TODO: addOnFrameMetricsAvailableListener
```
marandaneto, line 14: Try to use SDK checks and use 21 instead

posthog/src/main/java/com/posthog/internal/PostHogFeatureFlags.kt
```diff
@@ -65,6 +65,23 @@ internal class PostHogFeatureFlags(
...
+                            (response.sessionRecording as? Map<String, Any?>).let { sessionRecording ->
+                                config.sessionReplay = true && config.sessionReplay
+                                config.sessionReplaySampleRate = sessionRecording?.get("sampleRate") as? Double ?: config.sessionReplaySampleRate
+                                config.snapshotEndpoint = sessionRecording?.get("endpoint") as? String ?: config.snapshotEndpoint
+                                // TODO: implement session recording config
+                                // consoleLogRecordingEnabled
+                                // recorderVersion=v2
+                                // sampleRate
+                                // minimumDurationMilliseconds
+                                // linkedFlag
```
marandaneto, line 82: Check if we actually need them

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
```diff
@@ -103,7 +105,8 @@ internal class PostHogQueue(
     private fun canFlushBatch(): Boolean {
         if (pausedUntil?.after(dateProvider.currentDate()) == true) {
             config.logger.log("Queue is paused until $pausedUntil")
-            return false
+            // TODO: revert, just for testing
```
marandaneto, line 108: So I can test stuff faster instead of waiting the timer

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -211,7 +211,7 @@ public class PostHogReplayIntegration(
                 timestamp = timestamp,
             )
             events.add(event)
-            status.sentFullSnapshot = true
+//            status.sentFullSnapshot = true
```
marandaneto, line 214: Player only supports full snapshot for now, revert later

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -288,8 +288,12 @@ public class PostHogReplayIntegration(
         var type: String? = null
         // button inherits from textview
         if (view is TextView) {
-            // TODO: masking
-            text = view.text.toString()
+            text = if (!view.isNoCapture() && !config.maskAllInputs) {
+                // TODO: masking
+                view.text.toString()
+            } else {
+                "*".repeat(view.text.length)
```
pauldambra, line 294: one thing I wish we'd done on the web was limit the length of this replaced text... sometimes we send ourselves thousands of asterisks ... would be cool to be able to `*****<truncated>` after some length - like 200 characters or something

marandaneto: Not a bad idea, but if the bug is around the UI (text too big), the replay won't be useful since the original message is truncated.
Maybe `*****<truncated - 200 more chars>` :D

posthog/src/main/java/com/posthog/PostHogOkHttpInterceptor.kt
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        val requestMap = mapOf<String, Any>(
+            "name" to url,
+            "method" to method,
+            "status" to statusCode,
+            "timestamp" to end,
+            "duration" to (end - start),
+            "transferSize" to transferSize,
+        )
```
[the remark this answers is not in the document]
pauldambra: (will check)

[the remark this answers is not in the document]
pauldambra: I think it needs to be `responseStatus`

pauldambra (commented):
ha... accidentally left some comments pending here 🙃

posthog/src/main/java/com/posthog/PostHogOkHttpInterceptor.kt
```diff
@@ -0,0 +1,52 @@
...
+        val url = request.url.toString()
+        val method = request.method
+        val statusCode = response.code
+        val start = response.sentRequestAtMillis
+        val end = response.receivedResponseAtMillis
+        var transferSize = (response.body?.contentLength() ?: 0) + (request.body?.contentLength() ?: 0)
+
+        if (transferSize < 0) {
+            transferSize = 0
+        }
```
pauldambra, line 37: ooh, this is cool... `Response` has nullable `cacheResponse()` and `networkResponse()` so you can distinguish between cached and non-cached responses. Which would let you decide when to return `transferSize` = 0

We could have `responseFromCache` on the payload and defuckulate that in the UI too for more clarity. I don't want to pedantically stick to the spec just not vary for the fun of it :)

Glad that this has headers and body available for when we want to start letting people capture them... that's going to be 💸

posthog/src/main/java/com/posthog/PostHogOkHttpInterceptor.kt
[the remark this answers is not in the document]
marandaneto: Yeah, I believe this could be inferred from either a content type, mime type, or URL, but for now let's stick to `fetch`, which is not wrong, but very generic =D
I check the `response.cacheResponse` and set it to 0 if coming from response.

> If you will always be initiator type fetch - or almost always - then I'll default it in the UI/ingestion to save wasted bytes

Sounds good, since we don't have a "fetch/XHR" API, but everything is a request, so technically everything can be a `fetch` as well.
Maybe the filtering can be more clever when headers are there, so you can inspect content type.

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -236,6 +251,13 @@ public class PostHog private constructor(
             }
         }
 
+        synchronized(sessionLock) {
+            if (sessionId != sessionIdNone) {
+                props["\$session_id"] = sessionId.toString()
+                props["\$window_id"] = sessionId.toString()
```
marandaneto, line 257: @pauldambra do you know if we still need this? should I generate a new UUID or can I reuse the `session_id` for the `window_id`?

posthog/src/main/java/com/posthog/PostHogOkHttpInterceptor.kt
```diff
@@ -0,0 +1,63 @@
...
+                "duration" to (end - start),
+                // TODO: remove it if we don't need it (server defaults to fetch)
+                "initiatorType" to "fetch",
+                "entryType" to "resource",
+            ),
+        )
+        val requests = listOf(requestMap)
+        val payload = mapOf<String, Any>("requests" to requests)
+
+        listOf(RRPluginEvent("rrweb/network@1", payload, end)).capture()
```
marandaneto, line 61: @pauldambra I wonder if the UI relies on the plugin name? can I change `rrweb` to something else? same for logging.

posthog/src/main/java/com/posthog/PostHogOkHttpInterceptor.kt
```diff
@@ -0,0 +1,63 @@
...
+        requestMap.putAll(
+            mapOf(
+                "name" to url,
+                "method" to method,
+                "responseStatus" to statusCode,
+                "timestamp" to end,
+                "duration" to (end - start),
+                // TODO: remove it if we don't need it (server defaults to fetch)
+                "initiatorType" to "fetch",
+                "entryType" to "resource",
```
marandaneto, line 55: @pauldambra this is still required right? its not a default value yet in the FE and/or BE/

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -252,6 +274,11 @@ public class PostHog private constructor(
             props["\$groups"] = it
         }
 
+        // Replay needs distinct_id also in the props
```
marandaneto, line 277: Depends on https://github.com/PostHog/posthog/pull/18954

gradle.properties
```diff
@@ -23,4 +23,4 @@ android.defaults.buildfeatures.shaders=false
 android.defaults.buildfeatures.resvalues=false
 
 # Release
-versionName=3.0.0
+versionName=3.1.0
```
marandaneto, line 26: Just for testing

posthog/src/main/java/com/posthog/PostHogOkHttpInterceptor.kt
pauldambra: Yep, we should keep the name the same... we're asserting that the payload is parse-able as if it is that web plugin. To avoid having two sets of handling in the playback

posthog/src/main/java/com/posthog/PostHogOkHttpInterceptor.kt
pauldambra: Yeah, I think better to be explicit here.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -0,0 +1,706 @@
...
+            rawY
+        }
+    }
+
+    private fun View.isNoCapture(maskInput: Boolean): Boolean {
+        return (this.tag as? String)?.lowercase()?.contains("ph-no-capture") == true || maskInput
+    }
+
+    private fun String.mask(): String {
+        return "*".repeat(this.length)
```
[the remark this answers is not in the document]
marandaneto: yeah, we discussed this here https://github.com/PostHog/posthog-android/pull/69#discussion_r1409271116
Let's address that after being sure that the layout works as expected first, then we can make such improvements.

PR #72, opened 2023-12-19
chore: Add mutations

marandaneto:
chore: Add mutations

## :pencil: Checklist

### DIFF
--- a/posthog/src/main/java/com/posthog/internal/replay/RRAddedNode.kt
+++ b/posthog/src/main/java/com/posthog/internal/replay/RRAddedNode.kt
@@ -2,6 +2,7 @@ package com.posthog.internal.replay
 
 import com.posthog.PostHogInternal
 
+// Also used for UpdatedNode
 @PostHogInternal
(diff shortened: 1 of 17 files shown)


posthog/src/main/java/com/posthog/internal/replay/RRAddedNode.kt
```diff
@@ -2,6 +2,7 @@ package com.posthog.internal.replay
 
 import com.posthog.PostHogInternal
 
+// Also used for UpdatedNode
 @PostHogInternal
 public class RRAddedNode(
```
pauldambra, line 7: I went through the same journey in the transformer with `updates` being sent as additions.

Ended up renaming so that `adds` and `updates` are typed as `MobileNodeMutation` because it felt mismatched to need the same comment....

How about `RRMutatedNode` ??

marandaneto: Yep, that sounds good.

PR #73, opened 2024-01-03
flush forcefully sends events regardless the delay

marandaneto:
https://github.com/postHog/posthog-js-lite/ sends events regardless the timer if `flush` is called manually.

https://posthog.com/questions/flush-interval-seconds-not-working

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next
 
-- `flush` forcefully sends events regardless the delay ([#66](https://github.com/PostHog/posthog-android/pull/66))
+- `flush` forcefully sends events regardless the delay ([#73](https://github.com/PostHog/posthog-android/pull/73))
 
 ## 3.0.0 - 2023-12-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+- `flush` forcefully sends events regardless the delay ([#66](https://github.com/PostHog/posthog-android/pull/66))
+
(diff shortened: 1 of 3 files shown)


benjackwhite (commented):
I can't tell from the PR alone but I can at least clarify how it works elsewhere:

1. By default events are queued and are only flushed as a batch once a timer passes or the size threshold is hit.
2. However, a user can manually call the `flush` function to force whatever is in the buffer to be flushed.

If this code achieves that then 👍

pauldambra (approved):
🚢

marandaneto:
> I can't tell from the PR alone but I can at least clarify how it works elsewhere:
> 
> 1. By default events are queued and are only flushed as a batch once a timer passes or the size threshold is hit.
> 2. However, a user can manually call the `flush` function to force whatever is in the buffer to be flushed.
> 
> If this code achieves that then 👍

1. not really, since its an `and` operation, the timer passes `and` the threshold hits.
2. yes, this PR addresses this use case.

marandaneto:
> > I can't tell from the PR alone but I can at least clarify how it works elsewhere:
> > 
> > 1. By default events are queued and are only flushed as a batch once a timer passes or the size threshold is hit.
> > 2. However, a user can manually call the `flush` function to force whatever is in the buffer to be flushed.
> > 
> > If this code achieves that then 👍
> 
> 1. not really, since its an `and` operation, the timer passes `and` the threshold hits.
> 2. yes, this PR addresses this use case.

made new commits that address the 1. item, if this ever becomes a battery-draining issue, we can revert this part since its super simple.

PR #78, opened 2024-01-10
fix: Events Queue is only paused if there were errors

marandaneto:
The event Queue was paused even if there were no errors, so setting a lower `flushIntervalSeconds` did not have a lot of effects.

The `Do not report flag events if the flag has been reported with the same value` fix part of this PR was found after fixing the main issue and was necessary to be in the same PR to make the CI happy.

## :pencil: Checklist

### DIFF
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -497,7 +497,7 @@ public class PostHog private constructor(
             if (values.contains(value)) {
                 shouldSendFeatureFlagEvent = false
             } else {
-                values.add(key)
+                values.add(value)
                 featureFlagsCalled[key] = values
             }
         }
--- a/posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
@@ -257,8 +257,10 @@ internal class PostHogQueue(
     }
 
     private fun calculateDelay(retry: Boolean) {
-        val delay = if (retry) min(retryCount * retryDelaySeconds, maxRetryDelaySeconds) else config.flushIntervalSeconds
-        pausedUntil = config.dateProvider.addSecondsToCurrentDate(delay)
+        if (retry) {
+            val delay = min(retryCount * retryDelaySeconds, maxRetryDelaySeconds)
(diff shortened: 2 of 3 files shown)


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -497,7 +497,7 @@ public class PostHog private constructor(
             if (values.contains(value)) {
                 shouldSendFeatureFlagEvent = false
             } else {
-                values.add(key)
+                values.add(value)
```
marandaneto, line 500: `do not send feature flags called event twice` test already exists and it was failing correctly

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
```diff
@@ -257,8 +257,10 @@ internal class PostHogQueue(
     }
 
     private fun calculateDelay(retry: Boolean) {
-        val delay = if (retry) min(retryCount * retryDelaySeconds, maxRetryDelaySeconds) else config.flushIntervalSeconds
-        pausedUntil = config.dateProvider.addSecondsToCurrentDate(delay)
+        if (retry) {
+            val delay = min(retryCount * retryDelaySeconds, maxRetryDelaySeconds)
+            pausedUntil = config.dateProvider.addSecondsToCurrentDate(delay)
+        }
```
pauldambra, line 263: I'm assuming the implicit return of pausedUntil isn't swallowed by the `if` block

(it's so long since I wrote any kotlin 🙈)

pauldambra (approved):
🚢

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
marandaneto: do you mean:

```
if (!retry) return
val delay = min(retryCount * retryDelaySeconds, maxRetryDelaySeconds)
pausedUntil = config.dateProvider.addSecondsToCurrentDate(delay)
```

?

marandaneto: discussed offline

pauldambra: As discussed in Slack, I had missed this is a field and had completely parsed this incorrectly - so it did help me but not in the way I thought 🤣

PR #81, opened 2024-01-15
fix: Send wireframe children for updated wireframes - mutation

marandaneto:

When there are keyboard events (open/close), the view is redrawn and the wireframe has a new `height` due to making space for the keyboard hence the wireframe observed a mutation.
The mutation sends just the wireframe but not its children leading to all children being lost in the FE since its stateless.

https://us.posthog.com/project/41706/replay/5e3d9268-1d86-4d10-81dc-82239b885a40

## :pencil: Checklist

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -729,10 +729,11 @@ public class PostHogReplayIntegration(
             // we have to copy without the childWireframes, otherwise they all would be different
             // if one of the child is different, but we only wanna compare the parent
             val oldItem = oldMap[id]?.copy(childWireframes = null) ?: continue
-            val newItem = newMap[id]?.copy(childWireframes = null) ?: continue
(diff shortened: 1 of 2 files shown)


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -729,10 +729,11 @@ public class PostHogReplayIntegration(
             // we have to copy without the childWireframes, otherwise they all would be different
             // if one of the child is different, but we only wanna compare the parent
```
pauldambra, line 730: I guess the comment is out of date now... otherwise this saves me so much work 🙈

pauldambra (approved):

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
marandaneto: The comment is for the comparison, so it's still up to date.
Instead of comparing property per property, I compare the hashCode so I know "something" has changed.

PR #83, opened 2024-01-17
Do not capture console logs and network requests if session replay and session are not active

marandaneto:
If session replay was disabled or session not active, console logs and network requests were still captured since the check was done only early in the app lifecycle.

Running both with session replay and session active and not active

## :pencil: Checklist

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/internal/PostHogLogCatIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/internal/PostHogLogCatIntegration.kt
@@ -3,6 +3,7 @@ package com.posthog.android.replay.internal
 import android.annotation.SuppressLint
 import android.os.Build
 import com.posthog.PostHog
+import com.posthog.PostHogIntegration
 import com.posthog.android.PostHogAndroidConfig
 import com.posthog.internal.interruptSafely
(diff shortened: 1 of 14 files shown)


posthog-android/src/main/java/com/posthog/android/replay/internal/PostHogLogCatIntegration.kt
```diff
@@ -2,21 +2,26 @@ package com.posthog.android.replay.internal
...
+import com.posthog.PostHogIntegration
 import com.posthog.android.PostHogAndroidConfig
 import com.posthog.internal.interruptSafely
 import com.posthog.internal.replay.RRPluginEvent
 import com.posthog.internal.replay.capture
 import java.text.SimpleDateFormat
 import java.util.Locale
 
-internal class PostHogLogCatWatcher(private val config: PostHogAndroidConfig) {
+internal class PostHogLogCatIntegration(private val config: PostHogAndroidConfig) : PostHogIntegration {
```
marandaneto, line 14: the logcat integration now is totally decoupled from the app lifecycle and session replay integration, its installed but only captures something if the session is active and replay is enabled

pauldambra (approved):
:shipit:

PR #86, opened 2024-01-19
Set `Content-Length` header for `gzip` bodies because of more strict proxies

marandaneto:
Set `Content-Length` header for `gzip` bodies because of more strict proxies
https://posthog.com/questions/not-able-to-trigger-any-event

Testing using a self hosted version given by the user

## :pencil: Checklist


pauldambra (approved):
:shipit:

PR #98, opened 2024-02-22
roll back compile API to 33 to keep back compatibility

marandaneto:
       1.  Dependency 'androidx.lifecycle:lifecycle-livedata-core-ktx:2.7.0' requires libraries and applications that
           depend on it to compile against version 34 or later of the
           Android APIs.

           :app is currently compiled against android-33.

           Also, the maximum recommended compile SDK version for Android Gradle
           plugin 7.4.2 is 33.

           Recommended action: Update this project's version of the Android Gradle
           plugin to one that supports 34, then update this project to use
           compileSdkVerion of at least 34.

There were more like this, we cannot force people to use the latest version.

## :pencil: Checklist

### DIFF
--- a/buildSrc/src/main/java/PosthogBuildConfig.kt
+++ b/buildSrc/src/main/java/PosthogBuildConfig.kt
@@ -14,7 +14,7 @@ object PosthogBuildConfig {
     }
 
     object Android {
-        val COMPILE_SDK = 34
+        val COMPILE_SDK = 33
 
         // when changing this, remember to check the ANIMAL_SNIFFER_SDK_VERSION
         // Session Replay (addOnFrameMetricsAvailableListener requires API 26)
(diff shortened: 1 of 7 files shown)


buildSrc/src/main/java/PosthogBuildConfig.kt
```diff
@@ -47,7 +47,7 @@ object PosthogBuildConfig {
 
     object Dependencies {
         // runtime
-        val LIFECYCLE = "2.7.0"
+        val LIFECYCLE = "2.6.2"
```
marandaneto, line 50: 2.7.0 is compiled against API 34 which is incompatible with older versions, unless the final app also uses API 34

daibhin (approved):
Assuming we're still a long way off dropping support for 33?

marandaneto:
> Assuming we're still a long way off dropping support for 33?

Yep, the min. right now is 21

PR #101, opened 2024-02-27
do not allow empty or blank `distinct_id` or `anon_distinct_id`

marandaneto:
https://posthog.com/docs/product-analytics/identify#what-happens-if-you-call-identify-or-alias-with-invalid-inputs

A customer has events with empty `$anon_distinct_id` which is likely a side effect of calling `identify` with an empty `distinct_id`, adding some validations and loggings.

Just running it, tests are in place for non nullable fields at least

## :pencil: Checklist

### DIFF
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -420,6 +420,8 @@ public class PostHog private constructor(
         val anonymousId = this.anonymousId
         if (anonymousId.isNotBlank()) {
             props["\$anon_distinct_id"] = anonymousId
+        } else {
+            config?.logger?.log("identify called with invalid anonymousId: $anonymousId.")
         }
 
         capture(
@@ -434,6 +436,8 @@ public class PostHog private constructor(
             // We keep the AnonymousId to be used by decide calls and identify to link the previousId
             if (previousDistinctId.isNotBlank()) {
                 this.anonymousId = previousDistinctId
+            } else {
+                config?.logger?.log("identify called with invalid former distinctId: $previousDistinctId.")
             }
             this.distinctId = distinctId
 
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -207,6 +207,7 @@ public class PostHog private constructor(
         }
 
     private fun buildProperties(
+        distinctId: String,
         properties: Map<String, Any>?,
         userProperties: Map<String, Any>?,
         userPropertiesSetOnce: Map<String, Any>?,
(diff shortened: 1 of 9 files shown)


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -475,6 +494,14 @@ public class PostHog private constructor(
         @Suppress("UNCHECKED_CAST")
         val groups = getPreferences().getValue(GROUPS) as? Map<String, Any>
 
+        val distinctId = this.distinctId
+        val anonymousId = this.anonymousId
+
+        if (distinctId.isBlank() || anonymousId.isBlank()) {
```
marandaneto, line 500: @neilkakkar can you confirm that the decide API requires both fields non nullable non blank or just `distinctId` is required?

marandaneto: is the right approach here to not make the call at all or still let it happen?
if we don't make the call, should we clean the feature flags state from memory/disk?

pauldambra:
```
generate_exception_response(
                            "decide",
                            "Decide requires a distinct_id.",
                            code="missing_distinct_id",
                            type="validation_error",
                            status_code=status.HTTP_400_BAD_REQUEST,
                        ),
```

https://github.com/PostHog/posthog/blob/88e5815abddb1d12e481a82738ced06af3f862e4/posthog/api/decide.py#L162-L176

looks like if you send `disable_flags` with the decide call then you can not send distinct_id otherwise you have to send something that to python is not None.

(so just whitespace would be accepted and you'd hope not actually match anyone 🙈)

pauldambra (approved):
codewise looks good... interested to see what the correct behaviour is expected to be 🙈

posthog/src/main/java/com/posthog/PostHog.kt
neilkakkar: if you send `disable_flags` no flags are computed :P - the indent is probably hard to see, but disable_flags is for the entire flag calculation flow.

So, if you want flags in any case, we always need a distinctID. AnonID is optional

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -415,7 +434,11 @@ public class PostHog private constructor(
 
         if (previousDistinctId != distinctId) {
             // We keep the AnonymousId to be used by decide calls and identify to link the previousId
-            this.anonymousId = previousDistinctId
+            if (previousDistinctId.isNotBlank()) {
```
neilkakkar, line 437: bit confused, do we need these checks everywhere? Can we prevent this at a higher level, like not setting `this.distinctId` to be an invalid id ever?

Having this check here seems sus to me, since it's not really actionable by the user anymore

posthog/src/main/java/com/posthog/PostHog.kt
marandaneto: We don't have the `advanced_disable_feature_flags` option on the Mobile SDKs yet so I'll not consider it, will just make the `anonymousId` optional here and only request if there's a valid `distinctId` at least.

marandaneto: [`267f1a2` (#101)](https://github.com/PostHog/posthog-android/pull/101/commits/267f1a2e64be2473c8523fe2b4bfe9596b156d82)

neilkakkar: sounds good

posthog/src/main/java/com/posthog/PostHog.kt
neilkakkar: I see gotcha!

PR #107, opened 2024-03-01
fix do not allow nullable sessionId

marandaneto:
Fixes https://github.com/PostHog/posthog-flutter/issues/89

For some reason, the companion object const val (aka static in Java) isn't available when the `sessionId` is declared
The non companion object vars are created before the companion object ones so the `sessionId` was assigned with null instead of `sessionIdNone`, static code analyzer should have gotten that.

Bug introduced by #105

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 3.1.13 - 2024-03-01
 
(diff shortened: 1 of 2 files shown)


pauldambra (approved):
🚢

PR #110, opened 2024-03-05
chore: custom drawable converter

marandaneto:
Drawables using https://github.com/mikepenz/Android-Iconics/blob/develop/iconics-core/src/main/java/com/mikepenz/iconics/IconicsDrawable.kt won't be possible to convert to Bitmap automatically

```
            sessionReplayConfig.drawableConverter = PostHogDrawableConverter { drawable ->
                if (drawable is IconicsDrawable) {
                    drawable.toBitmap()
                }
                null
            }
```

## :pencil: Checklist

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogSessionReplayConfig.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogSessionReplayConfig.kt
@@ -24,4 +24,12 @@ public class PostHogSessionReplayConfig(
      */
     @PostHogExperimental
     public var captureLogcat: Boolean = true,
+
+    /**
+     * Converts custom Drawable to Bitmap
+     * By default PostHog tries to convert the Drawable to Bitmap, the supported types are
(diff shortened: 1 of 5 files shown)


posthog-android/src/main/java/com/posthog/android/replay/PostHogSessionReplayConfig.kt
```diff
@@ -24,4 +24,12 @@ public class PostHogSessionReplayConfig(
...
     @PostHogExperimental
     public var captureLogcat: Boolean = true,
+
+    /**
+     * Converts custom Drawable to Bitmap
+     * By default PostHog tries to convert the Drawable to Bitmap, the supported types are
+     * BitmapDrawable, ColorDrawable, GradientDrawable, InsetDrawable, LayerDrawable, RippleDrawable
+     */
+    @PostHogExperimental
+    public var drawableConverter: PostHogDrawableConverter? = null,
```
pauldambra, line 34: nice, so we do the best we can, but folk can poke in code when they're using something custom?

very cool!

pauldambra (approved):
🚀

PR #118, opened 2024-04-08
recording: multiple fixes for better frame rate, padding, drawables

marandaneto:
## :pencil: Checklist


posthog/src/main/java/com/posthog/internal/replay/RRStyle.kt
```diff
@@ -19,4 +19,6 @@ public data class RRStyle(
     public var paddingLeft: Int? = null,
     public var paddingRight: Int? = null,
     public var bar: String? = null,
+    public var iconLeft: String? = null,
+    public var iconRight: String? = null,
```
marandaneto, line 23: they hold a base64 value similar to `backgroundImage`, if present in text fields, they hold a small icon at the start (before the text), or at the end.

marandaneto: @pauldambra we don't have that, shall we add it?

> Rate and review has no icon

pauldambra: 🤔 seems good to add to me... 👍

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -556,14 +572,11 @@ public class PostHogReplayIntegration(
...
-                    base64 = if (convertedBitmap != null) {
-                        convertedBitmap.base64()
-                    } else {
-                        drawableCopy.base64(view.width, view.height)
-                    }
+                    base64 = drawable.cloneAndToBase64(view)
+//                    style.paddingTop = view.paddingTop.densityValue(displayMetrics.density)
+//                    style.paddingBottom = view.paddingBottom.densityValue(displayMetrics.density)
+//                    style.paddingLeft = view.paddingLeft.densityValue(displayMetrics.density)
+//                    style.paddingRight = view.paddingRight.densityValue(displayMetrics.density)
```
pauldambra, line 579: need to keep the commented code?

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -813,15 +855,15 @@ public class PostHogReplayIntegration(
...
 
-    private fun Drawable.copy(resources: Resources): Drawable? {
-        return constantState?.newDrawable(resources)
+    private fun Drawable.copy(): Drawable? {
+        return constantState?.newDrawable()
     }
 
     private fun String.mask(): String {
-        return "*".repeat(this.length)
+        return "*".repeat(length)
```
pauldambra, line 866: fly-by & non-blocking should we limit the max length of this?

posthog/src/main/java/com/posthog/internal/replay/RRStyle.kt
pauldambra: Ah, I'm catching up here maybe... is it a custom library 🤔 

I don't know how much we want to bake custom library stuff into the schema. 

Willing to be led by you - I've no idea how popular this library is

pauldambra: Shall we split things and get the padding fix in and then follow up with this icon change if we decide to do it?

pauldambra (approved):
👍

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
marandaneto: I don't think so otherwise if the label should break into the next line because its too big, it won't because we are truncating the string, so we should have the same amount of chars.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
marandaneto: yes, but it is a TODO (have to figure out the offset), the padding fixes some cases and breaks others because of the drawable size (vectors), the OS adjusts to the right size but we don't have access to this info, so in some cases the padding will fix it, in some cases it will break it.

posthog/src/main/java/com/posthog/internal/replay/RRStyle.kt
marandaneto: Its not a custom library, the icon is custom, but the drawables on the 4 possible positions are part of the UI system as well.
https://developer.android.com/reference/android/widget/TextView#getCompoundDrawables()

posthog-android/src/main/java/com/posthog/android/replay/internal/Debouncer.kt
```diff
@@ -31,6 +45,6 @@ internal class Debouncer(
     }
 
     companion object {
-        private const val ONE_FRAME_MS = 64L
+        private const val DELAY_MS = 500L
```
pauldambra, line 48: I can almost see us letting someone configure this... not-blocking here just seeding the idea in your head :)

posthog/src/main/java/com/posthog/internal/replay/RRStyle.kt
pauldambra: cool beans, I'll add this to schema/transformer 👍 

will need to figure out what the defaults/rules are around display but that's just mechanical

pauldambra (approved):
👍

posthog-android/src/main/java/com/posthog/android/replay/internal/Debouncer.kt
marandaneto: yeah, I thought the same, lets see if this approach works better first, and happy to make it configurable otherwise.

PR #125, opened 2024-04-24
fix: set correct User-Agent for Android and returns session recording even if authorized domains is enabled

marandaneto:
`User-Agent` was still posthog-java for Android

## :pencil: Checklist


pauldambra (approved):
:shipit:

PR #127, opened 2024-04-29
recording: add `screenshot` option for session replay instead of wireframe

marandaneto:
Adds a screenshot with high-fidelity UI instead of the wireframe option.
Adds more overhead to the app.
Enables Session replay for apps using Jetpack Compose.

The mutation updates seem to be buggy, waiting for @pauldambra so we can pair up on that, won't merge before this.

Running the sample and some open source apps

## :pencil: Checklist

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -116,7 +116,11 @@ public class PostHogReplayIntegration(
                             window.onDecorViewReady { decorView ->
                                 try {
                                     val listener =
-                                        decorView.onNextDraw(mainHandler, config.dateProvider) {
+                                        decorView.onNextDraw(
+                                            mainHandler,
+                                            config.dateProvider,
+                                            config.sessionReplayConfig.debouncerDelayMs,
+                                        ) {
                                             if (!isSessionReplayEnabled) {
                                                 return@onNextDraw
                                             }
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -460,7 +460,7 @@ public class PostHogReplayIntegration(
                 y = y,
                 width = width,
                 height = height,
-                type = "image",
+                type = "screenshot",
                 base64 = base64,
                 style = RRStyle(),
             )
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -305,7 +305,7 @@ public class PostHogReplayIntegration(
         val wireframe = view.toWireframe(window) ?: return
 
         // if the decorView has no backgroundColor, we use the theme color
-        if (wireframe.style?.backgroundColor == null) {
+        if (wireframe.style?.backgroundColor == null && !config.sessionReplayConfig.screenshot) {
             context.theme?.toRGBColor()?.let {
                 wireframe.style?.backgroundColor = it
             }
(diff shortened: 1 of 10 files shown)


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -399,6 +412,58 @@ public class PostHogReplayIntegration(
...
+                if (success) {
+                    base64 = bitmap.base64()
+                }
+            } catch (e: Throwable) {
+                config.logger.log("Session Replay PixelCopy failed: $e.")
+            } finally {
+                thread.quit()
+            }
+
+            return RRWireframe(
```
marandaneto, line 455: if something goes wrong, it means it's gonna be an empty image so the screen will be back blank/white (or placeholder).
the other option is to not send the mutation update at all so it will keep the last image in the replay, technically better but wrong?

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -399,6 +413,59 @@ public class PostHogReplayIntegration(
...
+            val latch = CountDownLatch(1)
+            val thread = HandlerThread("PostHogReplayScreenshot")
+            thread.start()
+
+            // unfortunately we cannot use the Looper.myLooper() because it will be null
+            val handler = Handler(thread.looper)
+
+            try {
+                var success = false
+                PixelCopy.request(window, bitmap, { copyResult ->
```
marandaneto, line 438: I've not found anything faster than that as discussed in the daily.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -399,6 +413,59 @@ public class PostHogReplayIntegration(
...
+                    if (copyResult == PixelCopy.SUCCESS) {
+                        success = true
+                    }
+                    latch.countDown()
+                }, handler)
+
+                // await for 1s max
+                latch.await(1000, TimeUnit.MILLISECONDS)
```
marandaneto, line 446: This is where the thread waits for the copying to finish, 1s is a lot but it should never take that long.

daibhin (approved):
Amazing how little code is involved! Clever to reuse the image from wireframe mode

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -399,6 +413,59 @@ public class PostHogReplayIntegration(
...
+
+                if (success) {
+                    base64 = bitmap.base64()
+                }
+            } catch (e: Throwable) {
+                config.logger.log("Session Replay PixelCopy failed: $e.")
```
pauldambra, line 452: we could add a custom event to the recording here so that we see when this is happening

(maybe with a debounce/throttle so we don't spam ourselves)

posthog-android/src/main/java/com/posthog/android/replay/PostHogSessionReplayConfig.kt
```diff
@@ -36,4 +38,12 @@ public class PostHogSessionReplayConfig(
...
     @PostHogExperimental
     public var screenshot: Boolean = false,
+    /**
+     * Deboucer delay used to reduce the number of snapshots captured and reduce performance impact
+     * This is used for capturing the view as a wireframe or screenshot
+     * The lower the number more snapshots will be captured but higher the performance impact
+     * Defaults to 500ms
+     */
+    @PostHogExperimental
+    public var debouncerDelayMs: Long = 500,
```
pauldambra, line 48: could nitpick and suggest we call this frames per second (so debounce delay of  500ms === an FPS of 2)

pauldambra: but no strong opinion

PR #129, opened 2024-05-02
chore: java sdk based on the android core sdk

marandaneto:

- Move away from https://github.com/postHog/posthog-java
- Start using the https://mvnrepository.com/artifact/com.posthog/posthog library
- Fix issues such as https://github.com/PostHog/posthog-java/issues/50, https://github.com/PostHog/posthog-java/issues/51 and https://github.com/PostHog/posthog-java/issues/53
- Support feature flags and payloads

Not super happy with the abstractions (PostHogFeatureFlagsInterface, PostHogFileInterface) but that was the easiest way to get something working fast and not doing breaking changes in the public APIs and changing the world, there's room for improvement though.

TODOs:

- [ ] Local evaluation (not on this PR - not for now, just using `decide`)
- [ ] Fix methods and classes comments (aka javadocs)
- [ ] Fix tests and add new ones for the new use cases
- [ ] Check and add method overloads for people using Java instead of Kotlin otherwise it will be super verbose (optional parameters in Kotlin are not optional in Java)
- [ ] Probably rename repo from `posthog-android` to `posthog-jvm` or `posthog-kotlin`, something more generic that will be a monorepo based on the Kotlin core.

## :pencil: Checklist

### DIFF
--- a/posthog-android/src/test/java/com/posthog/android/PostHogFake.kt
+++ b/posthog-android/src/test/java/com/posthog/android/PostHogFake.kt
@@ -23,7 +23,6 @@ public class PostHogFake : PostHogInterface {
         userProperties: Map<String, Any>?,
         userPropertiesSetOnce: Map<String, Any>?,
         groups: Map<String, Any>?,
-        groupProperties: Map<String, Any>?,
     ) {
         this.event = event
         this.properties = properties
@@ -65,6 +64,20 @@ public class PostHogFake : PostHogInterface {
         groups: Map<String, Any>?,
     ): Any? {
         return null
+    }
+
+    override fun getAllFeatureFlags(
+        distinctId: String?,
+        groups: Map<String, Any>?,
+    ): Map<String, Any>? {
+        return null
+    }
+
+    override fun getAllFeatureFlagsAndPayloads(
+        distinctId: String?,
+        groups: Map<String, Any>?,
+    ): Pair<Map<String, Any>?, Map<String, Any?>?> {
+        return Pair(null, null)
     }
 
     override fun flush() {
--- a/posthog-android/src/test/java/com/posthog/android/PostHogFake.kt
+++ b/posthog-android/src/test/java/com/posthog/android/PostHogFake.kt
@@ -22,6 +22,7 @@ public class PostHogFake : PostHogInterface {
         properties: Map<String, Any>?,
         userProperties: Map<String, Any>?,
         userPropertiesSetOnce: Map<String, Any>?,
+        groups: Map<String, Any>?,
         groupProperties: Map<String, Any>?,
     ) {
         this.event = event
@@ -43,6 +44,7 @@ public class PostHogFake : PostHogInterface {
         key: String,
         defaultValue: Boolean,
         distinctId: String?,
+        groups: Map<String, Any>?,
     ): Boolean {
         return false
     }
@@ -51,6 +53,7 @@ public class PostHogFake : PostHogInterface {
         key: String,
         defaultValue: Any?,
         distinctId: String?,
+        groups: Map<String, Any>?,
     ): Any? {
         return null
     }
@@ -59,6 +62,7 @@ public class PostHogFake : PostHogInterface {
         key: String,
         defaultValue: Any?,
         distinctId: String?,
+        groups: Map<String, Any>?,
     ): Any? {
         return null
     }
--- a/posthog-android/src/test/java/com/posthog/android/PostHogFake.kt
+++ b/posthog-android/src/test/java/com/posthog/android/PostHogFake.kt
@@ -42,6 +42,7 @@ public class PostHogFake : PostHogInterface {
     override fun isFeatureEnabled(
         key: String,
         defaultValue: Boolean,
+        distinctId: String?,
     ): Boolean {
         return false
     }
@@ -49,6 +50,7 @@ public class PostHogFake : PostHogInterface {
     override fun getFeatureFlag(
         key: String,
         defaultValue: Any?,
+        distinctId: String?,
     ): Any? {
         return null
     }
@@ -56,6 +58,7 @@ public class PostHogFake : PostHogInterface {
     override fun getFeatureFlagPayload(
         key: String,
         defaultValue: Any?,
+        distinctId: String?,
     ): Any? {
         return null
     }
@@ -76,17 +79,22 @@ public class PostHogFake : PostHogInterface {
         type: String,
         key: String,
         groupProperties: Map<String, Any>?,
+        distinctId: String?,
     ) {
     }
 
     override fun screen(
         screenTitle: String,
         properties: Map<String, Any>?,
+        distinctId: String?,
     ) {
         this.screenTitle = screenTitle
     }
 
-    override fun alias(alias: String) {
+    override fun alias(
+        alias: String,
+        distinctId: String?,
+    ) {
     }
 
     override fun isOptOut(): Boolean {
--- a/posthog/src/main/java/com/posthog/PostHogConfig.kt
+++ b/posthog/src/main/java/com/posthog/PostHogConfig.kt
@@ -95,7 +95,7 @@ public open class PostHogConfig(
      */
     public var propertiesSanitizer: PostHogPropertiesSanitizer? = null,
     public var featureFlagsRequestTimeoutSeconds: Int = 3,
-    public var batchRequestTimeoutSeconds: Int = 15,
+    public var requestTimeoutSeconds: Int = 15,
     public var disableGeoIP: Boolean = true,
 ) {
     // fix me: https://stackoverflow.com/questions/53866865/leaking-this-in-constructor-warning-should-apply-to-final-classes-as-well-as
@@ -140,16 +140,13 @@ public open class PostHogConfig(
     public var snapshotEndpoint: String = "/s/"
 
     @PostHogInternal
-    public var isAndroid: Boolean = false
+    public var isClientSDK: Boolean = false
 
     @PostHogInternal
     public var dateProvider: PostHogDateProvider = PostHogDeviceDateProvider()
 
     private val integrationsList: MutableList<PostHogIntegration> = mutableListOf()
     private val integrationLock = Any()
-
-    // TODO:
-    // disable_geoip
 
     /**
      * The integrations list
--- a/posthog/src/main/java/com/posthog/PostHogConfig.kt
+++ b/posthog/src/main/java/com/posthog/PostHogConfig.kt
@@ -137,6 +137,9 @@ public open class PostHogConfig(
     public var snapshotEndpoint: String = "/s/"
 
     @PostHogInternal
+    public var isAndroid: Boolean = false
+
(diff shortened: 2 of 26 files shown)


posthog/src/main/java/com/posthog/PostHogInterface.kt
```diff
@@ -21,15 +21,15 @@ public interface PostHogInterface {
...
+     * @param groups the group properties, set as a "$groups" property, Docs https://posthog.com/docs/product-analytics/group-analytics
      */
     public fun capture(
         event: String,
         distinctId: String? = null,
         properties: Map<String, Any>? = null,
         userProperties: Map<String, Any>? = null,
         userPropertiesSetOnce: Map<String, Any>? = null,
-        groupProperties: Map<String, Any>? = null,
+        groups: Map<String, Any>? = null,
```
marandaneto, line 32: That was a bug

posthog-android/src/test/java/com/posthog/android/PostHogFake.kt
```diff
@@ -22,7 +22,7 @@ public class PostHogFake : PostHogInterface {
         properties: Map<String, Any>?,
         userProperties: Map<String, Any>?,
         userPropertiesSetOnce: Map<String, Any>?,
-        groupProperties: Map<String, Any>?,
+        groups: Map<String, Any>?,
```
marandaneto, line 25: I wonder if this should have been `Map<String, String>` since `group(type: String, key: String)` are Strings only

posthog-android/src/test/java/com/posthog/android/PostHogFake.kt
```diff
@@ -42,24 +42,44 @@ public class PostHogFake : PostHogInterface {
...
+        distinctId: String?,
+        groups: Map<String, Any>?,
+    ): Map<String, Any>? {
+        return null
+    }
+
+    override fun getAllFeatureFlagsAndPayloads(
+        distinctId: String?,
+        groups: Map<String, Any>?,
+    ): Pair<Map<String, Any>?, Map<String, Any?>?> {
```
marandaneto, line 79: Not a big fan of `Pair`, but returning a map with 2 keys like in python:
`return {"featureFlags": None, "featureFlagPayloads": None}` isn't ideal either, people would need to know the key.

posthog-android/src/test/java/com/posthog/android/PostHogFake.kt
```diff
@@ -76,17 +96,22 @@ public class PostHogFake : PostHogInterface {
         type: String,
         key: String,
         groupProperties: Map<String, Any>?,
+        distinctId: String?,
```
marandaneto, line 99: `distinctId` is optional for Android since its cached in the disk/memory and required for Java, this has to be over documentation/method docs, otherwise I'd need to create another class and dupe a lot of code just because of this facility, is it worth it?

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
```diff
@@ -163,32 +174,33 @@ internal class PostHogQueue(
...
-                file.deleteSafely(config)
-                config.logger.log("File: ${file.name} failed to parse: $e.")
+            } else {
+                events.add(file.event(config))
             }
         }
 
         var deleteFiles = true
         try {
+            // TODO: python checks if the payload isnt bigger than 900kb
```
marandaneto, line 194: The tricky part here is that we serialize/stream from the disk directly to the network because of performance reasons (android specifically), so we don't load everything in memory.
if we wanna check the size before I'd need to change that and have one implementation per platform or pay the performance price on android.
Since we don't have this check yet on Android either (events are small anyways), I will keep as like this and implement whenever is needed

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
```diff
@@ -163,32 +174,33 @@ internal class PostHogQueue(
...
             if (events.isNotEmpty()) {
                 when (endpoint) {
                     PostHogApiEndpoint.BATCH -> api.batch(events)
                     PostHogApiEndpoint.SNAPSHOT -> api.snapshot(events)
                 }
             }
         } catch (e: PostHogApiError) {
+            // TODO: python retries up to 3 times with a 500ms interval
+            // if the API returns between 400 and 500
```
marandaneto, line 203: This probably has to be implemented in case we think its important to retry up to a few times even if the API returns an error.
Right now we drop the events since we prefer to not be shooting the API every time in case there's an issue with the backend.

posthog-android/src/test/java/com/posthog/android/PostHogFake.kt
pauldambra: "is it worth it"... 🤔 

it's a dev experience question I guess....

what does it mean for what I have to type?
what does it mean for how easy it is to make a mistake?
if i make a mistake how do I find out? will i get helpful errors

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
pauldambra: > events are small anyways

...wellllll until someone puts something massive in a property 🙈 	

posthog-js truncates properties to a configurable max length (and doesn't enforce a max number of properties) but doesn't otherwise check the size so it must be ok to let capture fail here...

posthog/src/main/java/com/posthog/PostHogConfig.kt
```diff
@@ -46,23 +46,23 @@ public open class PostHogConfig(
...
      * Defaults to 1000
      */
-    public var maxQueueSize: Int = 1000,
+    public var maxQueueSize: Int = 10000,
     /**
      * Number of maximum events in a batch call
      * Defaults to 50
      */
-    public var maxBatchSize: Int = 50,
+    public var maxBatchSize: Int = 100,
```
pauldambra, line 65: values and comments are out of sync 🙈

posthog/src/main/java/com/posthog/PostHogConfig.kt
```diff
@@ -136,6 +139,9 @@ public open class PostHogConfig(
     @PostHogInternal
     public var snapshotEndpoint: String = "/s/"
 
+    @PostHogInternal
+    public var isClientSDK: Boolean = false
```
pauldambra, line 143: since this is public config...

I don't actually know what client sdk means in this context...

(does it mean "android" or "java"? in which case is it better to be those strings?)

neilkakkar: To tack on top - is there a way to create a stateless + stateful core that inherits from stateless for the SDK code here? Similar to what we do in node.

It feels a bit awkward to have this interface where we add stateful bits based on `isClientSDK` in otherwise stateless functions. Makes it way too easy to introduce bugs + hard to reason about what can happen.

Alternatively, maybe we can extract the stateful parts out of the pure java SDK and into the android one? And have a clean interface for java that allows passing in whatever data you want, buuut doesn't depend on knowing whether it's client side or not.

posthog-android/src/test/java/com/posthog/android/PostHogFake.kt
marandaneto: If you call `capture` without `distinctId` on a Java backend app, it'll log in the console:

> PostHog.capture("event")
> vs
> PostHog.capture("event", distinctId = "123")

> "capture call not allowed, distinctId is invalid: $var."

I agree, I also prefer better dev UX but since both libs (core and Android) share the same interface, this is the issue we get hence my comment above.
If we prioritize better dev UX having an interface each lib we'd need to make major breaking changes, so we'd need to cut Android v4 as well, I was not aware or was an oversight from my side that `distinctId` has to be always passed during capture calls on Java since its stateless, this could be solved with scope management - thread local storage (if you set distinctId on thread/request X its always X, etc) but that's another big change not intended to be done now anyway.

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
marandaneto: https://github.com/PostHog/posthog-android/pull/130 at least reduces the batch size on the next retry so it's only gonna drop the event entirely if a single event is bigger than our limits

posthog/src/main/java/com/posthog/PostHogConfig.kt
marandaneto: > Fix methods and classes comments (aka javadocs)

yes, its part of the TODO list on the PR description already

posthog/src/main/java/com/posthog/PostHogConfig.kt
marandaneto: I understand the confusion since the JVM runs everywhere.
`client` means any app running on the user's device eg a Desktop app, an Android app, etc.
Sever means, well running on a server, this concept is known in the JVM world I guess unless I made it up years ago and kept it using lol (think of HTTP Server and HTTP Client), also its annotated with `@PostHogInternal` so no one should touch that besides us, with java docs this will be clearer.

I agree with the abstractions not being great, the goal here was a quick win as described in the PR description, not breaking the current interface so we don't have to cut Android v4 and offer a migration plan again. 
The reason is mainly because of [that](https://github.com/PostHog/posthog-android/pull/129#discussion_r1592489047).

miguelhrocha:
Hey @marandaneto!

I just noticed this PR 😅 and now I understand my proposals most likely won't get merged. However, I hope it helps for bringing the JVM version to life faster.

https://github.com/PostHog/posthog-java/pull/54

I think things like the POJOs, and the `FeatureFlagPoller` should be relevant enough in Kotlin world. One should be able to paste this files in IntelliJ and it would do the conversion to Kotlin.

Let me know if I can be of any further help!

marandaneto:
> Hey @marandaneto!
> 
> I just noticed this PR 😅 and now I understand my proposals most likely won't get merged. However, I hope it helps for bringing the JVM version to life faster.
> 
> [PostHog/posthog-java#54](https://github.com/PostHog/posthog-java/pull/54)
> 
> I think things like the POJOs, and the `FeatureFlagPoller` should be relevant enough in Kotlin world. One should be able to paste this files in IntelliJ and it would do the conversion to Kotlin.
> 
> Let me know if I can be of any further help!

Yep, most likely, the idea is to kill posthog-java and replace with the core parts of posthog-android since the Kotlin code is JVM compatible anyway and it has all the needed features but the feature flag poller.

True that, a lot of FeatureFlagPoller can be reused, thanks for that.

This is paused for now because, after the code review, we agreed that some breaking changes would need to be made to have a separate and friendly Java BE and Android FE interface, so there's no need for a few hacky solutions such as those mentioned in the comments above (public but internal fields, optional distinctId, etc), the goal is to get closer to the interfaces defined in the [Node/RN SDK](https://github.com/postHog/posthog-js-lite/) which is working well.

We might need a new major Android version very soon, so then we can do a single breaking change with proper migration steps at once, instead of 2 or more in such a short period, if that makes sense.

Thanks for the PR though.

PR #130, opened 2024-05-08
fix: reduce batch size if API returns 413

marandaneto:
Getting Flushing failed: PostHogApiError(statusCode=413, message=''). depending on the image size and batch size

Running and unit tests

## :pencil: Checklist

### DIFF
--- a/posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
@@ -5,7 +5,6 @@ import com.posthog.PostHogEvent
 import com.posthog.PostHogVisibleForTesting
 import java.io.File
 import java.io.IOException
-import java.lang.Math.floorDiv
 import java.util.Date
 import java.util.Timer
 import java.util.TimerTask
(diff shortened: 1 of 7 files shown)


posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
```diff
@@ -315,3 +315,26 @@ internal class PostHogQueue(
...
+): Boolean {
+    if (e.statusCode < 400) {
+        return false
+    }
+    // workaround due to png images exceed our max. limit in kafka
+    if (e.statusCode == 413 && config.maxBatchSize > 1) {
+        // try to reduce the batch size and flushAt until its 1
+        // and if it still throws 413 in the next retry, delete the files since we cannot handle anyway
+        config.maxBatchSize = calcFloor(config.maxBatchSize)
+        config.flushAt = calcFloor(config.flushAt)
```
pauldambra, line 335: nice... it makes me wonder how we'd know it was happening 🤔 but I guess we can figure out where we're returning 413 in capture and send an ingestion warning

pauldambra (approved):
👍

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
marandaneto: yeah we probably have 413s in our logging systems, adding one more custom event to the batch is likely to increase the payload :(

Issue #141, opened 2024-06-13
Use uuidv7 instead of uuidv4


marandaneto:
### Description

- [ ] Android
- [ ] iOS
- [X] [RN](https://github.com/PostHog/posthog-js-lite/blob/main/posthog-react-native/CHANGELOG.md#2112---2024-02-06)

After the migration, events from older versions of the SDKs won't work when using Session properties such as Session durations (Part of the new Mobile dashboard template).
Web analytics need to migrate from v4 to v7 to speed up a few things:

> @robbie-c It pretty much is a must, as it allows sampling by session ID, and even without sampling it's about 2x faster. I can't really release web analytics without this change

So we need to fit this within this sprint or latest the beginning of the next sprint before Web Analytics does the migration and breaks the current version of the SDKs

PR #142, opened 2024-06-14
chore: migrate UUID from v4 to v7

marandaneto:
Implement https://datatracker.ietf.org/doc/html/draft-ietf-uuidrev-rfc4122bis#section-5.7
Closes https://github.com/PostHog/posthog-android/issues/141 for Android

Considered
https://github.com/hfhbd/kotlinx-uuid/blob/main/kotlinx-uuid-core/src/commonMain/kotlin/kotlinx/uuid/UUID7.kt
https://github.com/f4b6a3/uuid-creator
https://github.com/cowtowncoder/java-uuid-generator

value example: 019016a2-45e0-7270-a583-a14f9a3e4af2
https://uuidgenerator.dev/uuid-validator says its a valid uuid v7

## :pencil: Checklist


posthog/src/test/java/com/posthog/vendor/uuid/UUIDTest.kt
```diff
@@ -0,0 +1,28 @@
...
+import kotlin.test.assertNotNull
+
+internal class UUIDTest {
+    @Test
+    fun `do not generate same value`() {
+        val uuid1 = TimeBasedEpochGenerator.generate()
+        val uuid2 = TimeBasedEpochGenerator.generate()
+        assertNotNull(uuid1)
+        assertNotNull(uuid2)
+        assertNotEquals(uuid1, uuid2)
```
pauldambra, line 16: really we should generate more than two 🙈 but I'm nit picking...

or at least I guess we should generate a large list and do three things

1) assert they're not too slow to generate - this is probably unnecessary 🤷 
2) assert they're unique
3) assert they sort in the order they're generated

it looks like a good source library so I'm not _Too_ bothered about us testing this here

pauldambra (approved):
one nit but :shipit:

posthog/src/test/java/com/posthog/vendor/uuid/UUIDTest.kt
marandaneto: 1. for speed there's already benchmarking https://www.baeldung.com/java-generating-time-based-uuids#3-benchmark-1 and the library has perf test as well https://github.com/cowtowncoder/java-uuid-generator/tree/master/src/test/java/perf
2 and 3 the library already does that https://github.com/cowtowncoder/java-uuid-generator/blob/4bd4b29c7ff61d7a839ee7b5f5981f8170b04621/src/test/java/com/fasterxml/uuid/UUIDComparatorTest.java#L111-L129 and there are more tests with 1m runs.
Since we are just vendoring, removing all the unnecessary bits from all the other UUID versions, and converting to Kotlin, I don't think we have to worry that much.

marandaneto: Added an adapted test with at least 10k units.
It tests the sorting and look for dupes.
https://github.com/PostHog/posthog-android/pull/142/commits/95aa074399643b2c54b3e7af3183c150e78da220

PR #153, opened 2024-08-07
fix: ensure activity name is used when activity label is not defined

### LINKED ISSUE #150: screen_name is not set on $screen events - Using compose
### Version

3.4.2

### Steps to Reproduce

1. Initialise posthog in the application onCreate:
```
val config = PostHogAndroidConfig(
    apiKey = BuildConfig.ANDROID_KEY_POSTHOG,
    host = "https://eu.i.posthog.com",
    captureScreenViews = true,
    captureApplicationLifecycleEvents = true,
).apply {
    sessionReplay = false
    debug = BuildConfig.DEBUG
}
PostHogAndroid.setup(this, config)
```
3. Start an Activity (AppCompatActivity) with startActivity(Intent)
4. Start more activities

### Expected Result

$screen events logged for each activity with `screen_name` being the `Activity::class.java.name` or something related to the activity opened

### Actual Result

$screen events logged for each activity with `screen_name` being the value from `rootProject.name = ...` in the root `settings.gradle.kts` file

I'm using multiple activities all with compose inside them.

Please let me know if more info is needed to reproduce

thisames:

The loadLabel method was loading the application label when an activity label was not defined. This change ensures that when an activity label is not defined, it defaults to the activity name instead of the application label. This improves the accuracy of screen name logging.

fix #150 

- Unit tests were updated and run to verify the behavior of the new activityLabelOrName method.
- Additional test scenarios were added to ensure that the correct labels are returned when activity and application labels are the same or empty.
- Verified that existing tests pass without any issues.

## :pencil: Checklist

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
@@ -51,7 +51,6 @@ internal class PostHogActivityLifecycleCallbackIntegration(
         }
     }
 
-
     override fun onActivityResumed(activity: Activity) {
     }
 
--- a/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
@@ -3,6 +3,7 @@ package com.posthog.android.internal
 import android.app.Activity
 import android.app.Application
 import android.app.Application.ActivityLifecycleCallbacks
+import android.content.pm.PackageManager.GET_META_DATA
 import android.os.Bundle
 import com.posthog.PostHog
 import com.posthog.PostHogIntegration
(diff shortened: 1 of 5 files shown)


posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
```diff
@@ -41,14 +41,16 @@ internal class PostHogActivityLifecycleCallbackIntegration(
         }
     }
 
-    override fun onActivityStarted(activity: Activity) {
-        if (config.captureScreenViews) {
-            val activityLabel = activity.activityLabel(config)
-            if (!activityLabel.isNullOrEmpty()) {
-                PostHog.screen(activityLabel)
+        override fun onActivityStarted(activity: Activity) {
```
marandaneto, line 44: identation is off here I think

marandaneto (approved):
Thanks @thisames

marandaneto:
it does not fix https://github.com/PostHog/posthog-android/issues/150 but addresses another issue.

thisames:
> @thisames would add a new item to the changelog file https://github.com/PostHog/posthog-android/blob/main/CHANGELOG.md ?

for sure =)

marandaneto (approved):

marandaneto:
@thisames type `make format` and commit to fix the lint issues

Issue #108, opened 2024-03-04
`identify` should not allow already identified users


marandaneto:
### Description

Relates to https://github.com/PostHog/posthog-js/pull/524

`const isKnownAnonymous = this.persistence.get_user_state() === 'anonymous'` is the check where we know if calling `identify` again is allowed or not.
If the user is already identified, calling `identify` again isn't allowed, if that's intentional, you should call `alias` instead.

If you want to identify another user again, you have to call `reset` first.
https://posthog.com/docs/product-analytics/identify#3-reset-after-logout

There are more restrictions described [here](https://posthog.com/docs/product-analytics/identify#alias-assigning-multiple-distinct-ids-to-the-same-user).

By not doing this, we might be merging users wrongly and we'll likely get a support request to split persons.

This has to be done on the other SDKs as well such as:
- [ ] iOS
- [ ] RN (js-lite)
- [X] Android

kingabnt:
> ### Description
> Relates to [PostHog/posthog-js#524](https://github.com/PostHog/posthog-js/pull/524)
> 
> `const isKnownAnonymous = this.persistence.get_user_state() === 'anonymous'` is the check where we know if calling `identify` again is allowed or not. If the user is already identified, calling `identify` again isn't allowed, if that's intentional, you should call `alias` instead.
> 
> If you want to identify another user again, you have to call `reset` first. https://posthog.com/docs/product-analytics/identify#3-reset-after-logout
> 
> There are more restrictions described [here](https://posthog.com/docs/product-analytics/identify#alias-assigning-multiple-distinct-ids-to-the-same-user).
> 
> By not doing this, we might be merging users wrongly and we'll likely get a support request to split persons.
> 
> This has to be done on the other SDKs as well such as:
> 
> * [ ]  iOS
> * [ ]  RN (js-lite)
> * [ ]  ...

PR #157, opened 2024-08-08
add `isIdentified` property and check on `identify()`

### LINKED ISSUE #136: Add `$is_identified`
### Description

https://github.com/PostHog/posthog/pull/22216

`$is_identified` is the property to filter identified users or not (anonymous).

https://github.com/PostHog/posthog-js/blob/7022ab69505959fb065d41870d2bceeb69145916/src/posthog-core.ts#L912
JS SDK uses the `$user_state` super property to track if the user has identified already or not with the values `identified` or `anonymous` (this super property is in deny list so not set in the event)

Add to:
- [ ] Android
- [ ] iOS
- [ ] RN

beradeep:
This change adds the `is_identified` property (absent earlier) to check user is identified or not (anonymous). This change also adds a fix to `identify()` to ensure that an "identify" event is captured if and only if the user isn't already identified (anonymous).  

fix #136 & #108

- Added unit test to assert against duplicate "identify" captures
- Added unit test to assert "identify" works as usual on reset
- Verified that this change doesn't break existing tests

## :pencil: Checklist

### DIFF
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -17,6 +17,7 @@ import com.posthog.internal.PostHogPrintLogger
 import com.posthog.internal.PostHogQueue
 import com.posthog.internal.PostHogSendCachedEventsIntegration
 import com.posthog.internal.PostHogSerializer
+import com.posthog.internal.PostHogSessionManager
 import com.posthog.internal.PostHogThreadFactory
 import com.posthog.vendor.uuid.TimeBasedEpochGenerator
 import java.util.UUID
@@ -48,13 +49,8 @@ public class PostHog private constructor(
     private val setupLock = Any()
     private val optOutLock = Any()
     private val anonymousLock = Any()
-    private val sessionLock = Any()
     private val groupsLock = Any()
 
-    // do not move to companion object, otherwise sessionId will be null
-    private val sessionIdNone = UUID(0, 0)
-
-    private var sessionId = sessionIdNone
     private val featureFlagsCalledLock = Any()
 
     private var config: PostHogConfig? = null
@@ -170,6 +166,10 @@ public class PostHog private constructor(
     public override fun close() {
         synchronized(setupLock) {
             try {
+                if (!isEnabled()) {
+                    return
+                }
+
                 enabled = false
 
                 config?.let { config ->
@@ -272,15 +272,13 @@ public class PostHog private constructor(
             }
         }
 
-        synchronized(sessionLock) {
-            if (sessionId != sessionIdNone) {
-                val sessionId = sessionId.toString()
-                props["\$session_id"] = sessionId
-                if (config?.sessionReplay == true) {
-                    // Session replay requires $window_id, so we set as the same as $session_id.
-                    // the backend might fallback to $session_id if $window_id is not present next.
-                    props["\$window_id"] = sessionId
-                }
+        PostHogSessionManager.getActiveSessionId()?.let { sessionId ->
+            val tempSessionId = sessionId.toString()
+            props["\$session_id"] = tempSessionId
+            if (config?.sessionReplay == true) {
+                // Session replay requires $window_id, so we set as the same as $session_id.
+                // the backend might fallback to $session_id if $window_id is not present next.
+                props["\$window_id"] = tempSessionId
             }
         }
 
@@ -697,25 +695,35 @@ public class PostHog private constructor(
     }
 
     override fun startSession() {
-        synchronized(sessionLock) {
-            if (sessionId == sessionIdNone) {
-                sessionId = TimeBasedEpochGenerator.generate()
-            }
+        if (!isEnabled()) {
+            return
         }
+
+        PostHogSessionManager.startSession()
     }
 
     override fun endSession() {
-        synchronized(sessionLock) {
-            sessionId = sessionIdNone
+        if (!isEnabled()) {
+            return
         }
+
+        PostHogSessionManager.endSession()
     }
 
     override fun isSessionActive(): Boolean {
-        var active: Boolean
-        synchronized(sessionLock) {
-            active = sessionId != sessionIdNone
+        if (!isEnabled()) {
+            return false
         }
-        return active
+
+        return PostHogSessionManager.isSessionActive()
+    }
+
+    override fun getSessionId(): UUID? {
+        if (!isEnabled()) {
+            return null
+        }
+
+        return PostHogSessionManager.getActiveSessionId()
     }
 
     override fun <T : PostHogConfig> getConfig(): T? {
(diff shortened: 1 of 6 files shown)


posthog/src/main/java/com/posthog/PostHogInterface.kt
```diff
@@ -164,6 +164,16 @@ public interface PostHogInterface {
...
 
+    /**
+     * Returns the [anonymousId] property
+     */
+    public fun anonymousId(): String
+
+    /**
+     * Returns the [isIdentified] property
+     */
+    public fun isIdentified(): Boolean
```
marandaneto, line 175: @beradeep this should be private since its an implementation detail, no need to expose the `isIdentified` method

posthog/src/main/java/com/posthog/PostHogInterface.kt
```diff
@@ -164,6 +164,16 @@ public interface PostHogInterface {
      */
     public fun distinctId(): String
 
+    /**
+     * Returns the [anonymousId] property
+     */
+    public fun anonymousId(): String
```
marandaneto, line 170: This is also implementation detail since users don't know about `anonymousId`, but only about `distinctId`.
It should not be part of the public API

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -227,6 +242,23 @@ public class PostHog private constructor(
...
 
+    private var isIdentified: Boolean
+        @JvmName("get-isIdentified")
+        get() {
+            var isIdentified: Boolean?
+            synchronized(identifiedLock) {
+                isIdentified = getPreferences().getValue(IS_IDENTIFIED) as? Boolean
+                if (isIdentified == null) {
+                    isIdentified = this.distinctId != this.anonymousId
+                    this.isIdentified = isIdentified as Boolean
```
marandaneto, line 253: Since `reset` will delete the preferences, we can assume its `false` here, if there's a value its always the value itself (always true in the case)

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -227,6 +242,23 @@ public class PostHog private constructor(
             getPreferences().setValue(DISTINCT_ID, value)
         }
 
+    private var isIdentified: Boolean
+        @JvmName("get-isIdentified")
+        get() {
+            var isIdentified: Boolean?
+            synchronized(identifiedLock) {
+                isIdentified = getPreferences().getValue(IS_IDENTIFIED) as? Boolean
```
marandaneto, line 250: We can probably read this during the SDK init instead so we do it only once, and after use a in-memory flag to avoid reading from preferences every time.

posthog/src/main/java/com/posthog/PostHogInterface.kt
beradeep: but @marandaneto the `PostHogInterface` is a public interface, and Kotlin doesn't allow non-public interface members.

beradeep: Ah, okay 👍

posthog/src/main/java/com/posthog/PostHog.kt
beradeep: but what will happen for already identified users on the current sdk? if we assume it be `false` won't it return `false` even though they're identified.

posthog/src/main/java/com/posthog/PostHog.kt
beradeep: Yes I think we can do that

posthog/src/main/java/com/posthog/PostHog.kt
marandaneto: Good point, for back compatibility, this makes a lot of sense.
We should change it then to, if there's a value, use the value, if there's no value, we do the `this.distinctId != this.anonymousId` check, so we can eventually remove this fallback once the new SDK is widely adopted.

beradeep: I agree. I think I've done exactly that 😅

posthog/src/main/java/com/posthog/PostHog.kt
beradeep: Also, I don't think we need to do this, as `SharedPreferences` itself caches the values in memory using a Map.

marandaneto: The `PostHogMemoryPreferences` does but the `PostHogSharedPreferences` does not really, indeed the Android `SharedPreferences`  does it but they have their internal locking which we don't have control so it's safer to have our flag I think, and it won't hurt anyway.

beradeep: I saw that the `PostHogSharedPreferences` is using Android's `SharedPreferences` under the hood, and what I've understood is that when the `setup` is done in the `Application.onCreate()`, the `PostHogSharedPreferences` is instantiated, and this is when `context.getPreferences()` is invoked and the preferences are cached into memory. Also, in my opinion, isn't it better to let the `SharedPreferences` handle it by default, because otherwise we would have to manually handle the invalidation/update of "memory prefs", e.g. on `identify()` or `reset()`. Please correct me if I'm wrong or I'm missing something.

marandaneto: Nope you are not wrong, `SharedPreferences` does that, but we need an atomic operation with multiple instructions, the `SharedPreferences` is only locking the preferences itself, and our lock locks multiple things, for example, multiple threads calling `identify` in parallel might give a wrong state.

beradeep: Okay. How should I approach implementing this then?

marandaneto: as you are doing already, using a var and a lock, `identifiedLock`.
the difference is that you will use a global var `isIdentified`, instead of local var, similar to `sessionId` and `sessionLock`.

beradeep: Can you please elaborate a bit what you mean by a global var and local var in this context?

marandaneto: something like this https://github.com/PostHog/posthog-android/pull/160/commits/a7bf27aa3b83887c93ffdef517177396465c25de
https://www.enjoytutorials.com/kotlin-local-and-global-variables-tutorial/

marandaneto:
@beradeep would you pick up this PR again? just so I don't get to it as well, thanks.

beradeep:
Yes, I'll pick it up. I think I should go with the solution you provided in [a7bf27a](https://github.com/PostHog/posthog-android/commit/a7bf27aa3b83887c93ffdef517177396465c25de).

beradeep:
When should I turn `isIdentifiedLoaded` back to false? Inside `onReset()`?

beradeep:
@marandaneto Sorry, I closed this PR accidentally. I've created a new one #167.

PR #163, opened 2024-08-22
chore: file name in storage matches the event uuid

marandaneto:
I was looking at the logs and trying to find the event id based on the logs but they don't match, they should
_#skip-changelog_

matching the ids

## :pencil: Checklist


marandaneto:
YOLO since i need this to debug something else.

PR #167, opened 2024-08-27
add isIdentified property and check on identify()

beradeep:

This change adds the is_identified property (absent earlier) to check user is identified or not (anonymous). This change also adds a fix to identify() to ensure that an "identify" event is captured if and only if the user isn't already identified (anonymous). Also adds an in-memory flag to check if the property is already loaded or not. 

fix https://github.com/PostHog/posthog-android/issues/136 & https://github.com/PostHog/posthog-android/issues/108

- Added unit test to assert against duplicate "identify" captures
- Added unit test to assert "identify" works as usual on reset
- Verified that this change doesn't break existing tests

## :pencil: Checklist

### DIFF
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -235,8 +235,8 @@ public class PostHog private constructor(
         get() {
             synchronized(identifiedLock) {
                 if (!isIdentifiedLoaded) {
-                    this.isIdentified = getPreferences().getValue(IS_IDENTIFIED) as? Boolean
-                        ?: (this.distinctId != this.anonymousId)
+                    isIdentified = getPreferences().getValue(IS_IDENTIFIED) as? Boolean
+                        ?: (distinctId != anonymousId)
                     isIdentifiedLoaded = true
                 }
             }
@@ -292,6 +292,8 @@ public class PostHog private constructor(
                     }
                 }
             }
+
+            props["\$is_identified"] = isIdentified
         }
 
         PostHogSessionManager.getActiveSessionId()?.let { sessionId ->
@@ -506,7 +508,13 @@ public class PostHog private constructor(
             config?.logger?.log("identify called with invalid anonymousId: $anonymousId.")
         }
 
-        if (previousDistinctId != distinctId && !this.isIdentified) {
+        if (previousDistinctId != distinctId && !isIdentified) {
+            // this has to be set before capture since this flag will be read during the event
+            // capture
+            synchronized(identifiedLock) {
+                isIdentified = true
+            }
+
             capture(
                 "\$identify",
                 distinctId = distinctId,
@@ -522,7 +530,6 @@ public class PostHog private constructor(
                 config?.logger?.log("identify called with invalid former distinctId: $previousDistinctId.")
             }
             this.distinctId = distinctId
-            this.isIdentified = true
 
             // only because of testing in isolation, this flag is always enabled
             if (reloadFeatureFlags) {
@@ -674,7 +681,13 @@ public class PostHog private constructor(
         queue?.clear()
         replayQueue?.clear()
         featureFlagsCalled.clear()
+
+        synchronized(identifiedLock) {
+            isIdentifiedLoaded = false
+        }
+
         endSession()
+        startSession()
     }
 
     private fun isEnabled(): Boolean {
@@ -732,9 +745,6 @@ public class PostHog private constructor(
             return
         }
 
-        synchronized(identifiedLock) {
-            isIdentifiedLoaded = false
-        }
         PostHogSessionManager.endSession()
     }
 
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -687,7 +687,6 @@ public class PostHog private constructor(
         }
 
         endSession()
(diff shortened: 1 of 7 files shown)


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -227,6 +231,24 @@ public class PostHog private constructor(
             getPreferences().setValue(DISTINCT_ID, value)
         }
 
+    private var isIdentified: Boolean = false
+        @JvmName("get-isIdentified")
```
marandaneto, line 235: we don't need this

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -227,6 +231,24 @@ public class PostHog private constructor(
...
+        get() {
+            synchronized(identifiedLock) {
+                if (!isIdentifiedLoaded) {
+                    println("hi")
```
marandaneto, line 239: we can remove this

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -227,6 +231,24 @@ public class PostHog private constructor(
...
+                    this.isIdentified = getPreferences().getValue(IS_IDENTIFIED) as? Boolean
+                        ?: (this.distinctId != this.anonymousId)
+                    isIdentifiedLoaded = true
+                }
+            }
+            return field
+        }
+        set(value) {
+            field = value
+            getPreferences().setValue(IS_IDENTIFIED, value)
```
marandaneto, line 249: I guess we should sync this with `identifiedLock` as well

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -484,27 +506,30 @@ public class PostHog private constructor(
...
             }
             this.distinctId = distinctId
+            this.isIdentified = true
 
             // only because of testing in isolation, this flag is always enabled
             if (reloadFeatureFlags) {
                 reloadFeatureFlags()
             }
+        } else {
+            config?.logger?.log("already identified with id : $distinctId.")
```
marandaneto, line 532:
```suggestion
            config?.logger?.log("already identified with id: $distinctId.")
```

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -707,6 +732,7 @@ public class PostHog private constructor(
             return
         }
 
+        isIdentifiedLoaded = false
```
marandaneto, line 735: we should sync this with `identifiedLock`

marandaneto:
Thanks @beradeep 
I left a few comments, after addressing, we can merge it
don't forget to add an entry to the changelog.

posthog/src/main/java/com/posthog/PostHog.kt
beradeep: Oh, my bad, I'm sorry I forgot to omit that.

posthog/src/main/java/com/posthog/PostHog.kt
beradeep: okay 👍

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -707,6 +732,9 @@ public class PostHog private constructor(
             return
         }
 
+        synchronized(identifiedLock) {
+            isIdentifiedLoaded = false
+        }
```
marandaneto, line 737: This should go in the reset method and not here

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -484,27 +506,30 @@ public class PostHog private constructor(
...
         }
 
-        capture(
-            "\$identify",
-            distinctId = distinctId,
-            properties = props,
-            userProperties = userProperties,
-            userPropertiesSetOnce = userPropertiesSetOnce,
-        )
+        if (previousDistinctId != distinctId && !this.isIdentified) {
```
marandaneto, line 509: I think the behavior here is different, see the JS SDK https://github.com/PostHog/posthog-js/blob/e8a645134a03477d29ab6839ae7f10fcd6e4157e/src/posthog-core.ts#L1316-L1354
if it's the very same `distinctId` again won't capture anything (and capture a $set event if needed)
if it's a new distinctId, clear all the cache from the previous distinctId, and identify again with the new distinctId, reload feature flags, etc

beradeep: Isn't it nearly the same behavior, except for the `setPersonProperties` and `setPersonPropertiesForFlags`, which are not yet implemented here.

marandaneto: nvm, I think I misread the PR, you are right.

marandaneto:
The flag `$is_identified` isn't being set though, so this issue can't be closed https://github.com/PostHog/posthog-android/issues/136

posthog/src/main/java/com/posthog/PostHog.kt
marandaneto: i pushed 2 commits to fix this

marandaneto:
> The flag `$is_identified` isn't being set though, so this issue can't be closed #136

i pushed 2 commits to fix this

marandaneto (approved):

PR #171, opened 2024-09-04
add support to person profiles

### LINKED ISSUE #159: Support `person_profiles`
### Description

https://github.com/PostHog/product-internal/pull/637
https://github.com/PostHog/posthog-js/blob/0b40e5af5381c7d356de31fc0458e91afbfa9a6b/src/types.ts#L206-L213

`person_profiles` should be `identified_only` by default.

Options: `always`, `never`, `identified_only`.

- [ ] Android
- [ ] [iOS](https://github.com/PostHog/posthog-ios/issues/185)
- [ ] [RN](https://github.com/PostHog/posthog-js-lite/issues/240)

thisames:

This PR adds support for the `personProfiles` configuration option in the PostHog SDK. The `personProfiles` option allows users to control how person profile data is processed during various events.

reference
https://github.com/PostHog/product-internal/pull/637
https://github.com/PostHog/posthog-js/blob/0b40e5af5381c7d356de31fc0458e91afbfa9a6b/src/types.ts#L206-L213

fix #159

- Manually tested by sending multiple events to the PostHog dashboard with all three `personProfiles` settings ("always", "never", "identified_only") to confirm that the events are processed according to the expected behavior for each setting.

- Added unit tests in `PostHogTest` to verify the behavior of the `personProfiles` setting:
  - Ensured that all events are captured when `personProfiles` is set to "always".
  - Validated that no profile processing occurs when `personProfiles` is set to "never".
  - Checked that non-identification events do not process profiles when `personProfiles` is set to "identified_only".
  - Confirmed that identification events process profiles correctly when `personProfiles` is set to "identified_only".

#### ⚠️ Observations
- Noticed a potential issue: when `personProfiles` is set to "never" or "identified_only", the flag `process_person_profile` is added to the "screen" event, but the event still appears on the dashboard. This behavior is unexpected, as the flag should filter out the event, preventing it from showing up on the dashboard. It seems that additional logic may be needed to properly handle "screen" events according to the `personProfiles` setting.

## :pencil: Checklist


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -378,49 +379,63 @@ public class PostHog private constructor(
...
-                )
+            captureInternal(event, newDistinctId, properties, userProperties, userPropertiesSetOnce, groups)
+        } catch (e: Throwable) {
+            config?.logger?.log("Capture failed: $e.")
+        }
+    }
 
-            // sanitize the properties or fallback to the original properties
-            val sanitizedProperties = config?.propertiesSanitizer?.sanitize(mergedProperties.toMutableMap()) ?: mergedProperties
+    private fun captureInternal(
```
marandaneto, line 398: I think this method should take a `processPersonProfile`, so you don't need the if condition in the L385, not an early return to avoid another call to `captureInternal`

marandaneto: Or even make this flag until the `buildProperties` method, and append the `$process_person_profile` there

posthog/src/main/java/com/posthog/internal/PersonProfileMode.kt
```diff
@@ -0,0 +1,22 @@
...
+        public fun fromConfig(config: String?): PersonProfileMode {
+            return values().find { it.mode == config } ?: ALWAYS
+        }
+    }
+
+    public fun shouldProcessProfile(event: String): Boolean {
+        return when (this) {
+            NEVER -> false
+            ALWAYS -> true
+            IDENTIFIED_ONLY -> event == "\$identify" || event == "\$create_alias" || event == "\$groupidentify"
```
marandaneto, line 19: `_hasPersonProcessing` does not have those checks in JS, or did I miss something?

marandaneto: This should check a few other things apparently such as https://github.com/PostHog/posthog-js/blob/34204f4aff107162bdb259d2a7c855d67bf0260c/src/posthog-core.ts#L1905-L1906

posthog/src/main/java/com/posthog/internal/PersonProfileMode.kt
```diff
@@ -0,0 +1,22 @@
...
+
+public enum class PersonProfileMode(public val mode: String) {
+    NEVER("never"),
+    ALWAYS("always"),
+    IDENTIFIED_ONLY("identified_only"),
+    ;
+
+    public companion object {
+        public fun fromConfig(config: String?): PersonProfileMode {
+            return values().find { it.mode == config } ?: ALWAYS
```
marandaneto, line 11: if default is identified_only, then here it should be the same

marandaneto:
Add a TODO somewhere that we need to implement `createPersonProfile` and `setPersonProperties`, which can be a separate PR, another time.

marandaneto:
I think this check is also missing somewhere https://github.com/PostHog/posthog-js/blob/34204f4aff107162bdb259d2a7c855d67bf0260c/src/posthog-core.ts#L829 which is about setting `userPropertiesSetOnce`

thisames:
I will fix all that you commented, soon update

thisames:
I'm sorry @marandaneto, I thought the implementation of this feature was based solely on the documentation. I didn't realize there was already a reference implementation for it, which is why my code looks confusing, as I developed it from my own logic.

I now understand exactly how to do it correctly, but I still have a few questions.

In the identify_only documentation, it states that it will only process events related to identifying:

'identified_only' - we will only process persons when you call posthog.identify, posthog.alias, posthog.setPersonProperties, posthog.group, posthog.setPersonPropertiesForFlags, or posthog.setGroupPropertiesForFlags. Anonymous users won't get person profiles.
But in the actual posthog-js implementation, it seems to process any type of event, or am I mistaken?

Additionally, in this PR, is it necessary to implement the persistence mechanism, like for dataSetOnce, for example?

marandaneto:
> I'm sorry @marandaneto, I thought the implementation of this feature was based solely on the documentation. I didn't realize there was already a reference implementation for it, which is why my code looks confusing, as I developed it from my own logic.
> 
> I now understand exactly how to do it correctly, but I still have a few questions.
> 
> In the `identify_only` documentation, it states that it will only process events related to identifying:
> 
> _`'identified_only' - we will only process persons when you call posthog.identify, posthog.alias, posthog.setPersonProperties, posthog.group, posthog.setPersonPropertiesForFlags, or posthog.setGroupPropertiesForFlags. Anonymous users won't get person profiles.`_
> 
> But in the actual posthog-js implementation, it seems to process any type of event, or am I mistaken?
> 
> Additionally, in this PR, is it necessary to implement the persistence mechanism, like for dataSetOnce, for example?

the flag is set for any type of event, the same as the JS implementation.
the dataSetOnce bits can be done later such as https://github.com/PostHog/posthog-android/pull/171#issuecomment-2330727362, so everything but this part can be the same as the JS implementation.

thisames:
@marandaneto I updated the PR with only what's necessary to make the personProfiles config feature work. Just the hasPersonProcessing function solves the problem.

However, I introduced a complete version with the dependencies that complete the personProfile feature. If possible, take a look at this version and see if it makes sense. Based on my tests in core-js and in this version's code, everything seems fine.

https://github.com/thisames/posthog-android/commit/0f0c39a37799e867c2a484b6862341595a211923

Please analyze whether it makes sense to include this in the PR or leave it for later.

marandaneto:
About the commit https://github.com/thisames/posthog-android/commit/0f0c39a37799e867c2a484b6862341595a211923
Looks good, I'd have a few comments, lets make a new PR so we can discuss.

posthog/src/main/java/com/posthog/PostHogConfig.kt
```diff
@@ -102,6 +103,14 @@ public open class PostHogConfig(
      * generating anonymous id (which as of now is just random UUID v4)
      */
     public var getAnonymousId: ((UUID) -> UUID) = { it },
+    /**
+     * Determines the behavior for processing user profiles.
+     * - `"always"`: Always processes user profile data for all events.
+     * - `"never"`: Never processes user profile data. This means that anonymous users will not be merged when they sign up or log in.
+     * - `"identified_only"` (default): Processes user profiles only when an identification event occurs, such as `identify`, `alias`, or `groupidentify`.
```
marandaneto, line 110:
```suggestion
     * - `"identified_only"` (default): Processes user profiles only when an identification event occurs, such as `identify`, `alias`, or `groupidentify`.
```

this is not up to date since the `$process_person_profile` is added for all events

thisames: > this is not up to date since the `$process_person_profile` is added for all events

sorry, I don't know if I understand these commit suggestion

thisames:
```suggestion
     * - `"identified_only"` (default): Processes user profiles for all events, with additional behavior when an identification event occurs, such as `identify`, `alias`, or `groupidentify`.
```

what do you think?

marandaneto:
@thisames see my PR https://github.com/PostHog/posthog-ios/pull/187
I'd just copy what its there (and I copied what was in JS).
Also its missing the `requirePersonProcessing` bits, see my PR as well

marandaneto:
@thisames I pushed a few changes to fix and improve a few things, thanks for the PR.

marandaneto (approved):

PR #176, opened 2024-09-13
chore: more session replay metadata

marandaneto:
Ideas for https://posthog.slack.com/archives/C043VJ93L3B/p1725259136850079

## :pencil: Checklist

### DIFF
--- a/posthog/src/main/java/com/posthog/internal/replay/RRUtils.kt
+++ b/posthog/src/main/java/com/posthog/internal/replay/RRUtils.kt
@@ -6,7 +6,7 @@ import com.posthog.PostHogInternal
 @PostHogInternal
 public enum class PostHogSessionReplayMode(public val mode: String) {
     WIREFRAME("wireframe"),
-    SCREENSHOT("screenshot")
+    SCREENSHOT("screenshot"),
 }
 
 @PostHogInternal
@@ -15,7 +15,7 @@ public fun List<RREvent>.capture(mode: PostHogSessionReplayMode = PostHogSession
         mutableMapOf(
             "\$snapshot_data" to this,
             "\$snapshot_source" to "mobile",
-            "\$snapshot_mode" to mode.mode
+            "\$snapshot_mode" to mode.mode,
         )
     PostHog.capture("\$snapshot", properties = properties)
 }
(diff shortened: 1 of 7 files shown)


posthog/src/main/java/com/posthog/internal/replay/RRUtils.kt
```diff
@@ -4,11 +4,18 @@ import com.posthog.PostHog
...
+    SCREENSHOT("screenshot"),
+}
+
+@PostHogInternal
+public fun List<RREvent>.capture(mode: PostHogSessionReplayMode = PostHogSessionReplayMode.WIREFRAME) {
     val properties =
         mutableMapOf(
             "\$snapshot_data" to this,
             "\$snapshot_source" to "mobile",
+            "\$snapshot_mode" to mode.mode,
```
marandaneto, line 18: maybe `source` and `mode` should only be part of the full snapshot?

technically the `mode` can also be inferred server side since we only send `wireframe.type=screenshot`.

the SDK and platform can also be inferred from `$lib`, so there is no need for something else here, eg `posthog-android`, `posthog-ios`, `posthog-react-native`, etc.

pauldambra: > maybe source and mode should only be part of the full snapshot?

if we do the same as snapshot source then we'll take the first not-null value we receive in a session as the only value for that property. so safest to be on everything for debugging, but safe to be on only some messages

(with the assumption that very few or no sessions will span web and mobile or wireframe and snapshot)

> technically the mode can also be inferred server side 

ideally we'd keep making ingestion less dependent on the content of the messages - i.e. we can't extra compress the web messages right now because we need to read the contents in blobby. so ideally we'd be explicit here - doesn't help old versions obvs but better in the long run IMHO

marandaneto: What's about the naming `$snapshot_mode` and as a sibling of `$snapshot_source`, is that ok?
Are we ok to infer the SDK from `$lib` as well or should we be more explicit here?

pauldambra: yep mode is good...
so long as we're sending $lib then i think we need that to be unique anyway so we're good there

marandaneto:
we are optimizing the screenshot snapshots with better encoding and some other things, closing this for now.
we already send $lib, so also do not need to do anything else.

PR #177, opened 2024-09-13
fix: respect session replay project settings from app start

marandaneto:
Similar to https://github.com/PostHog/posthog-ios/pull/150
If replay is active in the SDK config but disabled in the project settings, a recording would start and would end after ~1s when the decide API finishes.

## :pencil: Checklist


pauldambra (approved):
👍

PR #180, opened 2024-09-17
recording: react native has window attached but needs view ready callback

marandaneto:
follow up https://github.com/PostHog/posthog-android/pull/179
_#skip-changelog_
the window is already attached but we don't have the callback yet so we need to allow this

## :pencil: Checklist

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -126,6 +126,8 @@ public class PostHogReplayIntegration(
             view.phoneWindow?.let { window ->
                 var hasDecorView = false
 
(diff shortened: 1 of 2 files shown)


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -299,15 +306,7 @@ public class PostHogReplayIntegration(
...
-            view.phoneWindow?.let { window ->
-                window.peekDecorView()?.let { decorView ->
-                    val hasDecorView = decorViews[decorView] != null
-
-                    if (!hasDecorView) {
-                        addView(view, true)
-                    }
-                }
-            }
+            addView(view)
```
marandaneto, line 309: the check is done inside `addView` already

Issue #119, opened 2024-04-09
Set `$screen_name` to all events


marandaneto:
### Description

- [ ] Android
- [ ] iOS
- [X] Flutter (when calling `screen` via PosthogObserver or manually)
- [X] RN (when calling `screen` via wix, react-navigation or manually)

marandaneto:
@thisames would you like to take this one?

thisames:
> @thisames would you like to take this one?

yes

PR #181, opened 2024-09-23
chore: add support to set screen name to all event

thisames:

#119 

The tests have been updated to include checks for the new \$screen_name property in any event.
I tested it manually and analyse the props in dashboard.

## :pencil: Checklist

### DIFF
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -68,6 +68,8 @@ public class PostHog private constructor(
     private var isIdentifiedLoaded: Boolean = false
     private var isPersonProcessingLoaded: Boolean = false
 
+    private lateinit var currentScreen: String
+
     public override fun <T : PostHogConfig> setup(config: T) {
         synchronized(setupLock) {
             try {
@@ -317,6 +319,10 @@ public class PostHog private constructor(
                 }
             }
 
+            currentScreen.let {
+                props["\$screen_name"] = currentScreen
+            }
+
             userProperties?.let {
                 props["\$set"] = it
             }
@@ -505,6 +511,8 @@ public class PostHog private constructor(
 
         val props = mutableMapOf<String, Any>()
         props["\$screen_name"] = screenTitle
+
+        currentScreen = screenTitle
 
         properties?.let {
             props.putAll(it)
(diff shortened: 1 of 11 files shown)


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -68,6 +68,8 @@ public class PostHog private constructor(
     private var isIdentifiedLoaded: Boolean = false
     private var isPersonProcessingLoaded: Boolean = false
 
+    private lateinit var currentScreen: String
```
marandaneto, line 71: I think this approach leaves room for too many false positives.
eg I call `screen(x)` for screen X, but I don't call `screen(y)` for screen Y, when events are captured in the screen `y`, the screen name is `x`.

thisames: I see, you are correct. The problem with using `PostHogActivityLifecycleCallbackIntegration.kt` is that we would have to add method signature in `PostHogInterface`, and i dont know if it make sense.

see theses changes
https://github.com/thisames/posthog-android/commit/7d4a6f5694798f9560d3c629193bf608cd961819

do you think this make sense?

marandaneto: I think we should add a new concept to the SDK that allows enriching the event after being captured, for example, an event processor.
I will make an example and you can expand from there

marandaneto: See this example https://github.com/PostHog/posthog-android/compare/chore/processors?expand=1

thisames: Thank you, you made it much easier for me. I will update this PR with the changes and then you check and give me the ok, if it works I will do the unit tests

marandaneto: Another option is to use the [super properties](https://posthog.com/docs/libraries/android) for the screen name
https://github.com/PostHog/posthog-android/blob/3a0ede85b9ed25dd993ff3944f74a7187dc175a7/posthog/src/main/java/com/posthog/PostHog.kt#L783-L802
both would work the same

thisames: I prefer the solution with processors, it's working well

thisames:
@marandaneto ready for review

thisames:
api generated 😅

marandaneto:
Closing the issue as a part of large repository cleanup, due to it being inactive and/or outdated.
Please do not hesitate to ping me if it is still relevant, and I will happily reopen and work on it.
Cheers!

PR #185, opened 2024-09-26
fix: isFeatureEnabled wasn't sending the $feature_flag_called event

marandaneto:
Closes https://posthoghelp.zendesk.com/agent/tickets/18551 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/21011)

## :pencil: Checklist


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -692,18 +692,17 @@ public class PostHog private constructor(
         if (!isEnabled()) {
             return defaultValue
         }
-        return featureFlags?.isFeatureEnabled(key, defaultValue) ?: defaultValue
+        val value = featureFlags?.isFeatureEnabled(key, defaultValue) ?: defaultValue
```
neilkakkar, line 695: in most SDKs isFeatureEnabled calls getFeatureFlag, so all of this handling can be in one place. Think we can do the same here?

neilkakkar (approved):

posthog/src/main/java/com/posthog/PostHog.kt
marandaneto: because of this tiny bit of code its trickier https://github.com/PostHog/posthog-android/blob/3a0ede85b9ed25dd993ff3944f74a7187dc175a7/posthog/src/main/java/com/posthog/internal/PostHogFeatureFlags.kt#L205-L214
since this is a compiled language and it has strongly typed methods, will get it merged for now and think if its worth the refactoring 🧐

Issue #138, opened 2024-05-31
Adds referrerURL automatically


marandaneto:
### Description

Context https://github.com/ueman/flutter_referrer_info/issues/2

Android already listens to `ActivityLifecycleCallbacks` so it'd be easier to add the referrerURL reading the intent Intent.EXTRA_REFERRER, Intent.EXTRA_REFERRER_NAME and or `com.android.browser.application_id`.

Prop $referrer, and $referring_domain

- [x] Android
- [ ] iOS
- [ ] RN
- [ ] Flutter

marandaneto:
@thisames would you like to take this one?

marandaneto:
On Android, there's also this option https://developer.android.com/google/play/installreferrer

marandaneto:
On iOS, [this](https://developer.apple.com/help/app-store-connect/view-app-analytics/view-acquisition-sources) and [this](https://github.com/firebase/firebase-ios-sdk/blob/71d8e92d168333f79ee29a555d32fee086e66ca5/FirebaseAnalytics/Tests/SwiftUnit/AnalyticsAPITests.swift#L61-L62)
We cannot make it automatic but we can provide a method that takes a `NSUserActivity` at least.

beradeep:
@marandaneto I'll be happy to take this one. Is it already assigned?

marandaneto:
> @marandaneto I'll be happy to take this one. Is it already assigned?

feel free to take it.

PR #186, opened 2024-10-01
add referrer url automatically

beradeep:

#138

Added tests to include checks for `$referrer` and `$referring_domain` properties in case of referrals from apps and websites.

## :pencil: Checklist

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
@@ -38,7 +38,7 @@ internal class PostHogActivityLifecycleCallbackIntegration(
                     config.logger.log("Deep link $data has invalid query param names.")
                 } finally {
                     data?.let { props["url"] = it.toString() }
-                    activity.getReferrerInfo(config)?.let { props.putAll(it) }
+                    activity.getReferrerInfo(config).let { props.putAll(it) }
                     PostHog.capture("Deep Link Opened", properties = props)
                 }
             }
--- a/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
@@ -38,7 +38,7 @@ internal class PostHogActivityLifecycleCallbackIntegration(
                     config.logger.log("Deep link $data has invalid query param names.")
                 } finally {
                     data?.let { props["url"] = it.toString() }
-                    activity.getReferrerInfo(config).let { props.putAll(it) }
+                    activity.intent.getReferrerInfo(config).let { props.putAll(it) }
                     PostHog.capture("Deep Link Opened", properties = props)
                 }
             }
--- a/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
@@ -22,19 +22,23 @@ internal class PostHogActivityLifecycleCallbackIntegration(
         savedInstanceState: Bundle?,
     ) {
         if (config.captureDeepLinks) {
-            activity.intent.data?.let {
+            activity.intent?.let { intent ->
                 val props = mutableMapOf<String, Any>()
+                val data = intent.data
                 try {
-                    for (item in it.queryParameterNames) {
-                        val param = it.getQueryParameter(item)
-                        if (!param.isNullOrEmpty()) {
-                            props[item] = param
+                    data?.let {
+                        for (item in it.queryParameterNames) {
+                            val param = it.getQueryParameter(item)
+                            if (!param.isNullOrEmpty()) {
+                                props[item] = param
+                            }
                         }
                     }
                 } catch (e: UnsupportedOperationException) {
-                    config.logger.log("Deep link $it has invalid query param names.")
+                    config.logger.log("Deep link $data has invalid query param names.")
                 } finally {
-                    props["url"] = it.toString()
+                    props["url"] = data.toString()
+                    activity.getReferrerInfo(config)?.let { props.putAll(it) }
                     PostHog.capture("Deep Link Opened", properties = props)
                 }
             }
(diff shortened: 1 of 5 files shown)


marandaneto:
@beradeep whats about if the `[Browser.EXTRA_APPLICATION_ID](https://stackoverflow.com/questions/9902225/browser-extra-application-id-do-not-work-in-the-tablet-do-you-know-other-work-a)` is provided? wondering if this is possible to test it

posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidUtils.kt
```diff
@@ -144,6 +150,38 @@ internal fun Activity.activityLabelOrName(config: PostHogAndroidConfig): String?
...
+            return null
+        } else if (referrer == null) {
+            Uri.parse(referrerName)?.let { uri ->
+                referrerInfoMap["\$referrer"] = uri.toString()
+                uri.host?.let { referrerInfoMap["\$referring_domain"] = it }
+            }
+        } else {
+            referrerInfoMap["\$referrer"] = referrer.toString()
+            referrer.host?.let { referrerInfoMap["\$referring_domain"] = it }
+        }
```
marandaneto, line 177: I think it'd be easier to read if you do an `if` to `referrer` and another `if` to `referrerName` so you don't need the if/elseif/else

posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidUtils.kt
```diff
@@ -144,6 +150,38 @@ internal fun Activity.activityLabelOrName(config: PostHogAndroidConfig): String?
...
+            referrer = intent.getParcelableExtra(Intent.EXTRA_REFERRER)
+            referrerName = intent.getStringExtra(Intent.EXTRA_REFERRER_NAME)
+        } else {
+            referrer = intent.getParcelableExtra("android.intent.extra.REFERRER")
+            referrerName = intent.getStringExtra("android.intent.extra.REFERRER_NAME")
+        }
+        if (referrer == null && referrerName == null) {
+            return null
+        } else if (referrer == null) {
+            Uri.parse(referrerName)?.let { uri ->
```
marandaneto, line 170: does Uti.parse throw if its an invalid uri or just returns null?

beradeep: Oh, I made a mistake, Uri.parse does throw NullPointerException

posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidUtils.kt
beradeep: Okay, I'll change it

beradeep:
> @beradeep whats about if the `[Browser.EXTRA_APPLICATION_ID](https://stackoverflow.com/questions/9902225/browser-extra-application-id-do-not-work-in-the-tablet-do-you-know-other-work-a)` is provided? wondering if this is possible to test it

I'm curious why do we need this? Is it for another prop like `$browser`?

posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidUtils.kt
marandaneto: in this case its better to make an extension to Uri.parse that swallow the exception if its invalid/null instead of dropping all the data that could be captured (eg there's referrer but not referrerName and vice-versa)

marandaneto:
> > @beradeep whats about if the `[Browser.EXTRA_APPLICATION_ID](https://stackoverflow.com/questions/9902225/browser-extra-application-id-do-not-work-in-the-tablet-do-you-know-other-work-a)` is provided? wondering if this is possible to test it
> 
> I'm curious why do we need this? Is it for another prop like `$browser`?

no, I am wondering if the app is opened via the browser, if we need to account for this intent name or not, maybe we should just ignore it, or maybe we should read from it as well since the docs isn't that clear when it's used.

posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
```diff
@@ -22,19 +22,23 @@ internal class PostHogActivityLifecycleCallbackIntegration(
...
+                            }
                         }
                     }
                 } catch (e: UnsupportedOperationException) {
-                    config.logger.log("Deep link $it has invalid query param names.")
+                    config.logger.log("Deep link $data has invalid query param names.")
                 } finally {
-                    props["url"] = it.toString()
+                    data?.let { props["url"] = it.toString() }
+                    activity.intent.getReferrerInfo(config).let { props.putAll(it) }
```
marandaneto, line 41:
```suggestion
                    intent.getReferrerInfo(config).let { props.putAll(it) }
```

since you do `activity.intent?.let { intent ->` already

marandaneto:
happy to approve and merge after checking out this [comment](https://github.com/PostHog/posthog-android/pull/186#issuecomment-2385426475), and thanks for the PR o/

marandaneto:
don't forget about the changelog entry.

beradeep:
> > > @beradeep whats about if the `[Browser.EXTRA_APPLICATION_ID](https://stackoverflow.com/questions/9902225/browser-extra-application-id-do-not-work-in-the-tablet-do-you-know-other-work-a)` is provided? wondering if this is possible to test it
> > 
> > 
> > I'm curious why do we need this? Is it for another prop like `$browser`?
> 
> no, I am wondering if the app is opened via the browser, if we need to account for this intent name or not, maybe we should just ignore it, or maybe we should read from it as well since the docs isn't that clear when it's used.

i'm still not able to understand this. when the app is opened via a link from the browser, the deep link url and the referrer site does get read and captured from the intent built automatically by the Android system. i think you meant that you want to test this  feature. am i right?

beradeep:
> @beradeep https://developer.android.com/reference/android/provider/Browser#EXTRA_APPLICATION_ID
> 
> > The name of the extra data when starting the Browser from another application.
> > The value is a unique identification string that will be used to identify the calling application. The Browser will attempt to reuse the same window each time the application launches the Browser with the same identifier.
> > Constant Value: "com.android.browser.application_id"
> 
> Right now we don't do `intent.data.getString(Browser.EXTRA_APPLICATION_ID)`, do we have to? for this specific use case as described above? that is the question I'd like to answer.

I don't think we need that here. This is only required when we need to get the name of the application calling the browser.

PR #196, opened 2024-10-07
recording: OnTouchEventListener try catch guard to swallow unexpected errors take 2

marandaneto:
Tries to fix https://github.com/PostHog/posthog-react-native-session-replay/issues/4
It's unclear if PostHog SDK is responsible for that, since the SDK intercepts the touch, the SDK will be always part of the stack trace, it does not mean its the SDK causing it.

## :pencil: Checklist

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -75,7 +75,7 @@ import com.posthog.internal.replay.RRWireframe
 import com.posthog.internal.replay.capture
 import curtains.Curtains
 import curtains.OnRootViewsChangedListener
-import curtains.OnTouchEventListener
+import curtains.TouchEventInterceptor
 import curtains.onDecorViewReady
 import curtains.phoneWindow
 import curtains.touchEventInterceptors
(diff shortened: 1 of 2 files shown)


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -216,27 +216,34 @@ public class PostHogReplayIntegration(
...
+                        when (motionEvent.action.and(MotionEvent.ACTION_MASK)) {
+                            MotionEvent.ACTION_DOWN -> {
+                                generateMouseInteractions(timestamp, motionEvent, RRMouseInteraction.TouchStart)
+                            }
+                            MotionEvent.ACTION_UP -> {
+                                generateMouseInteractions(timestamp, motionEvent, RRMouseInteraction.TouchEnd)
+                            }
                         }
+                    } catch (e: Throwable) {
+                        config.logger.log("Executor#OnTouchEventListener $motionEvent failed: $e.")
```
marandaneto, line 239: if this happens, its our fault

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -216,27 +216,34 @@ public class PostHogReplayIntegration(
...
                     }
-                } catch (e: Throwable) {
-                    config.logger.log("OnTouchEventListener $motionEvent failed: $e.")
                 }
+                state
+            } catch (e: Throwable) {
+                config.logger.log("TouchEventInterceptor $motionEvent failed: $e.")
+                throw e
```
marandaneto, line 245: if this happens, its not our fault

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -216,27 +216,34 @@ public class PostHogReplayIntegration(
     }
 
     private val onTouchEventListener =
-        OnTouchEventListener { motionEvent ->
+        TouchEventInterceptor { motionEvent, dispatch ->
```
marandaneto, line 219: changed to a lower level interceptor, so we call the dispatch and log out

marandaneto:
> Only comment is that we may want to move the `isSessionReplayEnabled` check to when `generateMouseInteractios` is called so we get some error logs even when session replay is not enabled. If that makes sense to you. Otherwise LGMT

this code is not even executed if session replay is disabled, so moving the check would not have any effect

PR #198, opened 2024-10-16
feat: add replay masking to jetpack compose views

### LINKED ISSUE #158: Masking and redacting Jetpack Compose views
### Description

See note https://posthog.com/docs/session-replay/android

beradeep:
Fix  #158

Tested on a sample app. Auto as well as manual masking using `Modifier.postHogMaskReplay()`, both working as expected.

## :pencil: Checklist

### DIFF
--- a/buildSrc/src/main/java/PosthogBuildConfig.kt
+++ b/buildSrc/src/main/java/PosthogBuildConfig.kt
@@ -14,7 +14,7 @@ object PosthogBuildConfig {
     }
 
     object Android {
-        val COMPILE_SDK = 34
+        val COMPILE_SDK = 33
 
         // when changing this, remember to check the ANIMAL_SNIFFER_SDK_VERSION
         // Session Replay (addOnFrameMetricsAvailableListener requires API 26)
@@ -55,6 +55,7 @@ object PosthogBuildConfig {
         val OKHTTP = "4.11.0"
         val CURTAINS = "1.2.5"
         val ANDROIDX_CORE = "1.5.0"
+        val ANDROIDX_COMPOSE = "1.0.0"
 
         // tests
         val ANDROIDX_JUNIT = "1.1.5"
--- a/buildSrc/src/main/java/PosthogBuildConfig.kt
+++ b/buildSrc/src/main/java/PosthogBuildConfig.kt
@@ -14,7 +14,7 @@ object PosthogBuildConfig {
     }
 
     object Android {
-        val COMPILE_SDK = 33
+        val COMPILE_SDK = 34
 
         // when changing this, remember to check the ANIMAL_SNIFFER_SDK_VERSION
         // Session Replay (addOnFrameMetricsAvailableListener requires API 26)
--- a/posthog-android/build.gradle.kts
+++ b/posthog-android/build.gradle.kts
@@ -88,8 +88,8 @@ dependencies {
     implementation("androidx.lifecycle:lifecycle-process:${PosthogBuildConfig.Dependencies.LIFECYCLE}")
     implementation("androidx.lifecycle:lifecycle-common-java8:${PosthogBuildConfig.Dependencies.LIFECYCLE}")
     implementation("androidx.core:core:${PosthogBuildConfig.Dependencies.ANDROIDX_CORE}")
+    implementation("androidx.compose.ui:ui:${PosthogBuildConfig.Dependencies.ANDROIDX_COMPOSE}")
     implementation("com.squareup.curtains:curtains:${PosthogBuildConfig.Dependencies.CURTAINS}")
-    implementation("androidx.compose.ui:ui-android:1.7.3")
 
     // compatibility
     signature("org.codehaus.mojo.signature:java18:${PosthogBuildConfig.Plugins.SIGNATURE_JAVA18}@signature")
--- a/posthog-android/build.gradle.kts
+++ b/posthog-android/build.gradle.kts
@@ -88,8 +88,10 @@ dependencies {
     implementation("androidx.lifecycle:lifecycle-process:${PosthogBuildConfig.Dependencies.LIFECYCLE}")
     implementation("androidx.lifecycle:lifecycle-common-java8:${PosthogBuildConfig.Dependencies.LIFECYCLE}")
     implementation("androidx.core:core:${PosthogBuildConfig.Dependencies.ANDROIDX_CORE}")
-    implementation("androidx.compose.ui:ui:${PosthogBuildConfig.Dependencies.ANDROIDX_COMPOSE}")
     implementation("com.squareup.curtains:curtains:${PosthogBuildConfig.Dependencies.CURTAINS}")
+
+    // compile only
+    compileOnly("androidx.compose.ui:ui:${PosthogBuildConfig.Dependencies.ANDROIDX_COMPOSE}")
 
     // compatibility
     signature("org.codehaus.mojo.signature:java18:${PosthogBuildConfig.Plugins.SIGNATURE_JAVA18}@signature")
--- a/posthog-android/build.gradle.kts
+++ b/posthog-android/build.gradle.kts
@@ -89,6 +89,7 @@ dependencies {
     implementation("androidx.lifecycle:lifecycle-common-java8:${PosthogBuildConfig.Dependencies.LIFECYCLE}")
     implementation("androidx.core:core:${PosthogBuildConfig.Dependencies.ANDROIDX_CORE}")
     implementation("com.squareup.curtains:curtains:${PosthogBuildConfig.Dependencies.CURTAINS}")
+    implementation("androidx.compose.ui:ui-android:1.7.3")
 
     // compatibility
     signature("org.codehaus.mojo.signature:java18:${PosthogBuildConfig.Plugins.SIGNATURE_JAVA18}@signature")
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogMaskModifier.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogMaskModifier.kt
@@ -3,17 +3,21 @@ package com.posthog.android.replay
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.semantics.SemanticsPropertyKey
 import androidx.compose.ui.semantics.semantics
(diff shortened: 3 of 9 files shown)


buildSrc/src/main/java/PosthogBuildConfig.kt
```diff
@@ -14,7 +14,7 @@ object PosthogBuildConfig {
     }
 
     object Android {
-        val COMPILE_SDK = 33
+        val COMPILE_SDK = 34
```
marandaneto, line 17:
```suggestion
        val COMPILE_SDK = 33
```

Since 34 breaks compatibility with older versions, See https://github.com/PostHog/posthog-android/pull/98

posthog-android/build.gradle.kts
```diff
@@ -89,6 +89,7 @@ dependencies {
     implementation("androidx.lifecycle:lifecycle-common-java8:${PosthogBuildConfig.Dependencies.LIFECYCLE}")
     implementation("androidx.core:core:${PosthogBuildConfig.Dependencies.ANDROIDX_CORE}")
     implementation("com.squareup.curtains:curtains:${PosthogBuildConfig.Dependencies.CURTAINS}")
+    implementation("androidx.compose.ui:ui-android:1.7.3")
```
marandaneto, line 92: I think we should target the minimal version that has all the methods we need such as:

> import androidx.compose.ui.Modifier
> import androidx.compose.ui.semantics.SemanticsPropertyKey
> import androidx.compose.ui.semantics.semantics
> import androidx.compose.ui.node.RootForTest
> import androidx.compose.ui.semantics.SemanticsProperties
> import androidx.compose.ui.semantics.getAllSemanticsNodes
> import androidx.compose.ui.semantics.getOrNull

Dependency resolution will use the latest version anyway, but if we use a new one that breaks something and people depend on an older one, it's gonna be our SDK doing this.

posthog-android/src/main/java/com/posthog/android/replay/PostHogMaskModifier.kt
```diff
@@ -0,0 +1,20 @@
...
+import androidx.compose.ui.semantics.SemanticsPropertyKey
+import androidx.compose.ui.semantics.semantics
+
+public object PostHogMaskModifier {
+    internal val PostHogReplayMask = SemanticsPropertyKey<Boolean>("ph-no-capture")
+
+    /**
+     * Marks the element as not to be captured by PostHog Session Replay.
+     */
+    public fun Modifier.postHogMaskReplay(): Modifier {
```
marandaneto, line 13:
```suggestion
    public fun Modifier. postHogMask(isEnabled: Boolean = true): Modifier {
```

So that people can use the modifier using a flag instead of doing if/else when enabling or not this behavior.

marandaneto: If its `false`, then it should not be masked at all even if the other controls are enabled.

posthog-android/build.gradle.kts
marandaneto: Another option is to make this dependency `compileOnly`, and we check if its in the classpath before using its classes, probably with reflection as well (Class.forName)

marandaneto:
Nice @beradeep thanks for the PR
Left a few comments but that pretty much aligns with the solution I had in mind and is very similar to how it's done in SwiftUI as well.

posthog-android/src/main/java/com/posthog/android/replay/PostHogMaskModifier.kt
```diff
@@ -0,0 +1,20 @@
+package com.posthog.android.replay
+
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.SemanticsPropertyKey
+import androidx.compose.ui.semantics.semantics
+
+public object PostHogMaskModifier {
+    internal val PostHogReplayMask = SemanticsPropertyKey<Boolean>("ph-no-capture")
```
marandaneto, line 8: we can make a const for this ("ph-no-capture") as well since it's used many times

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -573,6 +582,55 @@ public class PostHogReplayIntegration(
...
+    ) {
+        val semanticsOwner =
+            (view as? RootForTest)?.semanticsOwner ?: run {
+                config.logger.log("View is not a RootForTest: $view")
+                return
+            }
+        val semanticsNodes = semanticsOwner.getAllSemanticsNodes(true)
+
+        semanticsNodes.forEach { node ->
+            val isTextInput = node.config.contains(SemanticsProperties.EditableText)
```
marandaneto, line 597: I think there are more options here such as `SemanticsProperties.Text`, etc

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -573,6 +582,55 @@ public class PostHogReplayIntegration(
...
+        return Rect(left.toInt(), top.toInt(), right.toInt(), bottom.toInt())
+    }
+
+    private fun View.isComposeView(): Boolean {
+        return isComposeAvailable && this.javaClass.name.contains(ANDROID_COMPOSE_VIEW)
+    }
+
+    private val isComposeAvailable by lazy(LazyThreadSafetyMode.PUBLICATION) {
+        try {
+            Class.forName(ANDROID_COMPOSE_VIEW_CLASS_NAME)
```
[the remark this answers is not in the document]
marandaneto: Otherwise we have to `-keepnames class androidx.compose.ui.platform.AndroidComposeView` in the proguard rules

posthog-android/build.gradle.kts
marandaneto: Unless we need something from `ui-android`, we could depend on the `androidx.compose.ui:ui`

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -553,6 +558,10 @@ public class PostHogReplayIntegration(
             }
         }
 
+        if (view.isComposeView()) {
+            findMaskableComposeRects(view, maskableWidgets)
+        }
```
marandaneto, line 563: I think all of the checks above and the one below (view.isNoCapture()) can be part of the else branch, so we don't check the other stuff unnecessarily, a view is either a compose view or not.

beradeep:
@marandaneto Thanks for the review! I'll try to incorporate all of the above.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
beradeep: But why do we need `SemanticProperties.Text`. For `maskAllTextInputs`, `SemanticsProperties.EditableText` is enough right?

marandaneto: `maskAllTextInputs` is for every type of text, whether it is a text (like just a label or something), an editable text (an edit text), etc.

beradeep: Ah, okay 👍

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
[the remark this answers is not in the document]
beradeep: Hi, I just tested this with on a release build, and it is working as intended.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
beradeep: I replaced it with a `when` block. I guess it'll make it a bit more readable too.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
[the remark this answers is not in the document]
marandaneto: did you test with minify enabled?

posthog-android/build.gradle.kts
marandaneto: the `compose` dependency is still `implementation` and not `compileOnly` so we still bring the dependency even if not using the modifier, have you tried to fix that?

marandaneto:
@beradeep sorry the delay I was on PTO, did another pass, almost there.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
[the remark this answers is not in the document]
beradeep: yes, I did test with these enabled on the sample app's `build.gradle`:
```
isMinifyEnabled = true
isShrinkResources = true
```

posthog-android/build.gradle.kts
beradeep: ah yes, I think we can simply change that to `compileOnly`

marandaneto: try to run the sample without any androidx.compose libs and enable the session reply just to be sure its compatible if someone isn't using compose at all, it should not crash or anything, if that works, we are gold.

marandaneto:
@beradeep add a new entry to https://github.com/PostHog/posthog-android/blob/main/CHANGELOG.md#next please :)

beradeep:
> @beradeep add a new entry to https://github.com/PostHog/posthog-android/blob/main/CHANGELOG.md#next please :)

ah yes, for sure 😄

posthog-android/build.gradle.kts
beradeep: I'll do it 👍

beradeep: just tried it right now. everything working normally :)

marandaneto: cool, i will run some tests tomorrow morning before merging and if all ok, that is DONE yay thanks for contributing.

Issue #189, opened 2024-10-07
Minimal duration support for Session replay


marandaneto:
### Description

https://posthog.com/docs/session-replay/how-to-control-which-sessions-you-record#minimum-duration

- [ ] Android
- [ ] iOS
- [ ] Flutter
- [ ] RN

karntrehan:
@marandaneto are there any good first issues to contribute to? I guess I can pick this up as well but would need some discussions with the team on the implementation - i.e. = If this should be a separate layer, if this should use an existing layer, what should be the default duration length, etc..

Looking to start contributing to the Android / Flutter SDK / BE of PostHog as we have been using it extensively across products. :)

marandaneto:
Hello @karntrehan, I am happy to hear that and glad to guide you through the implementation as well :)
The configuration will be read from the API here https://github.com/PostHog/posthog-android/blob/85f6ef1300919ac1ee5d9bf615c87e0fe79fb247/posthog/src/main/java/com/posthog/internal/PostHogFeatureFlags.kt#L127-L136
Most likely we will add this config here too https://github.com/PostHog/posthog-android/blob/main/posthog-android/src/main/java/com/posthog/android/replay/PostHogSessionReplayConfig.kt
See the JS implementation https://github.com/PostHog/posthog-js/blob/8e7e1ea6bc95199d9cc031ea975847cbc9209f55/src/types.ts#L386
Look for `minimumDurationMilliseconds`, `receivedMinimumDuration`, and `SESSION_RECORDING_MINIMUM_DURATION`.
I think a draft PR with making this work is a good start, by checking the JS SDK, and then we see how to structure, and organize the code and so on, WDYT?

karntrehan:
Hie @marandaneto,
I can see that API key `phc_QFbR1y41s5sxnNTZoyKG2NJo2RlsCIWkUfdpawgb40D` is added to the sample code itself [here](https://github.com/PostHog/posthog-android/blob/85f6ef1300919ac1ee5d9bf615c87e0fe79fb247/posthog-samples/posthog-android-sample/src/main/java/com/posthog/android/sample/MyApp.kt#L19). I hope that is not a concern. There are a bunch of other keys there as well. We could risk a DDoS? Not sure of the server side handling to prevent this. 

[...]

PR #199, opened 2024-10-17
Feature/189/session replay min duration

karntrehan:
#189 

WIP

## :pencil: Checklist

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -218,6 +218,8 @@ val config = PostHogAndroidConfig(apiKey).apply {
     sessionReplayConfig.screenshot = false
     // debouncerDelayMs is 500ms by default
     sessionReplayConfig.debouncerDelayMs = 1000
+    // minSessionDurationMs is 1000ms by default
+    sessionReplayConfig.minSessionDurationMs = 2000
 }
 ```
 
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -118,6 +118,10 @@ public class PostHogReplayIntegration(
     private val isSessionReplayEnabled: Boolean
         get() = PostHog.isSessionReplayActive()
 
+    private var sessionStartTime = 0L
+    private var sessionEndTime = 0L
+
+
     private fun addView(
         view: View,
         added: Boolean = true,
@@ -148,7 +152,11 @@ public class PostHogReplayIntegration(
 
                                         executor.submit {
                                             try {
-                                                generateSnapshot(WeakReference(decorView), WeakReference(window), timestamp)
+                                                generateSnapshot(
+                                                    WeakReference(decorView),
+                                                    WeakReference(window),
+                                                    timestamp
+                                                )
                                             } catch (e: Throwable) {
                                                 config.logger.log("Session Replay generateSnapshot failed: $e.")
                                             }
@@ -229,10 +237,19 @@ public class PostHogReplayIntegration(
                         }
                         when (motionEvent.action.and(MotionEvent.ACTION_MASK)) {
                             MotionEvent.ACTION_DOWN -> {
-                                generateMouseInteractions(timestamp, motionEvent, RRMouseInteraction.TouchStart)
+                                generateMouseInteractions(
+                                    timestamp,
+                                    motionEvent,
+                                    RRMouseInteraction.TouchStart
+                                )
                             }
+
                             MotionEvent.ACTION_UP -> {
-                                generateMouseInteractions(timestamp, motionEvent, RRMouseInteraction.TouchEnd)
+                                generateMouseInteractions(
+                                    timestamp,
+                                    motionEvent,
+                                    RRMouseInteraction.TouchEnd
+                                )
                             }
                         }
                     } catch (e: Throwable) {
@@ -265,7 +282,8 @@ public class PostHogReplayIntegration(
                     x = absX,
                     y = absY,
                 )
-            val mouseInteraction = RRIncrementalMouseInteractionEvent(mouseInteractionData, timestamp)
+            val mouseInteraction =
+                RRIncrementalMouseInteractionEvent(mouseInteractionData, timestamp)
             mouseInteractions.add(mouseInteraction)
         }
 
@@ -312,6 +330,9 @@ public class PostHogReplayIntegration(
 
         // workaround for react native that is started after the window is added
         // Curtains.rootViews should be empty for normal apps yet
+
+        sessionStartTime = config.dateProvider.currentTimeMillis()
+
         Curtains.rootViews.forEach { view ->
             addView(view)
         }
@@ -325,6 +346,13 @@ public class PostHogReplayIntegration(
 
     override fun uninstall() {
         try {
+            sessionEndTime = config.dateProvider.currentTimeMillis()
+
+            //TODO improve this
+            if (sessionEndTime - sessionStartTime <= config.sessionReplayConfig.minSessionDurationMs) {
+               //Delete existing session recording?
+            }
+
             Curtains.onRootViewsChangedListeners -= onRootViewsChangedListener
 
             decorViews.entries.forEach {
@@ -742,14 +770,17 @@ public class PostHogReplayIntegration(
                     style.verticalAlign = "center"
                     style.horizontalAlign = "center"
                 }
+
                 View.TEXT_ALIGNMENT_TEXT_END, View.TEXT_ALIGNMENT_VIEW_END -> {
                     style.verticalAlign = "center"
                     style.horizontalAlign = "right"
                 }
+
                 View.TEXT_ALIGNMENT_TEXT_START, View.TEXT_ALIGNMENT_VIEW_START -> {
                     style.verticalAlign = "center"
                     style.horizontalAlign = "left"
                 }
+
                 View.TEXT_ALIGNMENT_GRAVITY -> {
                     val horizontalAlignment =
                         when (view.gravity.and(Gravity.HORIZONTAL_GRAVITY_MASK)) {
@@ -769,6 +800,7 @@ public class PostHogReplayIntegration(
                         }
                     style.verticalAlign = verticalAlignment
                 }
+
                 else -> {
                     style.verticalAlign = "center"
                     style.horizontalAlign = "left"
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -119,8 +119,7 @@ public class PostHogReplayIntegration(
         get() = PostHog.isSessionReplayActive()
 
     private var sessionStartTime = 0L
-    private var sessionEndTime = 0L
-
+    private val events = mutableListOf<RREvent>()
 
     private fun addView(
         view: View,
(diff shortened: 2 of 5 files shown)


karntrehan:
@marandaneto could you please review the implementation? Thanks! Would then go ahead with the test cases.

karntrehan:
Hie @marandaneto just following up on this PR.

marandaneto:
> Hie @marandaneto just following up on this PR. 

Sorry i am on pto this week so a bit slow to review, maybe @ioannisj or @pauldambra can take a look for now? Otherwise i will have a look within the next couple days

karntrehan:
No problem. Take your time @marandaneto. Also, please feel free to mention me in other issues that you think I can contribute to. I have kept some time aside to make the hog hoggier!

USAGE.md
```diff
@@ -218,6 +218,8 @@ val config = PostHogAndroidConfig(apiKey).apply {
     sessionReplayConfig.screenshot = false
     // debouncerDelayMs is 500ms by default
     sessionReplayConfig.debouncerDelayMs = 1000
+    // minSessionDurationMs is 1000ms by default
+    sessionReplayConfig.minSessionDurationMs = 2000
```
marandaneto, line 222: I think the default should be null or 0, which means no min session duration

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -148,7 +158,11 @@ public class PostHogReplayIntegration(
 
                                         executor.submit {
                                             try {
-                                                generateSnapshot(WeakReference(decorView), WeakReference(window), timestamp)
+                                                generateSnapshot(
```
marandaneto, line 161: Its better if you do formatting in a different PR so I can review only real changes

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -312,6 +336,9 @@ public class PostHogReplayIntegration(
 
         // workaround for react native that is started after the window is added
         // Curtains.rootViews should be empty for normal apps yet
+
+        sessionStartTime = config.dateProvider.currentTimeMillis()
```
marandaneto, line 340: Here is just the setup, the session actually only starts when the first snapshot is captured, and that can be a console log, a network log, or a screenshot/wireframe snapshot.
Search for the usage of this extension https://github.com/PostHog/posthog-android/blob/351a86b7c7caf9c72aad684d3721fad2872cfa2c/posthog/src/main/java/com/posthog/internal/replay/RRUtils.kt#L7

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -453,13 +479,35 @@ public class PostHogReplayIntegration(
...
 
         status.lastSnapshot = wireframe
     }
 
+    private fun sessionLongerThanMinDuration(): Boolean {
+        //Check value only if threshold not crossed.
+        if (!minSessionThresholdCrossed) {
+            val serverMinDuration = replayPreferenceMap?.let { map ->
+                (map["minimumDurationMilliseconds"] as Number).toLong()
+            } ?: 0L
```
marandaneto, line 496: Instead of reading from `replayPreferenceMap` (the disk), we can just update the `config.sessionReplayConfig.minSessionDurationMs` once, when the SDK is init, we do something similar with `snapshotEndpoint`

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -853,7 +853,8 @@ public class PostHog private constructor(
     // this is used in cases where we know the session is already active
     // so we spare another locker
     private fun isSessionReplayFlagActive(): Boolean {
-        return config?.sessionReplay == true && featureFlags?.isSessionReplayFlagActive() == true
+        //FIXME -> Remove this true hardcoding
+        return true || config?.sessionReplay == true && featureFlags?.isSessionReplayFlagActive() == true
```
marandaneto, line 857: rollback

posthog/src/main/java/com/posthog/internal/PostHogPreferences.kt
```diff
@@ -33,7 +33,7 @@ public interface PostHogPreferences {
         internal const val OPT_OUT = "opt-out"
         internal const val FEATURE_FLAGS = "featureFlags"
         internal const val FEATURE_FLAGS_PAYLOAD = "featureFlagsPayload"
-        internal const val SESSION_REPLAY = "sessionReplay"
+        public const val SESSION_REPLAY: String = "sessionReplay"
```
marandaneto, line 36: rollback when https://github.com/PostHog/posthog-android/pull/199/files#r1819028697 is addressed most likely

marandaneto:
> No problem. Take your time @marandaneto. Also, please feel free to mention me in other issues that you think I can contribute to. I have kept some time aside to make the hog hoggier!

@karntrehan did another pass, let me know if something is unclear, I understand some parts might be tricky

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -118,6 +119,15 @@ public class PostHogReplayIntegration(
     private val isSessionReplayEnabled: Boolean
         get() = PostHog.isSessionReplayActive()
 
+    private var sessionStartTime = 0L
+    private val events = mutableListOf<RREvent>()
+    private var minSessionThresholdCrossed = false
```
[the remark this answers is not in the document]
karntrehan: This is not very clear to me. What we are doing here is setting the threshold once for 1 launch session of the user. 

The behavior we have currently is: User launches the app, we start caching events & wait for the minimum threshold, if the user closes the app before the threshold the events are not captured / dropped. Once the threshold is crossed, all cached events are captured immediately and subsequent events are captured right away. Is this the expected behavior? 

I am not sure of ViewTreeSnapshotStatus and where the session_id is set / unset. Could you help me with more info on this?

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
karntrehan: I have made this change as per my understanding. 

On every launch `preloadSessionReplayFlag` is taking precedence over `loadFeatureFlags` hence the server value is taking 1 launch delta to be updated. 
Is this expected as well? This wasn't happening in the preference lookup. 

Do let me know if my understanding is incorrect here.

karntrehan:
@marandaneto I have made some changes basis your feedback. Please check. 

It has helped me understand the code flow better. Thanks!

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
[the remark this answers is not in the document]
marandaneto: So all this information should be related to a specific sessionId.
Right now the `sessionStartTime` is gonna be used even if the session rotates, etc.
So we need a way of Map<SessionId, Object> where `Object` contains the properties (sessionStartTime, events, mouseInteractions, etc...)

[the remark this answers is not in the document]
marandaneto: An example is that if the app is in the background for more than 30 minutes, the sessionId rotates, and the new session will start once the app is in the background, but we are still using the `sessionStartTime` from the last session.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
marandaneto: Yes, it can happen that it will be using the last min duration and not the new one immediately but I will correct it as soon as the flags are loaded again, which is fine, its better than reading from the disk every time (too much IO), or awaiting a request before being able to use the feature, it's a trade-off we made.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -269,12 +274,14 @@ public class PostHogReplayIntegration(
...
 
-        if (mouseInteractions.isNotEmpty()) {
+        if (mouseInteractions.isNotEmpty() && sessionLongerThanMinDuration()) {
             // TODO: we can probably batch those
             // if we batch them, we need to be aware that the order of the events matters
             // also because if we send a mouse interaction later, it might be attached to the wrong
             // screen
+            config.logger.log("Session replay mouse events captured: " + mouseInteractions.size)
             mouseInteractions.capture()
+            mouseInteractions.clear()
```
marandaneto, line 284: I think we will face a problem here that is:
If a mouse interaction is captured but is not within the min duration and no other interaction is captured after the min duration, the interactions before the min duration are never going to be sent, so we have to call `mouseInteractions.capture()` somewhere as well to be sure that the list is flushed, eg in the `generateSnapshot` method as well.
Most likely we'd need a lock since `generateSnapshot` is executed on a background thread and `generateMouseInteractions` as well so mutations can't race.

posthog-android/src/main/java/com/posthog/android/replay/PostHogSessionReplayConfig.kt
```diff
@@ -48,4 +48,11 @@ public class PostHogSessionReplayConfig
          */
         @PostHogExperimental
         public var debouncerDelayMs: Long = 500,
+        /**
+         * Define the minimum duration for sessions to be recorded.
+         * This is useful if you want to exclude sessions that are too short to be useful.
+         * Defaults to 0ms
+         */
+        @PostHogExperimental
+        public var minSessionDurationMs: Long = 0,
```
marandaneto, line 57:
```suggestion
        public var minSessionDurationMs: Long? = null,
```

I think nullable makes more sense than 0, which means no limits

marandaneto:
@karntrehan left a few new comments, a few things to address but it's in the right direction.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
karntrehan: Understood. Will keep this change then.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
karntrehan: Ah. Good catch. Let me figure something out here.

karntrehan: Have taken a different approach here, where on uninstall I am flushing the events if the threshold was crossed.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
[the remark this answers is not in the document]
karntrehan: Got it. Looking at PostHogSessionManager right now. Thanks!

marandaneto:
@karntrehan any intention and time to finish this? :) thanks!

marandaneto:
Closing the issue as a part of large repository cleanup, due to it being inactive and/or outdated.
Please do not hesitate to ping me if it is still relevant, and I will happily reopen and work on it.
Cheers!

PR #201, opened 2024-10-25
recording: fix crash when calling view.isVisible

marandaneto:
Relates to https://github.com/PostHog/posthog-react-native-session-replay/issues/10

Running method out of the main thread
It should not crash from the main thread anyway but for some reason, it does for that issue

## :pencil: Checklist

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -7,6 +7,7 @@ import android.graphics.Bitmap
 import android.graphics.Canvas
 import android.graphics.Color
 import android.graphics.Paint
+import android.graphics.Point
 import android.graphics.PorterDuff
 import android.graphics.Rect
 import android.graphics.RectF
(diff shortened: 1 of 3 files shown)


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -460,20 +459,45 @@ public class PostHogReplayIntegration(
...
-        val visible = isShown && width >= 0 && height >= 0 && this !is ViewStub
-
-        // Between API 16 and API 29, this method may incorrectly return false when magnification
-        // is enabled. On other versions, a node is considered visible even if it is not on
-        // the screen because magnification is active.
-        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.Q) {
-            return visible
-        }
+        try {
+            if (isAttachedToWindow) {
```
marandaneto, line 467: this is vendoring the most important part of `onInitializeAccessibilityNodeInfo`
but we don't have control over the internal impl nor do I know if it requires the main thread or not since I cannot reproduce the original issue, so I copied and adapted and adapted a try-catch just in case

Issue #203, opened 2024-11-04
Note that observation on multiple threads in layout/draw is not supported. Make sure your measure/layout/draw for each Owner (AndroidComposeView) is executed on the same thread


xemniz:
### Version

3.9.0

### Steps to Reproduce

1. I use compose in fragments a lot 
2. It's really ocassional but im just switching tab bars, scrolling lazylists
3. App just crashes
4. previousThreadName from the stacktrace is PostHogReplayThread
5. when i switch off the replay, can't reproduce this

### Expected Result

App works

### Actual Result

java.lang.IllegalArgumentException: Detected multithreaded access to  : previousThreadId=158), currentThread={id=2, name=main}. Note that observation on multiple threads in layout/draw is not supported. Make sure your measure/layout/draw for each Owner (AndroidComposeView) is executed on the same thread.
                                                                                                    	at androidx.compose.runtime.PreconditionsKt.throwIllegalArgumentException(Preconditions.kt:26)
                                                                                                    	at androidx.compose.runtime.snapshots.SnapshotStateObserver.observeReads(SnapshotStateObserver.kt:708)
                                                                                                    	at androidx.compose.ui.node.OwnerSnapshotObserver.observeReads$ui_release(OwnerSnapshotObserver.kt:133)
                                                                                                    	at androidx.compose.ui.node.OwnerSnapshotObserver.observeMeasureSnapshotReads$ui_release(OwnerSnapshotObserver.kt:113)

marandaneto:
Hello @xemniz thanks for the issue.
Are you using the screenshot or wireframe mode?
Which SDK version?
We don't call `measure` and `layout` at all but we do use `draw` [here](https://github.com/PostHog/posthog-android/blob/d9052d0232d79211c73b2f5fd972e4e5efac2b27/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt#L1231) which should not be a problem anyway.
Can you provide a minimal reproducible example? I tried using a compose-sample app but did not succeed in reproducing the issue.

xemniz:
<img width="485" alt="image" src="https://github.com/user-attachments/assets/adfd197f-8c26-4ed5-b9b2-6ad95a07f0b6">
this commented lines made a difference, it was reproducible in all my screens, but randomly i'd say. in crashlytics crashfree sessions dropped to 65 percent with 3.9.0 version of posthog.

[...]

PR #204, opened 2024-11-11
recording: fix observation on multiple threads in layout/draw is not supported for compose

marandaneto:
Tries to close https://github.com/PostHog/posthog-android/issues/203

Running the sample with compose screen

## :pencil: Checklist

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -607,43 +607,59 @@ public class PostHogReplayIntegration(
         view: View,
         maskableWidgets: MutableList<Rect>,
     ) {
-        try {
-            val semanticsOwner =
-                (view as? RootForTest)?.semanticsOwner ?: run {
-                    config.logger.log("View is not a RootForTest: $view")
-                    return
-                }
-            val semanticsNodes = semanticsOwner.getAllSemanticsNodes(true)
+        val latch = CountDownLatch(1)
 
-            semanticsNodes.forEach { node ->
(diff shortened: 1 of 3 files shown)


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -607,43 +607,59 @@ public class PostHogReplayIntegration(
...
         maskableWidgets: MutableList<Rect>,
     ) {
-        try {
-            val semanticsOwner =
-                (view as? RootForTest)?.semanticsOwner ?: run {
-                    config.logger.log("View is not a RootForTest: $view")
-                    return
-                }
-            val semanticsNodes = semanticsOwner.getAllSemanticsNodes(true)
+        val latch = CountDownLatch(1)
```
marandaneto, line 610: given that this method is only executed when `view.isComposeView()`, its fine to do thread jumping since its a requirement for compose anyway.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -607,43 +607,59 @@ public class PostHogReplayIntegration(
...
+                // swallow possible errors due to compose versioning, etc
+                config.logger.log("Session Replay findMaskableComposeWidgets (main thread) failed: $e")
+            } finally {
+                latch.countDown()
             }
+        }
+
+        try {
+            // await for 1s max
+            latch.await(1000, TimeUnit.MILLISECONDS)
```
marandaneto, line 660: just an arbitrary value since it is a background thread, awaiting time wont block the screen, the time-sensitive is the part within the main thread block

PR #214, opened 2024-12-13
chore: change screenshots debouncing approach to throttling

marandaneto:
Using throttling instead of debouncing, similar to iOS and Flutter now.

## :pencil: Checklist


marandaneto:
@ioannisj would be cool to do the same on iOS if not throttling yet

pauldambra (approved):
👍

Issue #216, opened 2025-01-27
Initing the SDK with the `with` method will break session replay and autocapture


marandaneto:
### Version

3.x

### Steps to Reproduce

start the SDK with the `with` method instead of the `setup` method and do not call the `setup` method.
Relates to https://posthoghelp.zendesk.com/agent/tickets/23028
The reason is that we capture autocapture and snapshots events with the static instance but the static instance isn't set when you start the SDK with the `with` method

### Expected Result

the SDK should work as expected

### Actual Result

no autocapture and snapshots events are captured

marandaneto:
@ioannisj i think this can be an issue with iOS as well, can you take a look?

PR #217, opened 2025-01-27
fix: session replay and auto capture works with 'with' method

marandaneto:
Closes https://github.com/PostHog/posthog-android/issues/216

## :pencil: Checklist

### DIFF
--- a/posthog-android/api/posthog-android.api
+++ b/posthog-android/api/posthog-android.api
@@ -64,7 +64,7 @@ public final class com/posthog/android/replay/PostHogReplayIntegration : com/pos
 	public static final field ANDROID_COMPOSE_VIEW_CLASS_NAME Ljava/lang/String;
 	public static final field PH_NO_CAPTURE_LABEL Ljava/lang/String;
 	public fun <init> (Landroid/content/Context;Lcom/posthog/android/PostHogAndroidConfig;Lcom/posthog/android/internal/MainHandler;)V
-	public fun install ()V
+	public fun install (Lcom/posthog/PostHogInterface;)V
 	public fun uninstall ()V
 }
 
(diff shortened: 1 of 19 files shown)


posthog/src/main/java/com/posthog/PostHogOkHttpInterceptor.kt
```diff
@@ -6,66 +6,78 @@ import okhttp3.Interceptor
...
-            val payload = mapOf<String, Any>("requests" to requests)
+        val events = listOf(RRPluginEvent("rrweb/network@1", payload, end))
 
-            listOf(RRPluginEvent("rrweb/network@1", payload, end)).capture()
+        // its not guaranteed that the posthog instance is set
+        if (postHog != null) {
+            events.capture(postHog)
+        } else {
+            // use static instance
+            events.capture()
```
marandaneto, line 80: important part is here

posthog-android/api/posthog-android.api
```diff
@@ -64,7 +64,7 @@ public final class com/posthog/android/replay/PostHogReplayIntegration : com/pos
 	public static final field ANDROID_COMPOSE_VIEW_CLASS_NAME Ljava/lang/String;
 	public static final field PH_NO_CAPTURE_LABEL Ljava/lang/String;
 	public fun <init> (Landroid/content/Context;Lcom/posthog/android/PostHogAndroidConfig;Lcom/posthog/android/internal/MainHandler;)V
-	public fun install ()V
+	public fun install (Lcom/posthog/PostHogInterface;)V
```
marandaneto, line 67: technically a breaking change but we don't document this integration API anywhere so its ok (we only use intergrations internally)

pauldambra (approved):
🙌

posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
```diff
@@ -39,7 +41,10 @@ internal class PostHogActivityLifecycleCallbackIntegration(
                 } finally {
                     data?.let { props["url"] = it.toString() }
                     intent.getReferrerInfo(config).let { props.putAll(it) }
-                    PostHog.capture("Deep Link Opened", properties = props)
+
+                    if (props.isNotEmpty()) {
+                        postHog?.capture("Deep Link Opened", properties = props)
+                    }
```
marandaneto, line 47: fly by fix, no event should be captured if we don't know where it comes from

PR #219, opened 2025-01-28
fix: do not allow default integrations multiple times

marandaneto:
## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+- fix: do not allow default integrations multiple times ([#218](https://github.com/PostHog/posthog-flutter/pull/218))
+
 ## 3.11.0 - 2025-01-27
 
 - chore: Session Replay - GA
--- a/posthog-android/src/main/java/com/posthog/android/PostHogAndroid.kt
+++ b/posthog-android/src/main/java/com/posthog/android/PostHogAndroid.kt
@@ -87,6 +87,7 @@ public class PostHogAndroid private constructor() {
             if (context is Application) {
                 if (config.captureDeepLinks || config.captureScreenViews || config.sessionReplay) {
                     config.addIntegration(PostHogActivityLifecycleCallbackIntegration(context, config))
+                    config.addIntegration(PostHogActivityLifecycleCallbackIntegration(context, config))
                 }
             }
             if (config.captureApplicationLifecycleEvents) {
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,14 +1,16 @@
 ## Next
 
+- fix: do not allow default integrations multiple times ([#219](https://github.com/PostHog/posthog-android/pull/219))
+
 ## 3.11.0 - 2025-01-27
 
 - chore: Session Replay - GA
-- fix: session replay and auto capture works with 'with' method ([#217](https://github.com/PostHog/posthog-flutter/pull/217))
-- fix: sending cached events null check ([#218](https://github.com/PostHog/posthog-flutter/pull/218))
+- fix: session replay and auto capture works with 'with' method ([#217](https://github.com/PostHog/posthog-android/pull/217))
```
marandaneto, line 8: ups, copy-pasta error

marandaneto:
> I’m not convinced if this is the right approach tbh.
> 
> Integrations are effectively handled in a first-come first-served basis, so instance creation order matters here, which could be problematic. If instance creation happens in a non-deterministic order as well, then this will lead to unpredictable results down the line.
> 
> Consider this scenario:
> 
> * HostApp integrates with PostHogSDK.
> * HostApp also integrates with SomeOtherSDK.
> * SomeOtherSDK, in turn, integrates with PostHogSDK.
> * Both HostApp and SomeOtherSDK enable all integrations in config (session replay, autocapture, etc.).
> 
> In this setup, whichever SDK calls .setup() or .with() first takes precedence. This could introduce inconsistent behavior if the call order is not guaranteed—sometimes the HostApp gets all the active integrations, while other times SomeOtherSDK does.
> 
> Broadly speaking, I feel this ties back to how we handle multiple PostHog SDK instances.
> 
> Maybe a more robust solution would be to introduce a mechanism in the instance creation process that will ensure a single PostHogSDK instance **per API key**, each with its own configuration and set of active integrations:
> 
> * Calling .with(apiKey: "project_one_api_key") multiple times should always return the same PostHogSDK instance, ensuring that integrations are installed only once and that events, session replays, and autocapture are delivered to the "project_one_api_key" project.
> * Similarly, calling .with(apiKey: "project_two_api_key") should create a separate instance for that project with its own set of integrations, sending data specifically to the "project_two_api_key" project.
> 
> In the scenario above, HostAp and SomeOtherSDK will get a seprate PostHogSDK instance, with all integrations active for each.
> 
> wdyt?

discussed this offline.

marandaneto:
@ioannisj [`dda9648` (#219)](https://github.com/PostHog/posthog-android/pull/219/commits/dda96488ed82b798d207e1e544e5b1327c1976ca)

marandaneto:
@ioannisj @PostHog/team-replay can I *has* a review?

pauldambra (approved):
low context 👍

PR #225, opened 2025-02-20
chore(java-sdk): refactor core interface to support stateless server-side api

havenbarnes:

This PR is the first of a handful of a refactor to allow the core Kotlin code to support stateless server-side Java applications, eventually adding a new version of the Java SDK into this repo just like https://github.com/PostHog/posthog-js-lite

Note: No functional changes to the Android SDK are expected with these changes

No new server-side code here, so existing Android tests should suffice for confidence in these changes.

## :pencil: Checklist

### DIFF
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -44,7 +44,7 @@ public class PostHog private constructor(
             PostHogThreadFactory("PostHogSendCachedEventsThread"),
         ),
     private val reloadFeatureFlags: Boolean = true,
-) : PostHogInterface {
+) : PostHogInterface, PostHogStateless() {
     @Volatile
     private var enabled = false
 
@@ -85,7 +85,8 @@ public class PostHog private constructor(
                 config.cachePreferences = cachePreferences
                 val api = PostHogApi(config)
                 val queue = PostHogQueue(config, api, PostHogApiEndpoint.BATCH, config.storagePrefix, queueExecutor)
-                val replayQueue = PostHogQueue(config, api, PostHogApiEndpoint.SNAPSHOT, config.replayStoragePrefix, replayExecutor)
+                val replayQueue =
+                    PostHogQueue(config, api, PostHogApiEndpoint.SNAPSHOT, config.replayStoragePrefix, replayExecutor)
                 val featureFlags = PostHogFeatureFlags(config, api, featureFlagsExecutor)
 
                 // no need to lock optOut here since the setup is locked already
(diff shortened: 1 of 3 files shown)


posthog/src/main/java/com/posthog/PostHogInterface.kt
```diff
@@ -2,10 +2,7 @@ package com.posthog
 
 import java.util.UUID
 
-/**
- * The PostHog SDK entry point
- */
-public interface PostHogInterface {
+public interface PostHogCoreInterface {
```
marandaneto, line 5: lets create new files for new classes/interfaces (usually we don't do multiple things in one class in java world)

havenbarnes:
> since there will be multiple PRs, lets create a new branch? its likely a major bump.

Good call, will start a chain yes

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -894,11 +765,6 @@ public class PostHog private constructor(
         return PostHogSessionManager.getActiveSessionId()
     }
 
-    override fun <T : PostHogConfig> getConfig(): T? {
```
havenbarnes, line 897: Since `config` is a class property, this getter is generated already (and this was causing a ` Inherited platform declarations clash: The following declarations have the same JVM signature (getConfig()Lcom/posthog/PostHogConfig;)` error

posthog/src/main/java/com/posthog/internal/PostHogFeatureFlagsInterface.kt
```diff
@@ -0,0 +1,33 @@
+package com.posthog.internal
+
+import com.posthog.PostHogOnFeatureFlags
+
+public interface PostHogFeatureFlagsInterface {
```
havenbarnes, line 5: These public interfaces are needed now to satisfy the compiler due to internal modules like `com.posthog.internal.PostHogFeatureFlags` being exposed through inheritance from `PostHogStateless` -> `PostHog`

posthog/src/main/java/com/posthog/PostHogStateless.kt
```diff
@@ -0,0 +1,637 @@
...
+
+    /**
+     * Is Opt Out
+     */
+    public override fun isOptOut(): Boolean {
+        if (!isEnabled()) {
+            return true
+        }
+        return config?.optOut ?: true
+    }
```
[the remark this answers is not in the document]
marandaneto: i think a good example is also a staging/production env, where you can option and out depending on the env, so its useful even as stateless.
some other SDKs have extra flags such as `enabled`, which is IMO the same as optin-out so I'd rather keep in the stateless class as well

marandaneto:
moving to draft since this is outdated

PR #226, opened 2025-02-20
fix: getFeatureFlagPayload wasn't sending the  event

marandaneto:
https://github.com/PostHog/posthog-flutter/issues/161 for Android
iOS would need to do the same.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next
 
-- fix: getFeatureFlagPayload wasn't sending the $feature_flag_called event ([#185](https://github.com/PostHog/posthog-android/pull/185))
+- fix: getFeatureFlagPayload wasn't sending the $feature_flag_called event ([#226](https://github.com/PostHog/posthog-android/pull/226))
 
 ## 3.11.2 - 2025-02-04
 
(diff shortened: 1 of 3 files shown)


marandaneto:
before merging this, I'd like to ask if there's a reason why this is not done on the JS SDK either, since we ported the code from there.
https://github.com/PostHog/posthog-js/blob/1a5ce0bf6a7f77c44499de792ad4ec7aa45b8aa5/src/posthog-featureflags.ts#L358-L361

rafaeelaudibert:
@marandaneto I believe it's intentional in posthog-js. You can see that we actually use `getFeatureFlagPayload()` inside the code that actually sends the `$feature_flag_called` event: https://github.com/PostHog/posthog-js/blob/1a5ce0bf6a7f77c44499de792ad4ec7aa45b8aa5/src/posthog-featureflags.ts#L346. We definitely don't want two captures in that case.

This is all very intricate in the `posthog-js` library, we might need to decouple that a little to figure out how this should behave. It sucks that our SDKs aren't consistent - something the sdk-team-whom-shouldnt-be-mentioned could work on.

marandaneto:
> $feature_flag_payload

that is new, not added to all SDKs yet

marandaneto:
Closing it until @PostHog/team-feature-flags tell what's the way forward here

PR #228, opened 2025-02-22
feat(feature-flags): support quota limiting for feature flags

dmarticus:
with https://github.com/PostHog/posthog/pull/28564, we now start to respond different with the /decide and /local_evaluation APIs if users have gone over their quota limit. Now we need to change the SDKs to handle these new responses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next
 
-## 3.11.3 - 2025-02-24
+## 3.11.3 - 2025-02-25
 
 - feat: support quota limiting for feature flags ([#228](https://github.com/PostHog/posthog-android/pull/228))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next
 
-## 3.11.3 - 2025-02-21
+## 3.11.3 - 2025-02-24
 
 - feat: support quota limiting for feature flags ([#228](https://github.com/PostHog/posthog-android/pull/228))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Next
 
+## 3.11.3 - 2025-02-21
+
+- feat: support quota limiting for feature flags ([#228](https://github.com/PostHog/posthog-android/pull/228))
+
 ## 3.11.2 - 2025-02-04
 
 - fix: touches fall back to single touches if out of bounds ([#221](https://github.com/PostHog/posthog-android/pull/221))
--- a/posthog/src/main/java/com/posthog/internal/PostHogDecideResponse.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogDecideResponse.kt
@@ -17,5 +17,5 @@ internal data class PostHogDecideResponse(
     val featureFlagPayloads: Map<String, Any?>?,
     // its either a boolean or a map, see https://github.com/PostHog/posthog-js/blob/10fd7f4fa083f997d31a4a4c7be7d311d0a95e74/src/types.ts#L235-L243
     val sessionRecording: Any? = false,
-    val quotaLimited: Array<String>? = null,
+    val quotaLimited: List<String>? = null,
 )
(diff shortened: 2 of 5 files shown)


marandaneto (changes requested):
lets figure out https://github.com/PostHog/posthog-js-lite/pull/403#discussion_r1966462890 first before merging this

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Next
 
+## 3.11.3 - 2025-02-21
```
marandaneto, line 3:
```suggestion
## 3.11.3 - 2025-02-24
```

posthog/src/main/java/com/posthog/internal/PostHogDecideResponse.kt
```diff
@@ -16,4 +17,5 @@ internal data class PostHogDecideResponse(
     val featureFlagPayloads: Map<String, Any?>?,
     // its either a boolean or a map, see https://github.com/PostHog/posthog-js/blob/10fd7f4fa083f997d31a4a4c7be7d311d0a95e74/src/types.ts#L235-L243
     val sessionRecording: Any? = false,
+    val quotaLimited: Array<String>? = null,
```
marandaneto, line 20:
```suggestion
    val quotaLimited: List<String>? = null,
```

more common to use the List interface

posthog/src/main/java/com/posthog/internal/PostHogFeatureFlags.kt
```diff
@@ -89,6 +89,17 @@ internal class PostHogFeatureFlags(
...
                 response?.let {
                     synchronized(featureFlagsLock) {
+                        if (response.quotaLimited?.contains("feature_flags") == true) {
+                            config.logger.log("Feature flags are quota limited, clearing existing flags")
+                            this.featureFlags = null
+                            this.featureFlagPayloads = null
+                            config.cachePreferences?.let { preferences ->
+                                preferences.remove(FEATURE_FLAGS)
+                                preferences.remove(FEATURE_FLAGS_PAYLOAD)
+                            }
```
marandaneto, line 99: what should happen if session replay uses a feature flag and now is quota limited, should it stop working as well? if so, I guess we can also do `preferences.remove(SESSION_REPLAY)` if session replay has a `linkedFlag` value.

marandaneto (commented):
left a few comments otherwise all good

marandaneto:
@dmarticus this should be done for iOS as well right? is your team doing it? or do you need help from @ioannisj

dmarticus:
> @dmarticus this should be done for iOS as well right? is your team doing it? or do you need help from @ioannisj

Yup, I did it here: https://github.com/PostHog/posthog-ios/pull/308, tagged you and Ioannis for review :)

posthog/src/main/java/com/posthog/internal/PostHogFeatureFlags.kt
dmarticus: Oh, yeah, I didn't think of that scenario – can you tell me more about how session replay with flags works? I'm fine removing it but I don't really understand how session replay and flags overlap in this context.

pauldambra: well, it should fail closed 

generally speaking folk are setting a linked flag to reduce cost/volume. so failing open would be nasty

so, I have a linked flag, but I am past flag quota, decide/config continues to return the linked flag for replay, the SDK never shows it as received because i'm past quota, and replay never starts

pauldambra: then i think the UX is all around how we notify people about being past quota. feels way more important for flags than other products

pauldambra: yep, at least failing closed :)

ideally we'd register a super property on events if replay or flags or etc are past their quota for debugging (obvs doesn't help if events are past quota)

but yep, the onward journey will be the thing

marandaneto:
@dmarticus should I merge and release this?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Next
 
+## 3.11.3 - 2025-02-24
```
marandaneto, line 3:
```suggestion
## 3.11.3 - 2025-02-25
```

PR #229, opened 2025-02-24
allow optional reuse of anonymous id

### LINKED ISSUE #224: Give an option to reuse the anonId between user changes instead of creating a new anonId every time
### Description

https://posthog.com/questions/how-to-persist-anonymous-id-in-post-hog-for-guest-mode-in-an-android-app

beradeep:

Fix #224 

Added unit test.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+- feat: support reuse of `anonymousId` between user changes ([#229](https://github.com/PostHog/posthog-android/pull/229))
+
 ## 3.11.3 - 2025-02-26
 
 - feat: support quota limiting for feature flags ([#228](https://github.com/PostHog/posthog-android/pull/228) and [#230](https://github.com/PostHog/posthog-android/pull/230))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Next
 
+## 3.11.3 - 2025-02-26
+
+- feat: support quota limiting for feature flags ([#228](https://github.com/PostHog/posthog-android/pull/228) and [#230](https://github.com/PostHog/posthog-android/pull/230))
+
 ## 3.11.2 - 2025-02-04
 
 - fix: touches fall back to single touches if out of bounds ([#221](https://github.com/PostHog/posthog-android/pull/221))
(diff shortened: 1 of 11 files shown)


marandaneto:
@beradeep thanks
i will review later this week

marandaneto:
I dont think this will work.
https://github.com/PostHog/posthog-android/blob/6185d73c5dffe08dbbf83e6d7709d1b9cba0a7e8/posthog/src/main/java/com/posthog/PostHog.kt#L552-L557
once you `identify` an user, the given `distinctId` and the `anonymousId` are merged within a single user.

if you `reset` and keep the `anonymousId`, once you `identify` another user, now the older `distinctId`, the `anonymousId` and the new `distinctId` will also be merged within the same user.

Otherwise there will never be a "guest mode".
I think the approach we should take is that there's a specific `anonymousId` guest mode, similar to a`$device_id` actually, its always the same ID when the user is logged out.

for example:
```
val guestId = 123
// this is also the current anonymousId
posthog.identify(456)
// this should not merge user 123 and 456 (but it is), so we should not set `props["\$anon_distinct_id"] = anonymousId`
nor pass `featureFlags?.loadFeatureFlags(distinctId, anonymousId = anonymousId, groups, onFeatureFlags)` if `reuseAnonymousId` is enabled.
posthog.reset()
// now yes we don't delete the cached anonymousId, this is correct.
posthog.identify(789)
```

does that make sense?

beradeep:
Yes, it does make sense. I missed out on that part; we shouldn't merge all users on a device. Thanks for such a nice explanation.

marandaneto:
@beradeep i think we should not allow this as well https://github.com/PostHog/posthog-android/blob/6185d73c5dffe08dbbf83e6d7709d1b9cba0a7e8/posthog/src/main/java/com/posthog/PostHog.kt#L577
when the new flag is enabled, otherwise we will lose the anon id (guest mode).

marandaneto:
@beradeep looks good, i added a more detailed explanation [here](https://github.com/PostHog/posthog-android/pull/229/commits/4e40cc93b71144f5afaa2f3e9de1710d51052672)
can you add unit test for the last changes? right now we are only testing the `reset` changes.

marandaneto (approved):
missing 2 tests otherwise LGTM

PR #235, opened 2025-03-25
feat: use network time instead of device time

### LINKED ISSUE #223: Use network time instead of device time
### Description

- [ ] Android
- [ ] iOS
- [ ] Flutter
- [ ] RN

Alternatives: https://android-developers.googleblog.com/2025/02/trustedtime-api-introducing-reliable-approach-to-time-keeping-for-apps.html?m=1
https://developer.android.com/reference/android/os/SystemClock#currentNetworkTimeClock() (API 33)
https://github.com/instacart/truetime-android

Analytics data, recordings, etc, all depend on the timestamp. The server does clock drifting if it's off in the past or future, but that's not ideal.
Devices can be way in the past, future or just off by a few minutes due to user's settings, I believe using a network time would be way better

beradeep:

Fix #223 

Tested it alongside existing device provider, works fine.

## :pencil: Checklist


buildSrc/src/main/java/PosthogBuildConfig.kt
```diff
@@ -56,6 +56,7 @@ object PosthogBuildConfig {
         val CURTAINS = "1.2.5"
         val ANDROIDX_CORE = "1.5.0"
         val ANDROIDX_COMPOSE = "1.0.0"
+        val TRUE_TIME = "4.0.0.alpha"
```
marandaneto, line 59: most likely we dont want to bring a new dep that is alpha, also it really depends on the size of the library because of bundle size.

marandaneto: can we check the other options instead?

posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidDateProvider.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import com.instacart.truetime.time.TrueTimeImpl
+import com.posthog.internal.PostHogDateProvider
+import java.util.Date
+
+internal class PostHogAndroidDateProvider : PostHogDateProvider {
+    private val client = TrueTimeImpl()
+
+    init {
+        client.sync()
```
marandaneto, line 11: this cannot be `sync`, we cannot block the calling thread during the app start

posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidDateProvider.kt
```diff
@@ -0,0 +1,29 @@
...
+    override fun addSecondsToCurrentDate(seconds: Int): Date {
+        return Date(client.now().time + (seconds * 1000))
+    }
+
+    override fun currentTimeMillis(): Long {
+        return client.now().time
+    }
+
+    override fun nanoTime(): Long {
+        return System.nanoTime()
```
marandaneto, line 27: can we get the nano time from the `client`? otherwise nanoTime and timeMillis will be out of sync

posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidDateProvider.kt
beradeep: @marandaneto  `client.sync()` uses `Dispatchers.IO` internally

posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidDateProvider.kt
beradeep: shouldn't `nanoTime` only be used to measure to measure intervals? why do we need to be synced to wall-clock time?

buildSrc/src/main/java/PosthogBuildConfig.kt
beradeep: imo the only other good option left is `TrustedTime`, but i think it's dependent on play services, so impossible to unit test i guess?

marandaneto: what if we use https://developer.android.com/reference/android/os/SystemClock#currentNetworkTimeClock() only for API >= 33 and fallback to as is, so we know this is fixed going forward, but it's fine for older OS versions

marandaneto: TrustedTime also brings a dependency, so also not ideal if possible to avoid, and it'd not work in devices without play services (eg China)

beradeep: > what if we use https://developer.android.com/reference/android/os/SystemClock#currentNetworkTimeClock() only for API >= 33 and fallback to as is, so we know this is fixed going forward, but it's fine for older OS versions

yes i think this is a better approach from the dependency pov

beradeep: @marandaneto shall we go ahead with this, then?

marandaneto: Yes

posthog-android/build.gradle.kts
```diff
@@ -89,6 +89,7 @@ dependencies {
     implementation("androidx.lifecycle:lifecycle-common-java8:${PosthogBuildConfig.Dependencies.LIFECYCLE}")
     implementation("androidx.core:core:${PosthogBuildConfig.Dependencies.ANDROIDX_CORE}")
     implementation("com.squareup.curtains:curtains:${PosthogBuildConfig.Dependencies.CURTAINS}")
+    implementation("com.github.instacart:truetime-android:${PosthogBuildConfig.Dependencies.TRUE_TIME}")
```
marandaneto, line 92: we can remove this

buildSrc/src/main/java/PosthogBuildConfig.kt
marandaneto, line 59: we can remove this

posthog-android/src/main/java/com/posthog/android/PostHogAndroid.kt
```diff
@@ -79,6 +81,10 @@ public class PostHogAndroid private constructor() {
             config.replayStoragePrefix = config.replayStoragePrefix ?: replayPath.absolutePath
             val preferences = config.cachePreferences ?: PostHogSharedPreferences(context, config)
             config.cachePreferences = preferences
+            // Defaults to PostHogDeviceDateProvider when api < 33
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+                config.dateProvider = (config.dateProvider as? PostHogAndroidDateProvider) ?: PostHogAndroidDateProvider()
```
marandaneto, line 86:
```suggestion
                config.dateProvider = if (config.dateProvider is PostHogDeviceDateProvider) PostHogAndroidDateProvider() else config.dateProvider
```

I think like this makes it more readable?

settings.gradle.kts
```diff
@@ -10,6 +10,7 @@ dependencyResolutionManagement {
     repositories {
         google()
         mavenCentral()
+        maven { url = uri("https://jitpack.io") }
```
marandaneto, line 13: we can remove this

posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidDateProvider.kt
```diff
@@ -0,0 +1,38 @@
+package com.posthog.android.internal
+
+import android.os.Build
+import android.os.SystemClock
+import androidx.annotation.RequiresApi
+import com.posthog.internal.PostHogDateProvider
+import java.time.Instant
```
marandaneto, line 7: Instant requires desugaring https://developer.android.com/studio/write/java8-support-table (coreLibraryDesugaring)
so lets fallback everything to `java.util.Date` and not use `java.time.Instant` at all

posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidDateProvider.kt
```diff
@@ -0,0 +1,38 @@
...
+internal class PostHogAndroidDateProvider : PostHogDateProvider {
+    private val networkTimeClock =
+        runCatching {
+            SystemClock.currentNetworkTimeClock()
+        }
+
+    private val instant
+        get() =
+            networkTimeClock
+                .mapCatching { it.instant() }
```
marandaneto, line 20: Does this return the `now` instant or the timestamp when `SystemClock.currentNetworkTimeClock` was called?
those methods should always return the `now` time and not the cached timestamp, just to make sure I am reading this correctly

posthog-android/src/main/java/com/posthog/android/PostHogAndroid.kt
beradeep: the problem with this is that `PostDeviceDateProvider` is internal to the `posthog` module

posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidDateProvider.kt
beradeep: thanks for the info, i wasn't aware of this, was happy to learn about it 👍

posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidDateProvider.kt
beradeep: it returns the current `now` instant whenever called

beradeep:
@marandaneto sorry for delaying this, was busy with some work

posthog-android/src/main/java/com/posthog/android/PostHogAndroid.kt
marandaneto: we can make it public and mark it as internal, like a few other ones, use the `@PostHogInternal` annotation

posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidDateProvider.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    override fun addSecondsToCurrentDate(seconds: Int): Date {
+        return Date(currentTimeMillis() + seconds * 1000)
+    }
+
+    override fun currentTimeMillis(): Long {
+        val cal = Calendar.getInstance()
+        return networkTimeClock
+            .mapCatching { it.millis() }
+            .getOrDefault(cal.timeInMillis)
```
marandaneto, line 29:
```suggestion
        return networkTimeClock
            .mapCatching { it.millis() }
            .getOrDefault(System.currentTimeMillis())
```

marandaneto:
> @marandaneto sorry for delaying this, was busy with some work

no worries, thanks for the follow up, once you address the other comments, we can merge it!

beradeep:
> > @marandaneto sorry for delaying this, was busy with some work
> 
> no worries, thanks for the follow up, once you address the other comments, we can merge it!

hey i think i've addressed them. let me know if i missed any in case.

marandaneto:
> > > @marandaneto sorry for delaying this, was busy with some work
> > 
> > 
> > no worries, thanks for the follow up, once you address the other comments, we can merge it!
> 
> hey i think i've addressed them. let me know if i missed any in case.

there are a few left overs, eg https://github.com/PostHog/posthog-android/pull/235#discussion_r2020537899
just clean up work btw

marandaneto (approved):

PR #236, opened 2025-03-31
chore: surveys data classes

marandaneto:
data classes
_#skip-changelog_

## :pencil: Checklist


ioannisj (approved):
LG. Missing some @SerializedName I think but would probably come in next PRs

marandaneto:
> LG. Missing some @SerializedName I think but would probably come in next PRs

[`01584a4` (#236)](https://github.com/PostHog/posthog-android/pull/236/commits/01584a4907560016726344920a0a5991f01a950e)
its only needed when the property name does not match the json name, so I only added for the snake cases one which don't match the kotlin prop name

marandaneto:
> > > LG. Missing some @SerializedName I think but would probably come in next PRs
> > 
> > 
> > [`01584a4` (#236)](https://github.com/PostHog/posthog-android/pull/236/commits/01584a4907560016726344920a0a5991f01a950e) its only needed when the property name does not match the json name, so I only added for the snake cases one which don't match the kotlin prop name
> 
> So in case of `SurveyAppearance` this will be needed right?

no, https://github.com/PostHog/posthog/blob/eb8b837574a01ab8a126ca87d0f9d7cd55b7f8d6/posthog/api/survey.py#L100
the json prop name is called `appearance`, so it matches the kotlin prop name, `SurveyAppearance` is the class name, so this is not a problem.

PR #44, opened 2023-10-09
getFeatureFlagPayload returns non strigified JSON

marandaneto:
## :bulb: Motivation and Context
Relates to https://github.com/PostHog/posthog-ios/issues/49

## :green_heart: How did you test it?
Unit tests and creating a feature flag per type and testing them manually.

## :pencil: Checklist

- [X] I reviewed the submitted code.
- [X] I added tests to verify the changes.
- [ ] I updated the docs if needed.
- [ ] No breaking change or entry added to the changelog.

PR #241, opened 2025-04-10
chore(flags): Prepare tests for decide v4 work

haacked:
This is some pre-requisite work. It doesn't add support for v4.

Decide v4 changes the format for how flags are returned by the `/decide` API endpoint.

The existing tests and test files have been renamed to include v3 in the name because the library will need to be backwards compatible (I didn't rename the decide response files that have nothing to do with flags such as the ones dealing with session recordings because I won't need to implement a v4 version of those tests).

I also made the test responses more closely adhere to real world `/decide` responses. In particular:

1. In `featureFlagPayloads`, there were keys that were not in `featureFlags`. But in actuality, we only have payloads for active flags, so the key _must_ be in both places if there's a payload.
2. Feature flag payloads are always string encoded JSON or `null`.

I changed the files and tests to account for that.

Also removed an unused test file.

This will make it easier to review the upcoming changes.

Only unit tests are affected, no production code.

#skip-changelog

## :pencil: Checklist


posthog-android/lint-baseline.xml
```diff
@@ -4,8 +4,8 @@
     <issue
         id="GradleDependency"
         message="A newer version of androidx.lifecycle:lifecycle-process than 2.6.2 is available: 2.8.7"
-        errorLine1="    implementation(&quot;androidx.lifecycle:lifecycle-process:${PosthogBuildConfig.Dependencies.LIFECYCLE}&quot;)"
-        errorLine2="                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
```
haacked, line 8: These changes fixed `make compile` locally. I didn't want to change dependencies or anything as I don't have context. I don't think my changes introduced these build errors.

marandaneto: if you run `make generateLintBaseLine`, this will generate the new file with the latest versions, we don't want to depend on the latest version to avoid breaking changes.

posthog/src/test/java/com/posthog/PostHogTest.kt
```diff
@@ -33,7 +33,7 @@ internal class PostHogTest {
     private val serializer = PostHogSerializer(PostHogConfig(API_KEY))
     private lateinit var config: PostHogConfig
 
-    private val file = File("src/test/resources/json/basic-decide-no-errors.json")
+    private val file = File("src/test/resources/json/basic-decide-v3-no-errors.json")
```
marandaneto, line 36: maybe we just move these files into a v3 folder? its a bit more organized than a v3/v4 suffix in all files, not a blocker though, just an idea.

posthog/src/test/java/com/posthog/internal/PostHogRemoteConfigTest.kt
```diff
@@ -69,7 +69,7 @@ internal class PostHogRemoteConfigTest {
     }
 
     @Test
-    fun `load flags from decide api and call the onFeatureFlags callback`() {
+    fun `load flags from decide api and call the onFeatureFlags callback - v3`() {
```
marandaneto, line 72: Maybe we should move all 'flags' tests to a different file eg `PostHogFeatureFlagsV3Test.kt` and then v4? its more organized than a v3/v4 suffix in all tests.
This file would keep only the remote config tests which is a different API.

posthog-android/lint-baseline.xml
haacked: @marandaneto thanks for the tip!

posthog/src/test/java/com/posthog/PostHogTest.kt
haacked: Good call!

posthog/src/test/java/com/posthog/internal/PostHogRemoteConfigTest.kt
haacked: Wil do!

posthog/src/test/resources/json/basic-decide-recording-bool-linked-disabled.json
```diff
@@ -13,7 +13,7 @@
     "enable_collect_everything": true
   },
   "featureFlagPayloads": {
-    "thePayload": true
+    "4535-funnel-bar-viz": "true"
```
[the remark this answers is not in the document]
haacked: `true` is valid JSON, but it's not a valid payload. Decide payload's are always JSON encoded strings. The change to handle them was made in https://github.com/PostHog/posthog-android/pull/44.

posthog/src/test/java/com/posthog/internal/PostHogRemoteConfigTest.kt
haacked: I'm thinking the current version will just be named `PostHogFeatureFlagsTest.kt`. If we ever release a v5, then I'll rename the test to `PostHogFeatureFlagsV4Test.kt` and `PostHogFeatureFlagsTest.kt` will become v5. That way, we can see the diff of how the tests change from V4 to V5.

marandaneto (approved):

PR #243, opened 2025-04-15
feat:  Add more details such as version, id, and reason to $feature_flag_called events

haacked:

Add support for `/decide?v=4` (see https://github.com/PostHog/posthog/pull/29751). Similar to how it's done in https://github.com/PostHog/posthog-ios/pull/327.

The [`/decide` endpoint](https://posthog.com/docs/api/decide) version 4 changes how feature flags and payloads are delivered. Instead of two separate fields, they're combined into a new `flags` property that contains a flag object per flag key with details about the flag.

This PR updates the library to support this new format in order to pass more information to the `$feature_flag_called` event.

## Updates to the `$feature_flag_called` event

When capturing the `$feature_flag_called` event, additional information are now captured:

The end result is `$feature_flag_called` events will have additional properties:

| Property Name                |  Descirption
| ------------------------ | -------------
| `$feature_flag_version` | The version of the feature flag. This is visible in the diffs of the feature flag activity log.
| `$feature_flag_reason` | The reason the feature flag matched or didn't match.
| `$feature_flag_id`          | The database id of the feature flag (Just in case a flag was deleted and then recreated with the same key, we can determine the difference.

## Backwards compatibility:

The changes are all backwards compatible with `/decide?v=3`.

Uni tests. I don't know how to manually test it yet.

## :pencil: Checklist


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -757,11 +757,20 @@ public class PostHog private constructor(
         }
 
         if (config?.sendFeatureFlagEvent == true && shouldSendFeatureFlagEvent) {
+            var flagDetails = remoteConfig?.getFlagDetails(key)
+
+            val requestId = remoteConfig?.getRequestId()
```
marandaneto, line 762: just a nit but you could do a single check for `remoteConfig` or use `?.let`.
right now this will be 2 `if` conditions if `remoteConfig` is null or not

posthog/src/main/java/com/posthog/internal/PostHogDecideResponse.kt
```diff
@@ -15,5 +16,44 @@ internal data class PostHogDecideResponse(
...
+
+/**
+ * The feature flag data structure for calling the decide API v4. This is what each flag in "flags" will be.
+ * @property key the key of the feature flag
+ * @property enabled whether the feature flag is enabled
+ * @property variant the variant of the feature flag
+ * @property metadata the metadata of the feature flag
+ * @property reason the reason the feature flag was evaluated
+ */
+@IgnoreJRERequirement
```
marandaneto, line 32: we can stract those to new classes (files), its not common to have multiple classes in a single file on java/kotlin land

marandaneto: remember to add those classes that will be serialized/deserialized to the proguard file
https://github.com/PostHog/posthog-android/blob/d269ddb7a1e07590b00d68cd04b62d1b87716554/posthog-android/consumer-rules.pro#L13-L27
GSON uses reflection, and after minification, reflection breaks since classes are renamed, so we need to do this to keep it working when minification is enabled

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
```diff
@@ -331,6 +359,29 @@ internal class PostHogRemoteConfig(
         return parsedPayloads
     }
 
+    private fun normalizeDecideResponse(decideResponse: PostHogDecideResponse): PostHogDecideResponse {
+        val flags = decideResponse.flags
+        config.logger.log("Feature flags: $flags")
```
marandaneto, line 364: does that help? not sure if this stringifies correctly or just print out the class type

haacked: Ha nope! That was an attempt to add some debug output while looking into a bug. Not only did in not work, I forgot to clean it up. Nice catch!

haacked:
Thanks for all the feedback, @marandaneto. I haven't looked into actually running a manual test yet. Is it easier to get going than in ios or react native?

marandaneto:
> Thanks for all the feedback, @marandaneto. I haven't looked into actually running a manual test yet. Is it easier to get going than in ios or react native?

I'd say the same as ios, RN is def harder :D
as long as you have a proper set up installed, it should just work.
https://github.com/PostHog/posthog-android/blob/main/CONTRIBUTING.md
happy to fix it if something is off in the CONTRIBUTING file

bin/test
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+#/ Usage: bin/test
+#/ Description: Runs all the tests for this project
+source bin/helpers/_utils.sh
+set_source_and_root_dir
+
+./gradlew test
```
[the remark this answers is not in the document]
marandaneto: Just curious, why do you need those scripts? The make file is the very same thing so it feels just like duplication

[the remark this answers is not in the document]
haacked: I know it seems silly in this particular case, but the rationale for these scripts is in my RFC here: https://github.com/PostHog/meta/pull/309

I've been working on 11 repos in the past couple weeks and switching between them constantly. It's nice having a consistent approach.

haacked:
@marandaneto Appreciate all the help!

I followed the instructions and got this far:

Screenshot 2025-04-17 at 10 25 03 AM

Build output shows everything built correctly. However, I don't see the app in this screen. Any troubleshooting ideas? Does it work for you in this branch?

haacked:
Ok, the following commands worked:

```bash
adb kill-server
adb start-server
adb devices
```

haacked:
@marandaneto I updated the sample app to make a call to a feature flag and was able to confirm that my changes are working just fine! Thanks for your help again!

marandaneto:
> Ok, the following commands worked:
> 
> ```shell
> adb kill-server
> adb start-server
> adb devices
> ```

welcome to `adb` hell.
strange is that a fresh setup should just have worked, i usually have adb issues after changing environments.
this is not related to the project structure tho, more like android tooling flakiness really.

posthog-samples/posthog-android-sample/src/main/java/com/posthog/android/sample/NormalActivity.kt
```diff
@@ -33,13 +35,34 @@ class NormalActivity : ComponentActivity() {
...
 //            finish()
-            Thread {
-                client.newCall(
-                    okhttp3.Request.Builder()
-                        .url("https://google.com")
-                        .build(),
-                ).execute().closeQuietly()
-            }.start()
+            // Check if the "enable_network_request" feature flag is enabled
+            val isNetworkRequestEnabled = com.posthog.PostHog.isFeatureEnabled("enable_network_request", false)
```
marandaneto, line 39:
```suggestion
            val isNetworkRequestEnabled = PostHog.isFeatureEnabled("enable_network_request", false)
```

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -757,12 +757,22 @@ public class PostHog private constructor(
         }
 
         if (config?.sendFeatureFlagEvent == true && shouldSendFeatureFlagEvent) {
-            val props = mutableMapOf<String, Any>()
-            props["\$feature_flag"] = key
-            // value should never be nullabe anyway
-            props["\$feature_flag_response"] = value ?: ""
-
-            capture("\$feature_flag_called", properties = props)
+            remoteConfig?.run {
```
marandaneto, line 760: I'd use `let` instead
https://kotlinlang.org/docs/scope-functions.html#let
its the right scope for checking nullability

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -757,12 +757,22 @@ public class PostHog private constructor(
...
+                val flagDetails = getFlagDetails(key)
+                val requestId = getRequestId()
+
+                val props = mutableMapOf<String, Any>()
+                props["\$feature_flag"] = key
+                // value should never be nullabe anyway
+                props["\$feature_flag_response"] = value ?: ""
+                props["\$feature_flag_request_id"] = requestId ?: ""
+                if (flagDetails != null) {
```
marandaneto, line 769: `let` is a bit shorter as well

marandaneto (approved):
left a few other small comments otherwise LGTM

PR #246, opened 2025-04-29
fix: clear cache if there are no active flags server side

marandaneto:
fix for https://github.com/PostHog/posthog-ios/pull/346#discussion_r2066856115

## :pencil: Checklist

### DIFF
--- a/posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
@@ -137,6 +137,14 @@ internal class PostHogRemoteConfig(
                                 isFeatureFlagsLoaded = true
                                 clearFlags()
                             }
+
+                            // if we don't load the feature flags (because there are none), we need to call the callback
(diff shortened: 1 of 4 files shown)


posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
```diff
@@ -207,14 +217,7 @@ internal class PostHogRemoteConfig(
                             """Feature flags are quota limited, clearing existing flags.
                                     Learn more about billing limits at https://posthog.com/docs/billing/limits-alerts""",
                         )
-                        this.featureFlags = null
```
marandaneto, line 210: was it intentional that we didnt clear `requestId` here? if so, I should conditionally clear `requestId` or not when calling `clearFlags`

marandaneto: @dmarticus question for you since you added this part

marandaneto:
@ioannisj @PostHog/team-feature-flags ready for review

PR #247, opened 2025-05-06
feat: auto-rotate session after 24 hours

pauldambra:
see https://github.com/PostHog/posthog/issues/30889

android doesn't rotate session ID after 24 hours

i think this PR adds that 🙈


posthog/src/test/java/com/posthog/internal/FakePostHogDateProvider.kt
```diff
@@ -29,10 +29,10 @@ internal class FakePostHogDateProvider : PostHogDateProvider {
     }
 
     override fun currentTimeMillis(): Long {
-        return System.currentTimeMillis()
+        return currentDate?.time ?: System.currentTimeMillis()
```
pauldambra, line 32: this threw me 🤣

posthog/src/test/java/com/posthog/internal/FakePostHogDateProvider.kt
```diff
@@ -29,10 +29,10 @@ internal class FakePostHogDateProvider : PostHogDateProvider {
...
     }
 
     override fun nanoTime(): Long {
-        return System.nanoTime()
+        return currentDate?.time?.times(1000000) ?: System.nanoTime()
```
pauldambra, line 36: so I updated here too

posthog/src/main/java/com/posthog/internal/PostHogSessionManager.kt
```diff
@@ -4,6 +4,8 @@ import com.posthog.PostHogInternal
 import com.posthog.vendor.uuid.TimeBasedEpochGenerator
 import java.util.UUID
 
+public const val MAX_SESSION_AGE_MILLIS: Int = 24 * 60 * 60 * 60
```
pauldambra, line 7: I guess maybe this should be in config 🤷

pauldambra: ah, no, this is hard coded in posthog-js too

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -466,6 +469,35 @@ public class PostHog private constructor(
         }
     }
 
+    /**
+     * Because we use UUIDv7
+     * we can convert a session id into the session start timestamp
+     * and then use that to enforce the maximum session length
+     */
+    private fun enforceMaxSessionLength(buildProperties: Map<String, Any>): Map<String, Any> {
```
marandaneto, line 477: react native has its own session id rotation so we should not do this for RN I think
https://github.com/PostHog/posthog-js-lite/blob/b8372284df5489457d95b16b0e4b12a91dd66c26/posthog-core/src/index.ts#L1309-L1315
otherwise they will be out of sync (RN uses a different session id than Android)

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -466,6 +469,35 @@ public class PostHog private constructor(
...
+                ?: return buildProperties
+
+        val currentTimeMillis =
+            config?.dateProvider?.currentTimeMillis() ?: System.currentTimeMillis()
+        val sessionAge = currentTimeMillis - sessionStartTimestamp
+        if (sessionAge > MAX_SESSION_AGE_MILLIS) {
+            // rotate session
+            // update properties
+            PostHogSessionManager.endSession()
+            PostHogSessionManager.startSession(currentTimeMillis)
```
marandaneto, line 491: this changes the session id but the session replay would need to be notified to eg send a new full snapshot with metadata etc otherwise it'll assume its already sent and the replay will be broken (missing meta event and full snapshot for the new session id)

posthog/src/main/java/com/posthog/internal/PostHogSessionManager.kt
marandaneto, line 7: can it be private instead?
if needed you can put it inside a companion object

posthog/src/main/java/com/posthog/PostHog.kt
pauldambra: 🤔 in the JS SDK we have an event emitter that we pass around that can be used for this kind of "onBlah" behavior

I guess the sessionManager can have a `onSessionStarted` that accepts a callback?

Is there a more android-y way of doing it?

posthog/src/main/java/com/posthog/PostHog.kt
pauldambra: so, needs to be configurable? and then RN would turn off the Android one?
or sync up a release and RN stops doing its own?

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -432,16 +433,18 @@ public class PostHog private constructor(
...
-                    newDistinctId,
-                    properties = properties,
-                    userProperties = userProperties,
-                    userPropertiesSetOnce = userPropertiesSetOnce,
-                    groups = groups,
-                    // only append shared props if not a snapshot event
-                    appendSharedProps = !snapshotEvent,
-                    // only append groups if not a group identify event and not a snapshot
-                    appendGroups = !groupIdentify,
+                enforceMaxSessionLength(
```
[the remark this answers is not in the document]
pauldambra: i guess the foreground change is righ there

was assuming that we didn't want to run a timer just for this - although ultimately that's where JS SDK ended up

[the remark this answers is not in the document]
marandaneto: we already have a timer when the app is put in the background, see [here](https://github.com/PostHog/posthog-android/blob/cfab460b6dfbe6a25487f9499339ee7263076e73/posthog-android/src/main/java/com/posthog/android/internal/PostHogLifecycleObserverIntegration.kt#L47-L107C9)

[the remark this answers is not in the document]
marandaneto: maybe that timer should also call `enforceMaxSessionLength` when its back in the foreground so we don't rely on the next event but rather if the app is back visible to the user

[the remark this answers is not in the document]
pauldambra: me: "let's open a PR and see how much I don't know"
me: *regrets life choices

🤣

posthog/src/main/java/com/posthog/PostHog.kt
marandaneto: yeah a callback would be one way, since `PostHogSessionManager` lives in the java-only lib and most of the replay code is Android-dependent, we will need to pass callbacks around, or implement a pub/sub approach for "app is in background, app is in foreground, etc" so the java bits only can do things

posthog/src/main/java/com/posthog/internal/PostHogSessionManager.kt
```diff
@@ -16,10 +18,10 @@ public object PostHogSessionManager {
 
     private var sessionId = sessionIdNone
 
-    public fun startSession() {
+    public fun startSession(sessionStartTime: Long? = null) {
         synchronized(sessionLock) {
             if (sessionId == sessionIdNone) {
-                sessionId = TimeBasedEpochGenerator.generate()
+                sessionId = TimeBasedEpochGenerator.generate(sessionStartTime ?: System.currentTimeMillis())
```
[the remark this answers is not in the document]
marandaneto: thats fine, another approach would be to make a TimeBasedEpochGenerator an interface, and the current TimeBasedEpochGenerator is the production implementation and we could have a test implementation and swap during SDK initialization just for testing, the prod code would be cleaner but we would pay the price of an extra interface

PR #261, opened 2025-06-12
fix: start session recording after remote config and flags are evaluated

marandaneto:
Should fix https://github.com/PostHog/posthog-flutter/issues/184
the problem here is: we start session recording once the app starts (and sdk) as early as possible.
we rely on remote config and feature flags if needed, by this moment, maybe remote config and feature flags aren't loaded yet.
we use the cache if none but the cache might be empty when its a fresh app start (usually after app install).
we don't start recording until the app is restarted losing the recording of the very first app start, this is no good.
now we do, we run a callback once remote config and flags are loaded and try to start session recording again if not yet.

running the app on a fresh app start

## :pencil: Checklist


marandaneto:
@PostHog/team-replay @ioannisj review appreciated so i can ship this fix asap

pauldambra (approved):
quick scan, lgtm

marandaneto:
> LG - I wonder if we should always react to flag changes for session replay though, even when people manually call reloadFeatureFlags?

If say we should start session recording if it was disabled and now it is disabled after reloaded flags.
I'd say we should not stop session recording if already running if now it should be disabled, just because they already paid for the recording anyway, and it would be an even worse experience having a recording with a hard stop or even just eg lets say 1s after app start

marandaneto:
> LG - I wonder if we should always react to flag changes for session replay though, even when people manually call reloadFeatureFlags?

new commit addresses this use case, see [this comment as well](https://github.com/PostHog/posthog-android/pull/261#issuecomment-2969327185)

PR #263, opened 2025-06-13
fix callback onFeatureFlags

KopeikinaDarya:
added fix call onFeatureFlags when reloadFeatureFlags()  -> config?.onFeatureFlags 

public override fun reloadFeatureFlags(onFeatureFlags: PostHogOnFeatureFlags?) {
        if (!isEnabled()) {
            return
        }
        loadFeatureFlagsRequest(onFeatureFlags)
    }


marandaneto:
@KopeikinaDarya makes sense, thanks

marandaneto:
@KopeikinaDarya closing because of https://github.com/PostHog/posthog-android/pull/261
that PR addresses similar use case, thanks anyway

PR #273, opened 2025-08-07
fix: missing session_id property for snapshot events

marandaneto:
Bug introduced by https://github.com/PostHog/posthog-android/pull/266
> Logs -> Event dropped, because snapshot and session_id are missing

Testing replay

## :pencil: Checklist


posthog/src/test/java/com/posthog/PostHogTest.kt
```diff
@@ -1426,6 +1427,40 @@ internal class PostHogTest {
         sut.close()
     }
 
+    @Test
```
marandaneto, line 1430: we had tests for $session_id, but not for snapshot events specifically which was a special case

PR #287, opened 2025-09-27
feat: Core configuration factories for queue and feature flags

dustinbyrne:

This change allows core configuration to be extended via composition, hiding configuration that is not needed in a specific environment (e.g., server-side applications do not need survey configuration).

`PostHogConfig` now has two factory methods from which it will construct the event queue and remote config. By default, `PostHogQueue` and `PostHogRemoteConfig` will be used for backwards compatibility. This change will not affect any existing behavior.

As a result of this change, a whole bunch of things that were marked `internal` are now public (but typically annotated `@PostHogInternal`). This allows the server-side module to define it's own feature flags logic without adding an alternate implementation to the core module.

Existing tests are passing. This change does not affect behavior. I haven't yet compiled against Flutter or React Native, but I want to get this out first.

## :pencil: Checklist

### DIFF
--- a/posthog/src/main/java/com/posthog/internal/PostHogFlagsRequest.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogFlagsRequest.kt
@@ -6,9 +6,10 @@ package com.posthog.internal
 internal class PostHogFlagsRequest(
     apiKey: String,
     distinctId: String,
-    anonymousId: String?,
-    groups: Map<String, String>?,
-    // add person_properties, group_properties
+    anonymousId: String? = null,
+    groups: Map<String, String>? = null,
+    personProperties: Map<String, String>? = null,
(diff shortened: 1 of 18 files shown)


posthog/src/main/java/com/posthog/internal/PostHogFlagsRequest.kt
```diff
@@ -19,5 +20,11 @@ internal class PostHogFlagsRequest(
         if (groups?.isNotEmpty() == true) {
             this["\$groups"] = groups
         }
+        if (personProperties?.isNotEmpty() == true) {
+            this["\$properties"] = personProperties
+        }
+        if (groupProperties?.isNotEmpty() == true) {
+            this["\$group_properties"] = groupProperties
+        }
```
marandaneto, line 28: nice addition

marandaneto (approved):

Issue #292, opened 2025-09-30
Network failure in Android `testReleaseUnitTest` in CI


dustinbyrne:
### Description

E.g., https://github.com/PostHog/posthog-android/actions/runs/18135099382/job/51613467083

Unsure atm if this is intermittent or persistent. Looks to have began as of https://github.com/PostHog/posthog-android/commit/9776e3ab4d0feb677dd0ff56d2dbbfd386d1a3ce.

https://github.com/PostHog/posthog-android/commit/9ff32f55c644aef436daf935a17c9b35c5472466 would have been a likely culprit to me, but it actually passed here.

PR #293, opened 2025-09-30
fix(core): Deduplicate `$feature_flag_called` events

dustinbyrne:

From documentation:

> *Note:* Tracking whether it's a new user or if a flag value has changed happens in a local cache. This means that if you reinitialize the PostHog client, the cache resets as well – causing `$feature_flag_called` events to be sent again when calling `getFeatureFlag` or `isFeatureEnabled`. PostHog is built to handle this, and so duplicate `$feature_flag_called` events won't affect your analytics.

This appears not to have been true of the `PostHogStateless` client. This PR fixes that.

Additional unit tests. Some unrelated [Android tests are failing](https://github.com/PostHog/posthog-android/issues/292) and I'll be looking into that shortly in another PR.

## :pencil: Checklist

### DIFF
--- a/posthog/src/test/java/com/posthog/PostHogStatelessTest.kt
+++ b/posthog/src/test/java/com/posthog/PostHogStatelessTest.kt
@@ -797,6 +797,61 @@ internal class PostHogStatelessTest {
     }
 
     @Test
+    fun `feature flag called events are deduplicated per distinct ID`() {
+        val mockQueue = MockQueue()
+        val mockFeatureFlags = MockFeatureFlags()
+        mockFeatureFlags.setFlag("test_flag", "variant_a")
+
+        sut = createStatelessInstance()
+        config = createConfig(sendFeatureFlagEvent = true)
+
+        sut.setup(config)
+        sut.setMockQueue(mockQueue)
+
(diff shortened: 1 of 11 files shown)


posthog/src/test/java/com/posthog/PostHogStatelessTest.kt
```diff
@@ -796,6 +796,61 @@ internal class PostHogStatelessTest {
...
+        // Should generate two $feature_flag_called events (one per distinct ID)
+        assertEquals(2, mockQueue.events.size)
+        assertEquals("user123", mockQueue.events[0].distinctId)
+        assertEquals("variant_a", mockQueue.events[0].properties!!["${'$'}feature_flag_response"])
+        assertEquals("user456", mockQueue.events[1].distinctId)
+        assertEquals("variant_a", mockQueue.events[1].properties!!["${'$'}feature_flag_response"])
+    }
+
+    @Test
+    fun `feature flag called events are sent for different values of same flag`() {
```
haacked, line 829: Good test!

This covers when variants are different, but for completeness, I'd also make sure that two events are sent for same `distinct_id` and same flag when the evaluated value changes (aka `true`, `false`, and `null`).

PR #297, opened 2025-10-02
feat: Allow stateless override of event timestamp

dustinbyrne:

Some users wish to use their own timestamps when capturing an event to ensure order of events. This change allows users to set their own timestamps when capturing an event.

Unit tests and local testing via the sample Java project

## :pencil: Checklist

### DIFF
--- a/posthog-server/src/main/java/com/posthog/server/PostHog.kt
+++ b/posthog-server/src/main/java/com/posthog/server/PostHog.kt
@@ -41,6 +41,7 @@ public class PostHog : PostHogInterface {
         userProperties: Map<String, Any>?,
         userPropertiesSetOnce: Map<String, Any>?,
         groups: Map<String, String>?,
+        timestamp: java.util.Date?,
     ) {
         instance?.captureStateless(
             event,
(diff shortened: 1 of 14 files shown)


posthog-server/src/main/java/com/posthog/server/PostHog.kt
```diff
@@ -41,6 +41,7 @@ public class PostHog : PostHogInterface {
         userProperties: Map<String, Any>?,
         userPropertiesSetOnce: Map<String, Any>?,
         groups: Map<String, String>?,
+        timestamp: java.util.Date?,
```
marandaneto, line 44:
```suggestion
        timestamp: Date?,
```

posthog/src/main/java/com/posthog/PostHogStateless.kt
```diff
@@ -263,10 +271,17 @@ public open class PostHogStateless protected constructor(
         event: String,
         distinctId: String,
         properties: MutableMap<String, Any>,
+        timestamp: Date? = null,
```
marandaneto, line 274:
```suggestion
        timestamp: Date?,
```

marandaneto (approved):

PR #298, opened 2025-10-02
feat: Hook up stateless feature flag property overrides

dustinbyrne:

In anticipation of local evaluation, we'll need the ability to provide `group`/`personProperties`/`groupProperties` to our server-side PostHog client when requesting feature flags. These properties were partially included during the stateless refactor.

Technically, this is a 🚨  **breaking change** 🚨  to `PostHogStatelessInterface` as it changes the shape of the interface. This will be a major version bump to the core library.

For the server library, this is only a minor version bump.

I've added a number of unit tests.

## :pencil: Checklist


posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagOptions.kt
```diff
@@ -0,0 +1,119 @@
...
+         * Add a single custom property to the capture options
+         */
+        public fun group(
+            key: String,
+            value: String,
+        ): Builder {
+            if (groups == null) {
+                groups = mutableMapOf()
+            }
+            groups!![key] = value
```
marandaneto, line 37: not a big fan of `!!` but since its a builder, i dont see this being a problem (like concurrency issues)
maybe consider declaring the maps already created and later check if they are not empty instead?

marandaneto (approved):

PR #299, opened 2025-10-03
feat(server): Add local evaluation

### LINKED ISSUE #307: Add local evaluation
### Description

`posthog-server` should have the ability to poll for flag definitions and evaluate conditions locally.

dustinbyrne:

This PR adds local evaluation capabilities to feature flags. See the updated [USAGE.md](https://github.com/PostHog/posthog-android/blob/f7f9a26b98ca46999efee28a7e68cc07d665686a/posthog-server/USAGE.md) for details on how to use it. 

If you're reviewing this PR, skimming [`FlagEvaluator.kt`](https://github.com/PostHog/posthog-android/blob/320f16a864b3035f42fe80a6449b767af4ee51a7/posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt), [`LocalEvaluationPoller.kt`](https://github.com/PostHog/posthog-android/blob/320f16a864b3035f42fe80a6449b767af4ee51a7/posthog-server/src/main/java/com/posthog/server/internal/LocalEvaluationPoller.kt), and changes to [`PostHogFeatureFlags.kt`](https://github.com/PostHog/posthog-android/pull/299/files#diff-85378bde6f6b29576b0ac64b276dac3dabe37425aa5872e8a8ccb860d3b1c529) would be the best places to start.

This change is pretty massive. I'm sure there's still a couple of bugs lurking, but test coverage is pretty good. This PR doesn't really change the public API, so we can always follow up with smaller bug fixes. 

Resolves #307 

_Lots_ of unit testing and updating the Java sample app to use local eval

## :pencil: Checklist

### DIFF
--- a/posthog-server/src/test/java/com/posthog/server/internal/PostHogFeatureFlagsTest.kt
+++ b/posthog-server/src/test/java/com/posthog/server/internal/PostHogFeatureFlagsTest.kt
@@ -13,6 +13,7 @@ import okhttp3.mockwebserver.Dispatcher
 import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
 import okhttp3.mockwebserver.RecordedRequest
+import java.util.concurrent.CountDownLatch
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertFalse
@@ -835,20 +836,24 @@ internal class PostHogFeatureFlagsTest {
         // Shut down poller to control loading manually
         featureFlags.shutDown()
 
-        // Launch 5 concurrent threads all calling loadFeatureFlagDefinitions
         val threadCount = 5
+        val startLatch = CountDownLatch(threadCount)
         val threads =
             List(threadCount) {
                 Thread {
+                    // Wait for all threads to be ready before proceeding. This should reduce
+                    // any timing issues where one thread completes before others start - particularly
+                    // in CI.
+                    startLatch.countDown()
+                    startLatch.await()
                     featureFlags.loadFeatureFlagDefinitions()
                 }
             }
 
-        // Start all threads simultaneously
         threads.forEach { it.start() }
 
         // Wait for all to complete
-        threads.forEach { it.join(5000) } // 5 sec timeout
+        threads.forEach { it.join(5000) }
 
         // All threads should have completed successfully
         threads.forEach { thread ->


posthog-server/src/main/java/com/posthog/server/internal/FlagDefinitionModels.kt
```diff
@@ -0,0 +1,86 @@
...
+    val variant: String?,
+)
+
+/**
+ * A property condition for flag evaluation
+ */
+internal data class FlagProperty(
+    val key: String,
+    val value: Any?,
+    val operator: String?,
```
marandaneto, line 52: is `operator` a reserved name in kotlin or just gh diff coloring?
if its reserved, you can use backticks to escape the reserved name or rename and use `SerializedName`

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -0,0 +1,803 @@
...
+     * Given the same distinct_id and key, it'll always return the same float.
+     * These floats are uniformly distributed between 0 and 1.
+     */
+    private fun hash(
+        key: String,
+        distinctId: String,
+        salt: String = "",
+    ): Double {
+        val hashKey = "$key.$distinctId$salt"
+        val digest = MessageDigest.getInstance("SHA-1")
```
marandaneto, line 54: should this be a global var or do we need a fresh instance every time?

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -0,0 +1,803 @@
...
+
+    /**
+     * Build variant lookup table for efficient variant selection. Order of the variants matters,
+     * and this implementation mirrors the ordering provided by the local evaluation API.
+     */
+    private fun variantLookupTable(flag: FlagDefinition): List<VariantLookupEntry> {
+        val lookupTable = mutableListOf<VariantLookupEntry>()
+        var valueMin = 0.0
+
+        val variants = flag.filters.multivariate?.variants ?: emptyList()
```
marandaneto, line 90: yo could use `?.let` here so you dont need to create a new instance of `emptyList()` if its null

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -0,0 +1,803 @@
...
+    /**
+     * Match a property condition against property values
+     * Only looks for matches where key exists in propertyValues
+     */
+    fun matchProperty(
+        property: FlagProperty,
+        propertyValues: Map<String, Any?>,
+    ): Boolean {
+        val key = property.key
+        val operator = property.operator ?: "exact"
```
marandaneto, line 114: i'd make the hardcoded oprators const or an enu, so we dont DRY also defensive programming

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -0,0 +1,803 @@
...
+
+    private fun compareDates(
+        overrideValue: Any?,
+        value: Any?,
+        operator: String,
+    ): Boolean {
+        val parsedDate =
+            try {
+                parseDateValue(value.toString())
+            } catch (e: Exception) {
```
marandaneto, line 259: I'd always use `Throwable` since it catches more than just `Exception`

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -0,0 +1,803 @@
...
+
+        // Fall back to absolute date parsing
+        return parseOverrideDate(value)
+    }
+
+    /**
+     * Parse relative date format (e.g., "-1d" or "1d" for 1 day ago). Always produces a date in the past.
+     */
+    private fun parseRelativeDate(value: String): ZonedDateTime? {
+        val regex = Regex("^-?([0-9]+)([hdwmy])$")
```
marandaneto, line 303: can be a global var so we dont have to compure the regex for every `parseRelativeDate` call

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
dustinbyrne: `getInstance` is a bit of a misnomer here - it's actually creating a new, stateful digest instance. It's possible we could use a single instance, but then would need to be synchronized. In this case, I don't think it's necessarily warranted.

posthog-server/src/main/java/com/posthog/server/internal/FlagDefinitionModels.kt
dustinbyrne: It is a Kotlin keyword, but in this case it'd be shadowed. I went back and forth on this a bit. Since you've called it out, I've changed these names.

posthog/src/main/java/com/posthog/internal/PostHogLocalEvaluationModels.kt
```diff
@@ -0,0 +1,191 @@
...
+@IgnoreJRERequirement
+@PostHogInternal
+public data class FlagDefinition(
+    val id: Int,
+    val name: String,
+    val key: String,
+    val active: Boolean,
+    val filters: FlagFilters,
+    val version: Int,
+)
```
dustinbyrne: @marandaneto Is this okay to annotate with `@IgnoreJRERequirement` to suppress these Animal Sniffer errors?
```
[Undefined reference | android-api-level-21-5.0.1_r2] com.posthog.internal.(PostHogLocalEvaluationModels.kt:1)
  >> int Integer.hashCode(int)

[Undefined reference | android-api-level-21-5.0.1_r2] com.posthog.internal.(PostHogLocalEvaluationModels.kt:1)
  >> int Double.hashCode(double)
```

According to [gummy-bears docs](https://github.com/open-toast/gummy-bears#android), this is incorrectly flagged as being added in API 24 despite working in API 19

dustinbyrne: Thanks! I switched to regular classes to avoid generating those methods and dropped the `@IgnoreJRERequirement` annotations.

posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
```diff
@@ -72,28 +241,249 @@ internal class PostHogFeatureFlags(
...
+
+        synchronized(loadLock) {
+            while (isLoading) {
+                wasWaitingForLoad = true
+                try {
+                    loadLock.wait()
+                } catch (e: InterruptedException) {
+                    Thread.currentThread().interrupt()
+                    config.logger.log("Interrupted while waiting for flag definitions to load")
+                    return
```
dmarticus: This returns without notifying waiters afaict, which means that if it's interrupted, it'll return without setting `isLoading = false`, so other threads might wait.  

Maybe do something like this?

```
synchronized(loadLock) {
    while (isLoading) {
        try {
            loadLock.wait()
        } catch (e: InterruptedException) {
            Thread.currentThread().interrupt()
            isLoading = false
            loadLock.notifyAll()
            throw InterruptedException("Interrupted while waiting")
        }
    }
}
```

posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
```diff
@@ -1,22 +1,59 @@
...
+    private var isLoading = false
+
+    private val loadLock = Object()
+
+    init {
+        startPoller()
+        if (!localEvaluation) {
+            onFeatureFlags?.loaded()
+        }
+    }
```
dmarticus: you should exception handle here, too, e.g.
```
init {
    try {
        startPoller()
        if (!localEvaluation) {
            onFeatureFlags?.loaded()
        }
    } catch (e: Throwable) {
        stopPoller()  // Cleanup on failure
        throw e
    }
}
```

posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
```diff
@@ -43,27 +84,155 @@ internal class PostHogFeatureFlags(
...
             personProperties,
             groupProperties,
-        )?.get(key)?.metadata?.payload
+        )?.metadata?.payload
             ?: defaultValue
     }
 
-    override fun getFeatureFlags(
-        distinctId: String?,
+    private fun resolveFeatureFlag(
```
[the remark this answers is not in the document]
dmarticus: sweet!

PR #305, opened 2025-10-13
chore: exception autocapture

marandaneto:
part of https://github.com/PostHog/posthog/issues/38845

## :pencil: Checklist

### DIFF
--- a/posthog-samples/posthog-android-sample/src/main/java/com/posthog/android/sample/DoSomething.kt
+++ b/posthog-samples/posthog-android-sample/src/main/java/com/posthog/android/sample/DoSomething.kt
@@ -5,18 +5,9 @@ import com.posthog.PostHog
 class DoSomething {
     fun doSomethingNow() {
         try {
-//            val suppresed = Exception("suppressed")
-//
-//            val cause = Exception("case")
-//            cause.addSuppressed(suppresed)
-//
-//            throw Exception("Something went wrong", cause)
-            val cause1 = Exception("Exception cause 1")
-            val cause2 = Exception("Exception cause 2,", cause1)
-            throw Exception("Something went wrong", cause2)
+            throw MyCustomException("Something went wrong")
         } catch (e: Throwable) {
-            e.printStackTrace()
-            PostHog.captureException(e, mapOf("am-i-stupid" to true))
+            PostHog.captureException(e, mapOf("my-custom-error" to true))
         }
     }
 }
--- a/posthog-samples/posthog-android-sample/src/main/java/com/posthog/android/sample/MyApp.kt
+++ b/posthog-samples/posthog-android-sample/src/main/java/com/posthog/android/sample/MyApp.kt
@@ -26,12 +26,12 @@ class MyApp : Application() {
                 captureDeepLinks = false
                 captureApplicationLifecycleEvents = false
                 captureScreenViews = false
-                sessionReplay = false
-                preloadFeatureFlags = false
-                sendFeatureFlagEvent = false
+                sessionReplay = true
+                preloadFeatureFlags = true
+                sendFeatureFlagEvent = true
                 onFeatureFlags = PostHogOnFeatureFlags { print("feature flags loaded") }
                 addBeforeSend { event ->
-                    if (event.event == "test_name") {
+                    if (event.event == "test_event") {
                         null
                     } else {
                         event
--- a/posthog-samples/posthog-android-sample/src/main/java/com/posthog/android/sample/NormalActivity.kt
+++ b/posthog-samples/posthog-android-sample/src/main/java/com/posthog/android/sample/NormalActivity.kt
@@ -38,13 +38,13 @@ class NormalActivity : ComponentActivity() {
 //            finish()
             // Check if the "enable_network_request" feature flag is enabled
 
-            var str: String? = null
-            if (str!!.startsWith("")) {
-                str = "123"
-                Toast.makeText(this, str, Toast.LENGTH_SHORT).show()
-            }
-//            val domsehting = DoSomething()
-//            domsehting.doSomethingNow()
+//            var str: String? = null
+//            if (str!!.startsWith("")) {
+//                str = "123"
+//                Toast.makeText(this, str, Toast.LENGTH_SHORT).show()
+//            }
+            val doSomething = DoSomething()
+            doSomething.doSomethingNow()
 
             val isNetworkRequestEnabled = PostHog.isFeatureEnabled("enable_network_request", false)
 


posthog/src/main/java/com/posthog/internal/exceptions/ThrowableCoercer.kt
```diff
@@ -84,13 +96,13 @@ internal class ThrowableCoercer {
...
+                            "type" to mechanismType,
                         ),
                     "thread_id" to threadId,
                 )
-            if (theThrowable.message?.isNotEmpty() == true) {
-                exception["value"] = theThrowable.message
-            }
+
+            // value is required, using an empty placeholder
+            exception["value"] = theThrowable.message ?: ""
```
marandaneto: will undo once https://github.com/PostHog/posthog/pull/39571/ lands

posthog/src/main/java/com/posthog/internal/exceptions/UncaughtExceptionHandlerAdapter.kt
```diff
@@ -0,0 +1,23 @@
+package com.posthog.internal.exceptions
+
+internal interface UncaughtExceptionHandlerAdapter {
```
ioannisj: This is for testing aid right?

posthog/src/main/java/com/posthog/exceptions/PostHogExceptionAutoCaptureIntegration.kt
```diff
@@ -0,0 +1,75 @@
...
+
+        integrationInstalled = true
+        config.logger.log("Exception autocapture is enabled.")
+    }
+
+    override fun uninstall() {
+        if (!integrationInstalled) {
+            return
+        }
+        adapterExceptionHandler.setDefaultUncaughtExceptionHandler(defaultExceptionHandler)
```
ioannisj: Following the comment above, the order of `Thread.setDefaultUncaughtExceptionHandler` and `setup()` calls matters here. If they call setup() before adding their custom handler, we'll get a null value for defaultExceptionHandler, which means on reset/uninstall we'll clear their custom handler as well

posthog/src/main/java/com/posthog/internal/exceptions/UncaughtExceptionHandlerAdapter.kt
marandaneto: yes

posthog/src/main/java/com/posthog/exceptions/PostHogExceptionAutoCaptureIntegration.kt
```diff
@@ -0,0 +1,75 @@
...
+        val currentExceptionHandler = adapterExceptionHandler.getDefaultUncaughtExceptionHandler()
+
+        if (currentExceptionHandler != null) {
+            if (currentExceptionHandler !is PostHogExceptionAutoCaptureIntegration) {
+                defaultExceptionHandler = currentExceptionHandler
+            }
+        } else {
+            defaultExceptionHandler = null
+        }
+        adapterExceptionHandler.setDefaultUncaughtExceptionHandler(this)
```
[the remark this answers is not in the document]
marandaneto: as long as they call the previous handler, its always ok
its not common that you have multiple handlers, but it can happen
i dont think it has to be documented because all error tracking tools call the previous handler so its always chained

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
```diff
@@ -48,60 +48,89 @@ internal class PostHogQueue(
...
+
+    private fun removeEventSync() {
+        var removeFirst = false
+        if (deque.size >= config.maxQueueSize) {
+            removeFirst = true
+        }
 
-                val uuid = event.uuid ?: TimeBasedEpochGenerator.generate()
-                val file = File(dir, "$uuid.event")
+        if (removeFirst) {
```
ablaszkiewicz: Why not just this?

```
if (deque.size >= config.maxQueueSize)
```

posthog/src/main/java/com/posthog/internal/errortracking/ThrowableCoercer.kt
```diff
@@ -22,21 +22,33 @@ internal class ThrowableCoercer {
...
         val circularDetector = hashSetOf<Throwable>()
 
+        var handled = true
+        var isFatal = false
+        var mechanismType = "generic"
+
         var currentThrowable: Throwable? = throwable
+        val threadId: Long
+
+        if (throwable is PostHogThrowable) {
```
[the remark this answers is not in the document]
marandaneto: L41

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
```diff
@@ -48,60 +48,89 @@ internal class PostHogQueue(
...
+            try {
+                val first: File
                 synchronized(dequeLock) {
-                    deque.add(file)
+                    first = deque.removeFirst()
                 }
+                first.deleteSafely(config)
+                config.logger.log("Queue is full, the oldest event ${first.name} is dropped.")
```
oliverb123: Unsure how log levels work here - can we log this as a warning or error?

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
marandaneto: i was reusing removeFirst before but not anymore, good catch, will remove

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
```diff
@@ -48,60 +48,89 @@ internal class PostHogQueue(
 
     private val delay: Long get() = (config.flushIntervalSeconds * 1000).toLong()
 
-    override fun add(event: PostHogEvent) {
-        executor.executeSafely {
-            var removeFirst = false
-            if (deque.size >= config.maxQueueSize) {
-                removeFirst = true
+    private fun addEventSync(event: PostHogEvent): Boolean {
```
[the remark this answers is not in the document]
marandaneto: yes, thats correct.
its actually the oposite, its easier for the OS to write multiple files than read from a single file with a long queue, you have to buffer more and keep the file open in memory (more memory comsumption which is limited in mobile).
also you'd need to sync all operations to be thread safe otherwise you might corrupt the file.
this is a fire and forget approach, so super easy to write/delete.

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
marandaneto: right now its just a console log, no log levels implemented.
only used for debugging (and development) as well, logs are gone in production by default.

PR #311, opened 2025-10-24
feat: Add a Spring sample project

dustinbyrne:
#skip-changelog

This pull request adds a sample Spring Boot application demonstrating how to integrate the PostHog server-side SDK as a Spring-managed bean.

## Overview

This sample shows how to:

- Configure PostHog as a Spring Bean using dependency injection
- Inject the PostHog client into Spring controllers
- Capture events from REST API endpoints
- Configure PostHog via Spring properties

### DIFF
--- a/posthog-samples/posthog-spring-sample/.gitignore
+++ b/posthog-samples/posthog-spring-sample/.gitignore
@@ -1,0 +1,37 @@
+# Maven
+target/
+pom.xml.tag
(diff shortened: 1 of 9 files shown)


marandaneto (approved):
time to rename this repo to posthog-jvm/-java or something :D 
jvm isnt the best either because android runs on art which is not the same as jvm

PR #317, opened 2025-11-06
fix: Fallback for missing exception file names

dustinbyrne:

When compiling a Java application _without_ debug information (`javac -g:none`), filenames and line numbers are stripped from the compiled bytecode.

I've noticed that this causes the app to display "empty" frames:

image

A potential solution is to substitute the file name and line number with the class and method names:

image

It's worth mentioning that in the absence of file names and line numbers, this information _is_ still emitted in the exception stack trace as `module`, `function`. Is it appropriate to solve this here or elsewhere?
```json
{
    "event": "$exception",
    "distinct_id": "26142034-a912-4fcf-9175-b61330cb7366",
    "properties": {
        "$lib": "posthog-server",
        "$lib_version": "2.0.0",
        "$exception_level": "error",
        "$exception_list": [
            {
                "type": "RuntimeException",
                "mechanism": {
                    "handled": true,
                    "synthetic": false,
                    "type": "generic"
                },
                "thread_id": 1,
                "value": "This is a test exception",
                "module": "java.lang",
                "stacktrace": {
                    "frames": [
                        {
                            "module": "com.posthog.java.sample.PostHogJavaExample",
                            "function": "dangerousMethod",
                            "platform": "java",
                            "in_app": true
                        },
                        {
                            "module": "com.posthog.java.sample.PostHogJavaExample",
                            "function": "main",
                            "platform": "java",
                            "in_app": true
                        }
                    ],
                    "type": "raw"
                }
            }
        ],
        "service": "weather-api",
        "$process_person_profile": false
    },
    "timestamp": "2025-11-06T22:19:06.461Z",
    "uuid": "019a5b40-855d-77ca-8389-053fb609fbd8"
}
```

Tested against the java sample with `javac -g:none`. Also added some unit tests to simulate the same thing.

## :pencil: Checklist

### DIFF
--- a/posthog/src/main/java/com/posthog/internal/errortracking/ThrowableCoercer.kt
+++ b/posthog/src/main/java/com/posthog/internal/errortracking/ThrowableCoercer.kt
@@ -78,6 +78,18 @@ public class ThrowableCoercer {
                     val fileName = frame.fileName
                     if (fileName?.isNotEmpty() == true) {
                         myFrame["filename"] = fileName
+                    } else {
+                        // If the file name is not available (e.g., the application has been compiled without debug info),
+                        // substitute with class and method names.
+                        val frameClass = frame.className?.takeIf { it.isNotEmpty() }
+                        val frameMethod = frame.methodName?.takeIf { it.isNotEmpty() }
+                        val fullyQualifiedName =
+                            listOfNotNull(frameClass, frameMethod)
+                                .takeIf { it.isNotEmpty() }
+                                ?.joinToString(".")
+                        if (fullyQualifiedName?.isNotEmpty() == true) {
+                            myFrame["filename"] = fullyQualifiedName
+                        }
                     }
 
                     myFrame["in_app"] = isInApp(frame.className, inAppIncludes)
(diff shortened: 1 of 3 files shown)


posthog/src/main/java/com/posthog/internal/errortracking/ThrowableCoercer.kt
```diff
@@ -78,6 +78,18 @@ public class ThrowableCoercer {
...
+                        // substitute with class and method names.
+                        val frameClass = frame.className?.takeIf { it.isNotEmpty() }
+                        val frameMethod = frame.methodName?.takeIf { it.isNotEmpty() }
+                        val fullyQualifiedName =
+                            listOfNotNull(frameClass, frameMethod)
+                                .takeIf { it.isNotEmpty() }
+                                ?.joinToString(".")
+                        if (fullyQualifiedName?.isNotEmpty() == true) {
+                            myFrame["filename"] = fullyQualifiedName
+                        }
```
marandaneto: i think this should be a fallback on the frontend
also reported here https://posthog.slack.com/archives/C07AA937K9A/p1762352192041439?thread_ts=1762349699.084109&cid=C07AA937K9A
a reason would be that if the frame is  minified, it'd need to be symbolicated
another reason is that package+className != file name since they dont necessarily match, so its semantically wrong

PR #319, opened 2025-11-20
add sendFeatureFlagEvent in param getFeatureFlag

KopeikinaDarya:
the same problem https://github.com/wowrumal/posthog-ios/pull/3

My change addresses the behavior of sticky feature flags from [this PostHog tutorial](https://posthog.com/tutorials/sticky-feature-flags). When initially reading and assigning feature flags to a user, I want to avoid sending the $feature_flag_called event, since these evaluations are not real exposures yet.
If I filter out these events using the beforeSend block, I lose all analytics related to feature flag exposures—including the ones I actually want to track. This approach also breaks experiment accuracy, because I can no longer trigger exposure events exactly when needed.
My update ensures that exposure events are sent only at the appropriate time, keeping experiment tracking reliable and analytics data clean.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,10 @@ on:
     paths-ignore:
       - "**/*.md"
 
+permissions:
+  contents: read
+  actions: write  # for cancel-workflow-action
+
 jobs:
   cancel-previous-workflow:
     runs-on: ubuntu-latest
@@ -36,7 +40,7 @@ jobs:
           distribution: 'temurin'
 
       - name: Cache Gradle packages
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: |
             ~/.gradle/caches
--- a/.github/workflows/ci-security.yaml
+++ b/.github/workflows/ci-security.yaml
@@ -22,8 +22,9 @@ jobs:
             - name: Checkout code
               uses: actions/checkout@v6
             - name: Ensure SHA pinned actions
-              uses: zgosalvez/github-actions-ensure-sha-pinned-actions@9e9574ef04ea69da568d6249bd69539ccc704e74 # v4.0.0
+              uses: zgosalvez/github-actions-ensure-sha-pinned-actions@6124774845927d14c601359ab8138699fa5b70c3 # v4.0.1
               with:
                   allowlist: |
                       actions/
                       PostHog/
+                      github/
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,34 @@
+name: "CodeQL Advanced"
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  analyze:
+    name: Analyze Java/Kotlin
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+      packages: read
+      actions: read
+      contents: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v4
+        with:
+          languages: java
+          build-mode: autobuild
+          queries: security-and-quality
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v4
+        with:
+          category: "/language:java"
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -3,6 +3,10 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, edited, ready_for_review]
 
+permissions:
+  contents: read
+  pull-requests: write  # for danger to comment
+
 jobs:
   build:
     name: Changelog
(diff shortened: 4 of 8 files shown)


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -855,8 +855,9 @@ public class PostHog private constructor(
     public override fun isFeatureEnabled(
         key: String,
         defaultValue: Boolean,
+        sendFeatureFlagEvent:Boolean?
```
marandaneto:
```suggestion
        sendFeatureFlagEvent: Boolean?
```

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -884,7 +886,7 @@ public class PostHog private constructor(
             }
         }
 
-        if (config?.sendFeatureFlagEvent == true && shouldSendFeatureFlagEvent) {
+        if(sendFeatureFlagEvent == true || (config?.sendFeatureFlagEvent == true && shouldSendFeatureFlagEvent)){
```
marandaneto:
```suggestion
        if (sendFeatureFlagEvent == true || (config?.sendFeatureFlagEvent == true && shouldSendFeatureFlagEvent)) {
```

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -1258,12 +1260,14 @@ public class PostHog private constructor(
         public override fun isFeatureEnabled(
             key: String,
             defaultValue: Boolean,
-        ): Boolean = shared.isFeatureEnabled(key, defaultValue = defaultValue)
+            sendFeatureFlagEvent:Boolean?
```
marandaneto:
```suggestion
            sendFeatureFlagEvent: Boolean?
```

posthog/src/main/java/com/posthog/PostHog.kt
marandaneto: > In our case, we read the flag for the first time to send it to the sticky notes (ignoring the flag sending)

if you ignore sending the flag, the flag key wont be added to `featureFlagsCalled` so the next time `shouldSendFeatureFlagEvent` will be true and its not gonna break your use case

KopeikinaDarya: When posthog init, I have 
`config.sendFeatureFlagEvent = true`

First, I read flag’s values for sticky - call getFeatureFlag(“key” sendFeatureFlagEvent = false)
function sendFeatureFlagCalled() contains logic 

```
 var shouldSendFeatureFlagEvent = true
    synchronized(featureFlagsCalledLock) {
        val values = featureFlagsCalled[key] ?: mutableListOf()
        if (values.contains(value)) {
            shouldSendFeatureFlagEvent = false
        } else {
            values.add(value)
            featureFlagsCalled[key] = values
        }
    }
```

before capture event, and my flag’s key saved in featureFlagsCalled

After, I read flag’s values for a/b - call getFeatureFlag(“key” sendFeatureFlagEvent = true) and sendFeatureFlagCalled() function called.
My flag’s key contains in featureFlagsCalled and logic will ignore  (shouldSendFeatureFlagEvent will be true always)

`if ((sendFeatureFlagEvent == true || config?.sendFeatureFlagEvent == true) && shouldSendFeatureFlagEvent) {}`

marandaneto: @KopeikinaDarya i think the problem is with the `sendFeatureFlagCalled` method then
it should only append to the list with `featureFlagsCalled[key] = values` if the feature is actually enabled, that means `(sendFeatureFlagEvent == true || config?.sendFeatureFlagEvent == true)`

KopeikinaDarya:
will be more comments? @marandaneto @ioannisj

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -872,49 +873,52 @@ public class PostHog private constructor(
...
     ) {
-        var shouldSendFeatureFlagEvent = true
-        synchronized(featureFlagsCalledLock) {
-            val values = featureFlagsCalled[key] ?: mutableListOf()
-            if (values.contains(value)) {
-                shouldSendFeatureFlagEvent = false
-            } else {
-                values.add(value)
-                featureFlagsCalled[key] = values
+        if ((sendFeatureFlagEvent?:config?.sendFeatureFlagEvent) == true) {
```
ioannisj: nit: Not very readable to me tbh, maybe it' s better if we evaluate in a local var?
```kotlin
val effectiveSendFeatureFlagEvent = sendFeatureFlagEvent
    ?: config?.sendFeatureFlagEvent
    ?: false
```

KopeikinaDarya: updated

marandaneto (approved):

marandaneto:
@KopeikinaDarya i will merge and release this asap, right now we are closed to contributor PRs due to https://posthog.com/blog/nov-24-shai-hulud-attack-post-mortem , should be back later this week or early next week

PR #376, opened 2026-01-20
feat(odin): Allow collecting FCM device token in SDK & demonstrate it in sample app

odin-posthog:

We want to allow customers to send push notifications to their users in workflows. For that, we need to enable the SDK to collect and update the FCM device token that must be used for pushing notifications. 

We're not collecting it automatically when initializing so users that don't need this functionality aren't forced to install firebase dependencies, and they have control over which users have their tokens collected (eg. signed-in only)

Manually in my emulator through the sample app, and with some automated tests

## :pencil: Checklist

- [x] If there are related changes in the [core](https://github.com/PostHog/posthog-android/tree/main/posthog) package, I've already released them, or I'll release them before this one.


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -1307,6 +1370,8 @@ public class PostHog private constructor(
 
         private val apiKeys = mutableSetOf<String>()
 
+        private const val ONE_HOUR_IN_MILLIS = 60 * 60 * 1000L
```
marandaneto: Const usually goes within the companion object in kotlin

PR #385, opened 2026-01-22
feat: add evaluationContexts support to server PostHogConfig

igormq:
## Summary

- The server module's PostHogConfig was missing the `evaluationContexts` property that exists in the core PostHogConfig
- This meant that even when users set `evaluationContexts` in their server config, it wasn't being propagated to the core config and therefore wasn't sent in `/flags` API requests
- Adds `evaluationContexts` property to server PostHogConfig constructor
- Propagates `evaluationContexts` to core config in `asCoreConfig()`
- Adds `evaluationContexts()` method to the Builder class

---
Split from original PR addressing contributor feedback to keep changes focused.


dustinbyrne (dismissed):
Looks good

PR #423, opened 2026-02-18
fix: add extra validity checks during snapshots

ioannisj:

Helps with ticket: https://posthoghelp.zendesk.com/agent/tickets/50523 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/52576)

We have not seen this in Google SDK Console but a native crash was reported during snapshot generation. 
This adds a few extra checks on the view before attempting to generate a snapshot

## :pencil: Checklist

- [ ] If there are related changes in the [core](https://github.com/PostHog/posthog-android/tree/main/posthog) package, I've already released them, or I'll release them before this one.


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -403,6 +403,13 @@ public class PostHogReplayIntegration(
         val status = decorViews[view] ?: return
         val window = windowRef.get() ?: return
 
+        // Check view is still valid before any access to avoid native crashes
+        if (!view.isAliveAndAttachedToWindow()) return
+        if (view.width <= 0 || view.height <= 0) return
```
marandaneto: toScreenshotWireframe reads width and height
instead of doing this here then we just return early if its zeroed, maybe it'd even make sense to be part of "isVisible"

marandaneto (approved):

PR #439, opened 2026-02-24
feat: add postHogUnmask modifier and ph-no-mask support

marandaneto:
Add a new `postHogUnmask` modifier that explicitly prevents masking of elements in session replay, even when `maskAllTextInputs` or `maskAllImages` is enabled. It takes precedence over global masking config and `postHogMask`.

```kotlin
import com.posthog.android.replay.PostHogMaskModifier.postHogUnmask

Text(
    text = "Always visible in replay",
    modifier = Modifier.postHogUnmask()
)
```

will update https://posthog.com/docs/session-replay/privacy?tab=Android once i have a stamp

running the sample

## :pencil: Checklist

### If releasing new changes

- [X] Ran `pnpm changeset` to generate a changeset file
- [X] Added the "release" label to the PR to indicate we're publishing new versions for the affected packages

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -647,7 +647,7 @@ public class PostHogReplayIntegration(
 
     private fun View.isAnyInputSensitive(ancestorUnmasked: Boolean = false): Boolean {
         if (ancestorUnmasked || isUnmasked()) return false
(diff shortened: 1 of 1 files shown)


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -964,10 +999,12 @@ public class PostHogReplayIntegration(
     }
 
     private fun ImageView.shouldMaskImage(): Boolean {
+        if (isUnmasked()) return false
```
ioannisj: Do we need to do this for each view type? We have an early check in findMaskableWidgets, should we add a similar one to toWireframe?

PR #476, opened 2026-04-02
add: getAllFeatureFlagDetails method

### LINKED ISSUE #449: Add getAllFeatureFlags method
### Problem Statement

It would be great to have a getAllFeatureFlags method available in the PostHog Android SDK, similar to some server-side SDKs like Python:
https://github.com/PostHog/posthog-python/blob/master/posthog/client.py

Without such a method, client-side discovery of feature flags is not possible, since the SDK currently requires knowing flag keys in advance (e.g., when calling isFeatureEnabled("flag_key")). This means applications must hardcode or otherwise know all possible feature flags beforehand.
Implementing it would among other things enable dynamic client-side feature configuration without hardcoding all flag keys.

### Solution Brainstorm

_No response_

MohitMandalia:

Fixes: #449

Yes wrote unit tests

## :pencil: Checklist

### If releasing new changes


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -1136,6 +1136,22 @@ public class PostHog private constructor(
...
+        val flags = remoteConfig?.getFeatureFlags()
+        val results =
+            flags?.map { (key, value) ->
+                val (enabled, variant) =
+                    when (value) {
+                        is Boolean -> value to null
+                        is String -> true to value
+                        else -> false to null
+                    }
+                FeatureFlagResult(key, enabled, variant, null)
```
dustinbyrne: We should populate the payload instead of always passing `null`

dustinbyrne (commented):
Thanks for putting this together, @MohitMandalia. I've added some comments below!

marandaneto (approved):
see comments otherwise lgtm

.changeset/polite-memes-look.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog": minor
```
marandaneto, line 2: you also need to add for posthog-android since its the SDK that people install

PR #481, opened 2026-04-08
feat: add device bucketing support

dmarticus:
## Summary

- Adds a persistent `$device_id` to the Android SDK for stable device-level feature flag bucketing across identity changes
- The device ID is seeded from the anonymous ID on first init, persists across `identify()` and `reset()` calls, and is sent in `/flags` request payloads
- Ports the equivalent feature from [posthog-js#3340](https://github.com/PostHog/posthog-js/pull/3340) (React Native SDK) to Android

## Problem

Users running experiments targeting anonymous users see variant assignments flip when `identify()` or `reset()` changes the `distinct_id`. The hash changes, causing a different bucket and different variant for the same physical device.

## Changes

- **`PostHogPreferences`**: Add `DEVICE_ID` constant and include in `ALL_INTERNAL_KEYS`
- **`PostHogFlagsRequest`**: Add optional `deviceId` parameter, serialized as `$device_id`
- **`PostHogApi.flags()`**: Thread `deviceId` parameter to `PostHogFlagsRequest`
- **`PostHogRemoteConfig`**: Read `deviceId` from `cachePreferences` in `executeFeatureFlags()` and pass to API
- **`PostHog.kt`**:
  - `initDeviceId()` — seeds device_id from anonymous ID on first setup
  - `reset()` — preserves `DEVICE_ID` across resets (alongside `VERSION` and `BUILD`)
  - `getDeviceId()` — public method with lazy-init fallback for SDK upgrades
- **`PostHogInterface`**: Add `getDeviceId()` to public API surface

## Test plan

- [x] 3 new `PostHogFlagsRequestTest` tests (includes/excludes device_id)
- [x] 7 new `PostHogTest` tests (init, persistence, identify/reset preservation, flags payload, lazy-init upgrades)
- [x] All existing core tests pass (116 total)
- [x] All existing server tests pass

### DIFF
--- a/.changeset/device-bucketing.md
+++ b/.changeset/device-bucketing.md
@@ -1,0 +1,5 @@
+---
+"posthog": minor
+---
+
+feat: add device bucketing support for stable feature flag assignment across identity changes
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -10,6 +10,7 @@ import com.posthog.internal.PostHogOnRemoteConfigLoaded
 import com.posthog.internal.PostHogPreferences.Companion.ALL_INTERNAL_KEYS
 import com.posthog.internal.PostHogPreferences.Companion.ANONYMOUS_ID
 import com.posthog.internal.PostHogPreferences.Companion.BUILD
+import com.posthog.internal.PostHogPreferences.Companion.DEVICE_ID
 import com.posthog.internal.PostHogPreferences.Companion.DISTINCT_ID
 import com.posthog.internal.PostHogPreferences.Companion.GROUPS
 import com.posthog.internal.PostHogPreferences.Companion.IS_IDENTIFIED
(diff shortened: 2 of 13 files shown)


marandaneto (approved):

marandaneto:
@dmarticus missing changesets

.changeset/device-bucketing.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog": minor
```
marandaneto, line 2: @dmarticus you also need to release the android sdk here otherwise it wont have any effect

marandaneto (approved):
missing changeset for the android sdk otherwise LGTM

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -175,6 +177,11 @@ public class PostHog private constructor(
 
                 legacyPreferences(config, config.serializer)
 
+                // Initialize device_id if not already set. getDeviceId() handles lazy init
+                // by seeding from the anonymous ID, providing a stable identifier for
+                // device-level feature flag bucketing that survives identify() and reset().
+                getDeviceId()
```
marandaneto, line 183: @dmarticus does this have any effect since the sdk is still not enabled? This should be called after enabled=true

PR #482, opened 2026-04-09
feat: session replay minimum recording duration

ioannisj:

Adds support for the minimumDurationMilliseconds remote config setting for session replay and follows the implementation as described in https://github.com/PostHog/posthog-ios/pull/548

## :pencil: Checklist

### If releasing new changes


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayQueue.kt
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * A replay queue that wraps an inner queue (for actual API sends) and a
+ * [PostHogReplayBufferQueue] (for buffering snapshots until minimum session
+ * duration is met).
+ *
+ * The queue is passive — it delegates all buffering decisions to a
+ * [PostHogReplayBufferDelegate]. When [PostHogReplayBufferDelegate.isBuffering]
+ * is true, snapshots are routed to the buffer and `flush()` calls are suppressed.
+ */
+public class PostHogReplayQueue internal constructor(
```
dustinbyrne: Can this be internal? Does it need to be public?

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayQueue.kt
```diff
@@ -0,0 +1,108 @@
...
+        get() = replayDir?.listFiles()?.size ?: 0
+
+    override fun add(event: PostHogEvent) {
+        if (bufferDelegate?.isBuffering == true) {
+            bufferQueue.add(event)
+            config.logger.log("Buffered replay event '${event.event}'. Buffer depth: ${bufferQueue.depth}")
+            bufferDelegate?.onReplayBufferSnapshot(this)
+        } else {
+            innerQueue.add(event)
+        }
```
dustinbyrne: This is a synchronous write to disk. Perhaps this queue should be constructed with an `ExecutorService` similar to `PostHogQueue` and serialize buffer writes through it to avoid blocking the calling thread?

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -1603,6 +1624,8 @@ public class PostHogReplayIntegration(
 
         if (!resumeCurrent) {
             clearSnapshotStates()
+            // Reset minimum duration buffering state for the new session
+            resetBufferingState()
```
dustinbyrne: `start` is only called once globally afaik, so it seems like this call would be better off in `onSessionIdChanged`?

dustinbyrne (commented):
Got through some of this but I have to step away. I'll circle back when I can if nobody else gets to it first!

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayQueue.kt
marandaneto: unless the caller is guaranteed to be within an ExecutorService already, then i'd add a comment to make that clear

marandaneto:
@turnipdabeets pls review!

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
ioannisj: Right, we were not resetting buffer onSessionIdChanged. We now track current session id and reset when the session actually changes

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayQueue.kt
ioannisj: Updated PostHogReplayQueue to take the replay ExecutorService and serialize buffered disk writes through it, similar to PostHogQueue. The buffering state is also rechecked inside the executor before writing, so events won’t be incorrectly buffered if the minimum-duration threshold flips while the task is queued.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayQueue.kt
ioannisj: Fixed and made this internal.  It ended up public because `PostHogReplayIntegration` is already a public class, and PostHogReplayQueue was used in a public method when implementing `PostHogReplayBufferDelegate`. 

`PostHogReplayIntegration` should be internal as well though? I checked hybrid SDKs, and it seems that this class is not used there directly. So, unless I'm missing something (cc @marandaneto), I would also change replay integration to internal since I think we exposed it publicly accidentally

turnipdabeets (approved):
LGTM, it might be worth adding a few more tests though. Seems like iOS has happy-path migration coverage (PostHogQueue target, file move, ordering, concurrent writes during migration) that this PR doesn't.

PR #491, opened 2026-04-17
chore: attach release info to proguard mapping upload

turnipdabeets:

The posthog-cli's `exp proguard upload` accepts `--release-name`, `--release-version`, and `--build` (via flattened `ReleaseArgs`), but this plugin only passed `--path` and `--map-id`. As a result, proguard uploads either fell back to git-derived release info (repo name + commit SHA) or created no release association at all — not the app's actual release identity.

iOS (posthog-ios#558) and React Native (posthog-js) already attach this info on their symbol/sourcemap uploads. This PR brings Android to parity.

Captura de pantalla 2026-04-17 a la(s) 1 27 59 p m

Built `:posthog-samples:posthog-android-sample:assembleRelease` twice against environment 381971:

**Before:**
```
posthog-cli exp proguard upload --path ... --map-id bbb498d5...
```
→ CLI auto-created release `posthog-android@eac1385...` (git repo name + commit SHA).

**After (with `versionCode=2`, `versionName="1.1"` for validation):**
```
posthog-cli exp proguard upload --path ... --map-id bbb498d5... \
  --release-name com.posthog.android.sample --release-version 1.1 --build 2
```
→ Release associated with the app's actual identity.

Empty-string values from the variant are filtered out so the CLI still falls back to git-derived info when `applicationId` / `versionName` / `versionCode` aren't set.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.changeset/attach-release-info-to-proguard-upload.md
+++ b/.changeset/attach-release-info-to-proguard-upload.md
@@ -1,0 +1,5 @@
+---
+"posthog-android": minor
+---
+
+Attach release info (`applicationId`, `versionName`, `versionCode`) to proguard mapping uploads via the new posthog-cli `--release-name`, `--release-version`, and `--build` flags.
--- a/posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
+++ b/posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
@@ -109,12 +109,16 @@ internal class PostHogAndroidGradlePlugin : Plugin<Project> {
         variant: ApplicationVariant,
         mappingFiles: Provider<FileCollection>,
     ): TaskProvider<PostHogUploadProguardMappingsTask> {
+        val primaryOutput = variant.outputs.firstOrNull()
         val uploadMapIdTask =
             PostHogUploadProguardMappingsTask.register(
                 project = project,
                 generateMapIdTask = generateMapIdTask,
                 mappingFiles = mappingFiles,
                 taskSuffix = variant.name.capitalizeUS(),
+                releaseName = variant.applicationId,
+                releaseVersion = primaryOutput?.versionName?.map { it.orEmpty() },
+                build = primaryOutput?.versionCode?.map { it?.toString().orEmpty() },
             )
         return uploadMapIdTask
     }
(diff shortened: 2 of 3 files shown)


posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
```diff
@@ -109,12 +109,16 @@ internal class PostHogAndroidGradlePlugin : Plugin<Project> {
...
+        val primaryOutput = variant.outputs.firstOrNull()
         val uploadMapIdTask =
             PostHogUploadProguardMappingsTask.register(
                 project = project,
                 generateMapIdTask = generateMapIdTask,
                 mappingFiles = mappingFiles,
                 taskSuffix = variant.name.capitalizeUS(),
+                releaseName = variant.applicationId,
+                releaseVersion = primaryOutput?.versionName?.map { it.orEmpty() },
+                build = primaryOutput?.versionCode?.map { it?.toString().orEmpty() },
```
marandaneto: can this be a number? if so, would that work anyway?

.changeset/attach-release-info-to-proguard-upload.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-android": patch
```
marandaneto: since we didnt have it at all, maybe a minor makes more sense?

marandaneto (approved):
missing the release label? LGTM

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
turnipdabeets: I think makes more sense as String since versionCode is nullable in AGP. If it's not set, empty just drops the flag. With Int we'd have to pick some fake value like 0. This also follows what we do on iOS and RN too.

marandaneto: what i meant is if `primaryOutput?.versionCode` can be a number since i know it was at some point depending on the AGP version, and if it is, `.map { it?.toString().orEmpty() }` should not fail which i believe it doesnt anyway
i agree with it being a string in the CLI and PostHog

turnipdabeets: Good call out, I updated to `primaryOutput?.versionCode`. I did some testing and didn't realize when it's empty that AGP actually sets a negative number so as a String you'd see -1 which would be strange!

PR #470, opened 2026-03-26
fix: Enforce a 24-hour maximum session duration

marandaneto:
## :bulb: Motivation and Context

Enforce a 24-hour maximum session duration. When a session exceeds 24 hours, it is automatically rotated to a new one. This prevents excessively long sessions from accumulating unbounded data.

Closes https://github.com/PostHog/posthog-android/issues/220

## :green_heart: How did you test it?

- Added unit tests for `rotateSession()`, `getSessionStartedAt()`, and `sessionStartedAt` lifecycle in `PostHogSessionManagerTest`.
- Added integration tests in `PostHogLifecycleObserverIntegrationTest` verifying:
  - Session is rotated when it exceeds 24 hours.
  - Session is not rotated when under 24 hours.
  - Session rotation is skipped for React Native (RN manages its own sessions).
- Used `PostHogDateProvider` for time mocking to make tests deterministic.
- Ran `make checkFormat` — all formatting checks pass.

## :pencil: Checklist

- [x] I reviewed the submitted code.
- [x] I added tests to verify the changes.
- [ ] I updated the docs if needed.
- [x] No breaking change or entry added to the changelog.

### If releasing new changes

- [x] Ran `pnpm changeset` to generate a changeset file
- [x] Added the "release" label to the PR to indicate we're publishing new versions for the affected packages

PR #494, opened 2026-04-21
fix: 24-hour max + 30-min inactivity session rotation, align with iOS

### LINKED ISSUE #220: Enforce max session duration
### Description

Similar to https://github.com/PostHog/posthog-ios/issues/245

- [x] [iOS](https://github.com/PostHog/posthog-ios/issues/245)
- [ ] Android
- [ ] React Native
- [ ] Flutter

turnipdabeets:

Brings Android session-id behavior into parity with iOS: 30-min inactivity rotation, 24-hour max-duration cap, foreground/background-aware rotate-or-clear, sampling-aware replay restart on rotation, and replay-independent touch-driven activity tracking.

Continues @marandaneto's work from #470 (auto-closed by stale-bot; branch was rebased + force-pushed so GitHub wouldn't allow reopening).

Closes https://github.com/PostHog/posthog-android/issues/220 for Android.

What's in scope:

- Inline expiry check in `PostHogSessionManager.getActiveSessionId()` (mirrors iOS's `getSessionId(readOnly:)`); foreground rotates, background clears.
- New `touchSession()` — refresh-or-rotate on user activity. Called from lifecycle, the new replay-independent touch interceptor, and `capture()`.
- New `PostHogTouchActivityIntegration` — touch-driven activity tracking decoupled from session replay, so behavior is consistent regardless of replay sampling.
- New `PostHogInterface.restartSessionReplay()` (`@PostHogInternal`) — sampling-aware restart that does not rotate the session id, so the listener path can avoid double-rotation when invoked from a silent rotation.
- New `PostHogSessionManager.peekSessionId()` — read-only sibling of `getActiveSessionId()`; lets the replay listener inspect the new id without re-entering the mutating getter.
- `setSessionId(sameId)` is now a no-op (Android-specific divergence from iOS, requested in review). Without it, RN/Flutter re-asserting the same id on every event would reset the 24-hour and inactivity clocks and defeat the cap.
- The session-id-changed listener now fires from `startSession`/`endSession`/`setSessionId`, not only silent getter rotations.
- `isAppInBackground` defaults to `true` (iOS parity) so an expired session before the first `onStart` is cleared rather than silently rotated.

In-PR follow-ups landed during review:

- Removed a stale direct call `sessionReplayHandler?.onSessionIdChanged()` in `PostHog.startSession()` that was producing a duplicate listener fire on every bg→fg now that the manager fires the listener itself.
- After `start(false)` clears snapshot states, post `postInvalidate()` on each tracked decor view so the new session emits its `type:4` meta + full-wireframe keyframes immediately rather than on the next user-driven onDraw (was ~49 s on a static UI; now <1 s).
- Eliminated a redundant `MutableSet<View>` tracker in `PostHogTouchActivityIntegration` — Curtains' `rootViews` and each window's `touchEventInterceptors` are already authoritative.
- Wrapped the `onSessionIdChanged` listener invocation in try/catch so a misbehaving replay handler can't propagate into session-mutating call sites.
- Marked `PostHogSessionManager.dateProvider` `@Volatile` since it's set once at setup and read from any thread under the session lock.

Unit tests in `PostHogSessionManagerTest`, `PostHogTest`, `PostHogLifecycleObserverIntegrationTest`, `PostHogReplayIntegrationTest`, and `PostHogTouchActivityIntegrationTest` cover:

- `startSession`/`endSession`/`setSessionId` listener-fire semantics (including `setSessionId(sameId)` no-op)
- `touchSession` refresh, rotate-on-idle, bg/RN/no-session no-ops
- `getActiveSessionId` rotate-fg vs clear-bg for both inactivity and max-duration
- `peekSessionId` read-only contract
- `PostHog.restartSessionReplay` — SDK closed, replay flag disabled, active+sampling-pass, inactive+sampling-pass, sampling-fail, no-rotation guarantee
- `PostHogReplayIntegration.onSessionIdChanged` fires `restartSessionReplay` regardless of `isSessionReplayActive`; no-op when session is cleared
- `PostHogLifecycleObserverIntegration` — 24h foreground rotation, RN bypass, replay restart on bg→fg, `onStop` touches session before flipping bg

`make checkFormat` and `./gradlew :posthog:test :posthog-android:testDebugUnitTest :posthog:apiCheck :posthog-android:apiCheck` pass.

Manual tests on Pixel 4 (Android 13) with shortened constants (2-min max, 30-s inactivity):

- ✅ 30-s foreground idle → rotation; different `$session_id` before/after; `[Session Replay] Session changed` log fires.
- ✅ Active foreground (taps every 20 s for 60 s+) → same `$session_id` throughout. Touch decoupling refreshes activity even with replay sampled out.
- ✅ Background 35 s → session cleared (timer fires `endSession` during bg, `[Session Replay] Session cleared` log); foreground returns under fresh `$session_id`.
- ✅ Continuous foreground past 2-min cap → session rotates mid-stream; new session emits `type:4` + full-wireframe keyframes within ~1 s of rotation.
- ✅ With `sessionReplayConfig.sampleRate = 0.5` and 6 rotations: roughly half passed (events shipped), half failed (`Sample rate (0.5) has determined that this sessionId X will not be sent`), confirming `restartSessionReplay` re-rolls sampling each time.

<details>
<summary>Manual test setup</summary>

Temporarily flip both constants in `PostHogSessionManager.kt` and the lifecycle observer:

```kotlin
// posthog/src/main/java/com/posthog/internal/PostHogSessionManager.kt
private const val SESSION_MAX_DURATION = (1000L * 60 * 2)   // 2 min for testing
private const val SESSION_INACTIVITY_DURATION = (1000L * 30) // 30 s for testing

// posthog-android/.../PostHogLifecycleObserverIntegration.kt
private val sessionMaxInterval = (1000L * 30) // 30 s for testing
```

Build + install: `./gradlew :posthog-samples:posthog-android-sample:installDebug`

Watch logs: `adb logcat -s PostHog`

After testing, revert constants to 24 h / 30 min.

</details>

## :pencil: Checklist

### If releasing new changes


dustinbyrne:
Slight behavioral drift vs. iOS:

In iOS when we retrieve the active session ID, it checks to see if it's still valid before returning the identifier. If it's not, it's immediately rotated. https://github.com/PostHog/posthog-ios/blob/c165ead50ecf66368685a8467fa625fc0960bace/PostHog/PostHogSessionManager.swift#L109-L114

This differs from the current implementation in that we only rotate the session on a fg/bg transition. 

iOS calls this getter when building event properties, so it can happen during any event transmission: https://github.com/PostHog/posthog-ios/blob/c165ead50ecf66368685a8467fa625fc0960bace/PostHog/PostHogSDK.swift#L421-L423

There's a test that validates this, it's probably worth porting over: https://github.com/PostHog/posthog-ios/blob/c165ead50ecf66368685a8467fa625fc0960bace/PostHogTests/PostHogSessionManagerTest.swift#L282-L283

marandaneto:
@ioannisj implemented this on iOS so worth an extra pair of eyes (this also affects Flutter iirc)

posthog/src/main/java/com/posthog/internal/PostHogSessionManager.kt
```diff
@@ -28,6 +63,7 @@ public object PostHogSessionManager {
         synchronized(sessionLock) {
             if (sessionId == sessionIdNone) {
                 sessionId = TimeBasedEpochGenerator.generate()
+                sessionStartedAt = dateProvider?.currentTimeMillis() ?: System.currentTimeMillis()
```
ioannisj: I think we may want to update `sessionStartedAt` for custom sessions through `setSessionId()` as well

posthog/src/main/java/com/posthog/internal/PostHogSessionManager.kt
```diff
@@ -40,13 +76,60 @@ public object PostHogSessionManager {
...
     public fun getActiveSessionId(): UUID? {
+        var sessionChanged = false
         var tempSessionId: UUID?
         synchronized(sessionLock) {
-            tempSessionId = if (sessionId != sessionIdNone) sessionId else null
+            if (sessionId == sessionIdNone || isReactNative) {
+                tempSessionId = if (sessionId != sessionIdNone) sessionId else null
+            } else {
+                val now = dateProvider?.currentTimeMillis() ?: System.currentTimeMillis()
+                val expired = sessionStartedAt > 0L && (sessionStartedAt + SESSION_MAX_DURATION) <= now
```
ioannisj: On iOS we are also checking for 30min inactivity here. If I remember correctly, android detects this purely through background/foreground event. iOS has an activity timestamp tracking through [touchSession()](https://github.com/PostHog/posthog-ios/blob/c165ead50ecf66368685a8467fa625fc0960bace/PostHog/PostHogSessionManager.swift#L159) which basically rotates the session if the user has not been interacting with the app in 30mins (background or not).  Would be good to close this gap as well between the two

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -190,6 +190,10 @@ public class PostHog private constructor(
 
                 queue.start()
 
+                PostHogSessionManager.setOnSessionIdChangedListener {
+                    sessionReplayHandler?.onSessionIdChanged()
```
ioannisj: I don't see any changes on resetting state for session replay inside onSessionIdChanged() (unless it's already properly handled), since this is now a new recording and we'll need metadata/full snapshot. Can you please test how replay behaves with session rotations (maybe set max session length to a couple of mins locally)

posthog/src/main/java/com/posthog/internal/PostHogSessionManager.kt
turnipdabeets: added!

posthog/src/main/java/com/posthog/internal/PostHogSessionManager.kt
turnipdabeets: Good catch! 

Added touchSession() mirroring iOS, wired into lifecycle onStart, replay touch listener, and capture() (the last one is an Android-specific deviation since we don't have UIEvent swizzling). Verified manually with shortened timers: foreground idle rotates, active foreground doesn't, brief bg preserves session.

posthog/src/main/java/com/posthog/PostHog.kt
turnipdabeets: Updated! onSessionIdChanged() now does stop() + start(resumeCurrent = false) on rotation (which clears per-view snapshot state so the next onDraw emits fresh type-4 + type-2), or stop() on bg-clear. Verified end-to-end with shortened constants: bg→fg rotation produces clean keyframes immediately and mid-foreground rotation produces keyframes on the next visible redraw.

turnipdabeets:
@ioannisj when you get a moment, mind giving this another review?

posthog/src/main/java/com/posthog/internal/PostHogSessionManager.kt
```diff
@@ -16,45 +17,176 @@ public object PostHogSessionManager {
...
+            onSessionIdChangedListener?.invoke()
+        }
+    }
+
     public fun setSessionId(sessionId: UUID) {
-        // RN can only set its own session id directly
         synchronized(sessionLock) {
+            val now = dateProvider?.currentTimeMillis() ?: System.currentTimeMillis()
             this.sessionId = sessionId
+            // Stamp the start so an externally-set session participates in the 24h
```
ioannisj: We probably only need to do this if this.sessionId != sessionId (before assigning right above)

posthog/src/main/java/com/posthog/internal/PostHogSessionManager.kt
ioannisj: > wired into lifecycle onStart, replay touch listener, and capture()

Awesome. For replay touch listener, does this mean that we'll get a different behavior on session rotation when replay is not enabled? Session id affects [session-metrics calcs](https://posthog.com/tutorials/session-metrics) so it would be good to decouple.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
ioannisj: I think this can lead to infinite recursion since getSessionId() is mutating, and we are reacting to a new session id change? Not sure how much of an edge case this can be, if at all, so happy for a push back here. I feel a read-only operation on PostHogSessionManager would be safer though

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -1678,6 +1681,20 @@ public class PostHogReplayIntegration(
                 config.logger.log("[Session Replay] Session changed. Stopping until trigger is matched.")
                 stop()
             }
+        } else if (isSessionReplayActive) {
```
ioannisj: I think we should react even if session replay is not active. Replay may be enabled in config but Session A may not have been sampled and not started. Now that we rotate to session B, sampling may return true and session replay should start?

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -1678,6 +1681,20 @@ public class PostHogReplayIntegration(
...
+            // Posting to main: getter can be invoked from any thread that calls capture(),
+            // and start(resumeCurrent = false) iterates a non-thread-safe WeakHashMap.
+            if (currentSessionId == null) {
+                config.logger.log("[Session Replay] Session cleared. Stopping recording.")
+                mainHandler.handler.post { stop() }
+            } else {
+                config.logger.log("[Session Replay] Session changed. Re-initializing recording for new session.")
+                mainHandler.handler.post {
+                    stop()
+                    start(resumeCurrent = false)
```
ioannisj: Similar to the comment above, this will skip sampling check and start the session anyway? Maybe we can route through PostHog.startSessionReplay() which checks sampling?

posthog/src/main/java/com/posthog/internal/PostHogSessionManager.kt
turnipdabeets: added - Worth flagging that iOS always re-stamps sessionStartTimestamp/sessionActivityTimestamp and always fires onSessionIdChanged, maybe we should add this check to iOS as well? I think repeated setSessionId(sameId) calls (e.g. RN/Flutter re-asserting the current id on every event) reset both the 24h max-duration and 30-min inactivity clocks so sessions effectively never expire on iOS, which defeats the whole point of those thresholds. Is that right?

posthog/src/main/java/com/posthog/internal/PostHogSessionManager.kt
turnipdabeets: Decoupled. Created PostHogTouchActivityIntegration which lives outside replay and wires its own Curtains touch interceptor → PostHogSessionManager.touchSession(). The replay integration's touch listener no longer touches the session manager, only does mouse-interaction snapshots.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
turnipdabeets: Good call. Added PostHogSessionManager.peekSessionId() (read-only, skips the expiry checks) and onSessionIdChanged now reads the current id via peekSessionId() instead of postHog.getSessionId()

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
turnipdabeets: Done. The rotation branch in onSessionIdChanged no longer gates on isSessionReplayActive — it proceeds to the restart path even when replay was previously inactive, so a session that was sampled out before may now pass sampling on the new id.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
turnipdabeets: Routed through a sampling-aware path. Added PostHogInterface.restartSessionReplay() which will: 
1. bails if SDK disabled or replay flag disabled,
2. stops if active,
3. runs shouldRecordSession() (the same sampling decision as startSessionReplay),
4. on pass, calls start(false) to clear snapshot state for fresh keyframes.

posthog/src/main/java/com/posthog/PostHogInterface.kt
```diff
@@ -216,6 +216,18 @@ public interface PostHogInterface : PostHogCoreInterface {
...
+     * Stops any in-progress recording, re-runs sampling for the current session id, and
+     * resumes recording with fresh meta + full-snapshot keyframes if sampling passes.
+     *
+     * Unlike [startSessionReplay] with `resumeCurrent = false`, this does not rotate the
+     * session id, so it's safe to call from a path that fired because the session already
+     * rotated (avoids a double-rotation).
+     */
+    @PostHogInternal
+    public fun restartSessionReplay() {
+    }
```
dustinbyrne: Is it possible to avoid placing an internal method here? IMO anything marked internal but public on this interface might as well be public

dustinbyrne: Strongly recommend against adding publicly internal methods to this interface. This is the core interface provided by the SDK, so even if marked internal, it's highly visible.

posthog-android/src/main/java/com/posthog/android/internal/PostHogLifecycleObserverIntegration.kt
```diff
@@ -98,14 +117,35 @@ internal class PostHogLifecycleObserverIntegration(
...
+        if (wasExpired) {
+            cancelTask()
+            val wasReplayActive = postHog?.isSessionReplayActive() == true
+            postHog?.endSession()
+            // Synchronous stop guarantees replay is torn down before the process suspends;
+            // the listener-driven path posts to main and may not run in time.
+            postHog?.stopSessionReplay()
+            replayActiveBeforeRotation.set(wasReplayActive)
+            // Zeroing forces the next onStart into the "create a fresh session" branch.
+            lastUpdatedSession.set(0L)
```
dustinbyrne: Should this class still be inferring session state from its own timestamp?

My concern is that this could drift from `PostHogSessionManager`

dustinbyrne (commented):
Overall the logic in this change seems good. I've left a couple of comments below, but mainly just concerned about the first.

posthog/src/main/java/com/posthog/PostHogInterface.kt
turnipdabeets: refactored without internal method

posthog-android/src/main/java/com/posthog/android/internal/PostHogLifecycleObserverIntegration.kt
turnipdabeets: good call out, refactored this as well.

posthog-android/src/test/java/com/posthog/android/internal/PostHogLifecycleObserverIntegrationTest.kt
dustinbyrne: nit: comment is describing stale behavior

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -1662,22 +1670,53 @@ public class PostHogReplayIntegration(
...
+                if (isSessionReplayActive) stop()
+                return@post
+            }
+            if (remoteConfig.makeSamplingDecision(currentSessionId).not()) {
+                if (isSessionReplayActive) stop()
+                return@post
+            }
+            if (isSessionReplayActive) stop()
+            start(resumeCurrent = false)
         }
```
dustinbyrne, line 1727: In iOS, session replay is only started automatically if `config.sessionReplay` is true (this is a simplification, we actually won't install the integration if it's false).

On main, we don't start automatically recording at any point unless `config.sessionReplay` is true (`isSessionReplayConfigEnabled()`).

This `onSessionIdChanged` implementation _will_ start automatic recording on session rotation if session replay is enabled via remote config. It doesn't automatically start on initialization because the listener isn't bound yet.

dustinbyrne (commented):
One more thing worth addressing

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
dustinbyrne: Worth a test case or two

turnipdabeets: Added a check for config.sessionReplay plus a test

posthog-android/src/test/java/com/posthog/android/internal/PostHogLifecycleObserverIntegrationTest.kt
turnipdabeets: removed

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -476,6 +492,8 @@ public class PostHog private constructor(
                 config?.logger?.log("PostHog is in OptOut state.")
                 return
             }
+            // Any capture counts as activity so an idle session rotates before its id is read.
+            PostHogSessionManager.touchSession()
```
ioannisj: Not 100% sure this should be the case here. We could be capturing without active user engagement (e.g replay network snapshots). Also the event captured could have a custom $session_id, in which case we should not be touching the current session. 

I'll leave it up to you but I would personally just consider gestures and/or backgrounding-foregrounding the app as points to touch the current session

marandaneto (approved):

marandaneto:
either before or after merging, make sure that this does not break anything on react native and flutter since they use some of those classes eg the session manager etc

posthog/src/main/java/com/posthog/internal/PostHogSessionManager.kt
```diff
@@ -16,45 +16,187 @@ public object PostHogSessionManager {
...
+     */
+    private var sessionActivityTimestamp: Long = 0L
+
     @Volatile
     public var isReactNative: Boolean = false
 
+    // Defaults to true so an expired session before the first onStart is cleared rather
+    // than silently rotated against a process that has no UI yet.
+    @Volatile
+    private var isAppInBackground: Boolean = true
```
ioannisj: We default to true here, but what happens if this is not Android (so we never set to false?) I imagine the session will be cleared and not rotated after 30mins right?

posthog-android/src/main/java/com/posthog/android/internal/PostHogLifecycleObserverIntegration.kt
```diff
@@ -93,19 +89,35 @@ internal class PostHogLifecycleObserverIntegration(
...
+            timer.schedule(timerTask, bgEndSessionDelayMs)
         }
     }
 
     override fun onStop(owner: LifecycleOwner) {
+        val currentTimeMillis = config.dateProvider.currentTimeMillis()
+        // Capture expiry before flipping the bg flag: once bg=true, getActiveSessionId
+        // would clear an expired session and zero sessionStartedAt, hiding it from the
+        // wasExpired branch below.
+        val wasExpired = PostHogSessionManager.isSessionExceedingMaxDuration(currentTimeMillis)
```
ioannisj, line 101: Will this return false on RN (since any session rotation logic should be no-op for RN)?

posthog/src/main/java/com/posthog/PostHog.kt
turnipdabeets: Done — removed touchSession() from capture(); activity tracking is now strictly gestures (PostHogTouchActivityIntegration) + lifecycle bg/fg. Replay snapshots
and caller-provided $session_id events no longer refresh the SDK's session timestamp.

posthog-android/src/main/java/com/posthog/android/internal/PostHogLifecycleObserverIntegration.kt
turnipdabeets: Good catch — added the RN bypass (returns false when isReactNative)

posthog/src/main/java/com/posthog/internal/PostHogSessionManager.kt
turnipdabeets: I did an audit and it's only affecting Android, but switched to false and explicitly set android setAppInBackground(true) so that it's more clear.

turnipdabeets:
> either before or after merging, make sure that this does not break anything on react native and flutter since they use some of those classes eg the session manager etc


> either before or after merging, make sure that this does not break anything on react native and flutter since they use some of those classes eg the session manager etc

Tested on RN and Flutter and confirmed they don't break

ioannisj (approved):

ioannisj:
Let's update https://github.com/PostHog/posthog/issues/30889 once this is out as well

PR #495, opened 2026-04-21
feat: add tracing headers to OkHttp requests

dustinbyrne:

Add Android support for PostHog tracing headers on outgoing OkHttp requests, matching the recent iOS work and bringing the Android SDK closer to browser SDK parity for request correlation.

This adds a configurable `addTracingHeaders` host list and injects `X-POSTHOG-DISTINCT-ID` and `X-POSTHOG-SESSION-ID` on exact hostname matches via `PostHogOkHttpInterceptor`.

- Added `PostHogOkHttpInterceptorTest` coverage for:
  - exact host matches
  - hostname normalization
  - unlisted/subdomain non-matches
  - behavior with telemetry capture enabled/disabled
  - existing-header overwrite behavior
  - ended-session behavior
- Added default config coverage in `PostHogConfigTest`
- Ran:
  - `./gradlew :posthog:test --tests "com.posthog.PostHogOkHttpInterceptorTest" --tests "com.posthog.PostHogConfigTest"`
  - `./gradlew spotlessKotlinCheck -PspotlessFiles='posthog/src/main/java/com/posthog/PostHogConfig.kt,posthog/src/main/java/com/posthog/PostHogOkHttpInterceptor.kt,posthog/src/test/java/com/posthog/PostHogConfigTest.kt,posthog/src/test/java/com/posthog/PostHogOkHttpInterceptorTest.kt'`
  - `make api`
  - `./gradlew apiCheck`

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.changeset/odd-swans-wave.md
+++ b/.changeset/odd-swans-wave.md
@@ -1,5 +1,6 @@
 ---
 'posthog': minor
+'posthog-android': minor
 ---
 
-Add tracing header support for OkHttp requests.
+Add tracing header support for Android OkHttp requests.
--- a/.changeset/odd-swans-wave.md
+++ b/.changeset/odd-swans-wave.md
@@ -1,0 +1,5 @@
+---
+'posthog': minor
+---
+
+Add tracing header support for OkHttp requests.
--- a/posthog/src/main/java/com/posthog/PostHogConfig.kt
+++ b/posthog/src/main/java/com/posthog/PostHogConfig.kt
@@ -299,7 +299,11 @@ public open class PostHogConfig(
      * - The Android SDK does not send `X-POSTHOG-WINDOW-ID` because mobile apps do not have a per-window/tab concept
      * - Existing values for these headers will be overwritten when PostHog provides a value
      */
-    public var addTracingHeaders: List<String>? = null
+    public var tracingHeaders: List<String>? = null
+        get() = field?.toList()
+        set(value) {
+            field = value?.toList()
+        }
 
     /**
      * Optional custom OkHttpClient for HTTP requests.
--- a/posthog/src/main/java/com/posthog/PostHogConfig.kt
+++ b/posthog/src/main/java/com/posthog/PostHogConfig.kt
@@ -285,6 +285,9 @@ public open class PostHogConfig(
      */
     public var releaseIdentifier: String? = null,
 ) {
+    @Volatile
+    private var tracingHeadersList: List<String>? = null
+
     /**
      * When set, PostHog injects tracing headers into OkHttp requests whose destination hostname
      * exactly matches one of the configured hostnames.
@@ -298,11 +301,13 @@ public open class PostHogConfig(
      * - Hostname matching is exact and does not include ports or subdomain wildcards
      * - The Android SDK does not send `X-POSTHOG-WINDOW-ID` because mobile apps do not have a per-window/tab concept
      * - Existing values for these headers will be overwritten when PostHog provides a value
+     * - Reading this property returns a snapshot
      */
-    public var tracingHeaders: List<String>? = null
-        get() = field?.toList()
+    @get:JvmSynthetic
+    public var tracingHeaders: List<String>?
+        get() = tracingHeadersList?.let { ArrayList(it) }
         set(value) {
-            field = value?.toList()
+            tracingHeadersList = value?.let { ArrayList(it) }
         }
 
     /**
--- a/posthog/src/main/java/com/posthog/PostHogConfig.kt
+++ b/posthog/src/main/java/com/posthog/PostHogConfig.kt
@@ -286,6 +286,22 @@ public open class PostHogConfig(
     public var releaseIdentifier: String? = null,
 ) {
     /**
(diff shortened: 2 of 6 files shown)


posthog/src/main/java/com/posthog/PostHogOkHttpInterceptor.kt
```diff
@@ -6,36 +6,54 @@ import okhttp3.Interceptor
...
 
+/**
+ * Interceptor that captures network telemetry for session replay and, when
+ * [PostHogConfig.addTracingHeaders] is configured, injects PostHog tracing headers into matching
+ * OkHttp requests.
+ *
+ * Install this interceptor on each [okhttp3.OkHttpClient] whose requests should be captured or
+ * annotated with tracing headers.
+ */
 public class PostHogOkHttpInterceptor(
```
dustinbyrne, line 17: A nice improvement for later would be to auto-install this at build-time via `posthog-android-gradle-plugin`

posthog/src/main/java/com/posthog/PostHogConfig.kt
```diff
@@ -285,6 +285,22 @@ public open class PostHogConfig(
...
+     * - `X-POSTHOG-DISTINCT-ID`
+     * - `X-POSTHOG-SESSION-ID`
+     *
+     * Notes:
+     * - Requires installing [PostHogOkHttpInterceptor] on each [OkHttpClient] that should send tracing headers
+     * - Hostname matching is exact and does not include ports or subdomain wildcards
+     * - The Android SDK does not send `X-POSTHOG-WINDOW-ID` because mobile apps do not have a per-window/tab concept
+     * - Existing values for these headers will be overwritten when PostHog provides a value
+     */
+    public var addTracingHeaders: List<String>? = null
```
marandaneto:
```suggestion
    public var tracingHeaders: List<String>? = null
```

https://github.com/PostHog/posthog-android/blob/65cca7c95edb34198acc8e82dd0967a6a2a04808/posthog/src/main/java/com/posthog/PostHogConfig.kt#L376-L413
i'd do this insteda, the `add` is semantically wrong here since you can also remove
letting users to add/remove with a list instance might run into concurrentmodificationexception

.changeset/odd-swans-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog': minor
```
marandaneto, line 2: we'd need android here as well

posthog/src/main/java/com/posthog/PostHogOkHttpInterceptor.kt
marandaneto: yep, hello byte code manipulation (i dont love it, breaks super often due to code stripping, minifiers, etc, AGP version, etc)

posthog/src/main/java/com/posthog/PostHogConfig.kt
dustinbyrne: Renamed to `tracingHeaders`, and getting the list now returns a copy so we shouldn't have any concurrent modification issues 👍

posthog/src/main/java/com/posthog/PostHogOkHttpInterceptor.kt
```diff
@@ -84,3 +102,44 @@ public class PostHogOkHttpInterceptor(
...
+    return requestBuilder.build()
+}
+
+private fun normalizeTracingHeaderHostnames(hostnames: List<String>?): Set<String> {
+    return hostnames
+        ?.asSequence()
+        ?.map { it.trim().lowercase() }
+        ?.filter { it.isNotEmpty() }
+        ?.toSet()
+        ?: emptySet()
```
marandaneto: i think we should return as is here so we dont need to allocate an empty set

posthog/src/main/java/com/posthog/PostHogOkHttpInterceptor.kt
```diff
@@ -84,3 +102,44 @@ public class PostHogOkHttpInterceptor(
...
+
+    val requestBuilder = request.newBuilder()
+
+    if (!sessionId.isNullOrBlank()) {
+        requestBuilder.header(POSTHOG_SESSION_ID_HEADER, sessionId)
+    }
+
+    if (distinctId.isNotBlank()) {
+        requestBuilder.header(POSTHOG_DISTINCT_ID_HEADER, distinctId)
+    }
```
marandaneto, line 137: not sure if its worth it but if we move this above and return early in case both values are null or empty, we dont need to do the normalize dance

posthog/src/main/java/com/posthog/PostHogConfig.kt
marandaneto: i think this would throw UnsupportedOperationException in Java since Java wont know the difference between List and MutableList in Kotlin (because the List is a getter but does not hide the .add)
I dont think its a big issue since Android apps have been Kotlin first anyway
We have the same problem with `evaluationContexts`, probably missed during review
Not a blocker

marandaneto (approved):
left 2 suggestions otherwise LGTM

PR #497, opened 2026-04-27
fix(surveys): include dismissal responses on Android

lucasheriques:
- include in-progress survey responses on `survey dismissed` events
- add `$survey_partially_completed` for dismissed surveys
- store both legacy index-based response keys and `$survey_response_<questionId>` keys on Android
- align `$survey_questions` with RN/iOS so it contains `{ id, question, response }` objects instead of only question strings
- add a patch changeset for release tracking

## Companion PRs
- React Native: https://github.com/PostHog/posthog-js/pull/3468
- iOS: https://github.com/PostHog/posthog-ios/pull/580
- Android: https://github.com/PostHog/posthog-android/pull/497

- added focused dismissal event coverage in `PostHogSurveysEventPayloadTest`
- added focused `survey sent` payload coverage for legacy keys, question-id keys, and richer `$survey_questions`
- local Gradle test execution is blocked in this environment because no Android SDK is configured (`ANDROID_HOME` / `local.properties` missing)

## :pencil: Checklist

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
@@ -280,7 +280,10 @@ public class PostHogSurveysIntegration(
                 val nextQuestion = getNextQuestion(originalSurvey, questionIndex, response)
 
                 // Store the response for survey completion tracking
-                currentSurveyResponses[getResponseKey(questionIndex)] = response
+                currentSurveyResponses[getLegacyResponseKey(questionIndex)] = response
+                originalSurvey.questions.getOrNull(questionIndex)?.id?.takeIf { it.isNotEmpty() }?.let { questionId ->
+                    currentSurveyResponses[getQuestionIdResponseKey(questionId)] = response
+                }
 
                 // Check if survey is completed (needed on close event)
                 activeSurveyCompleted = nextQuestion.isSurveyCompleted
@@ -297,6 +300,7 @@ public class PostHogSurveysIntegration(
         val onSurveyClosed: OnPostHogSurveyClosed = onSurveyClosed@{ _ ->
             // Get current active survey and completion state
             val currentActiveSurvey = activeSurvey
+            val surveyResponses = currentSurveyResponses.toMap()
 
             // Validate that this survey matches the currently active survey
             if (currentActiveSurvey == null || originalSurvey.id != currentActiveSurvey.id) {
@@ -306,7 +310,7 @@ public class PostHogSurveysIntegration(
 
             // Send survey dismissed event if survey was not completed
             if (!activeSurveyCompleted) {
-                sendSurveyDismissedEvent(originalSurvey)
+                sendSurveyDismissedEvent(originalSurvey, surveyResponses)
             }
 
             // Mark survey as seen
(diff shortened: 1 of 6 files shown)


dustinbyrne (approved):

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
```diff
@@ -700,6 +696,37 @@ public class PostHogSurveysIntegration(
...
+                            ?: getLegacyResponseKey(index)
+                    responsesProperties[responseKey]?.let { put("response", it) }
+                }
+            }
+
+        return mapOf("\$survey_questions" to surveyQuestions) + responsesProperties
+    }
+
+    private fun surveyHasResponses(responses: Map<String, PostHogSurveyResponse>): Boolean {
+        return responses.values.any { it.toResponseValue() != null }
```
dustinbyrne, line 729: Given how [Rating](https://github.com/PostHog/posthog-android/blob/e5b7d79913e3c41d38f8f258a0da9340c98f08e8/posthog/src/main/java/com/posthog/surveys/PostHogSurveyResponse.kt#L33) accepts a nullable integer, and [how it's converted to a response value](https://github.com/PostHog/posthog-android/blob/main/posthog/src/main/java/com/posthog/surveys/PostHogSurveyResponse.kt#L52), it looks like it could return the string `"null"` and cause this function to return true.

Seems like a weird edge case, though, and given the way we're tracking responses, it looks unlikely? I'll leave it up to you.

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
```diff
@@ -280,7 +280,10 @@ public class PostHogSurveysIntegration(
                 val nextQuestion = getNextQuestion(originalSurvey, questionIndex, response)
 
                 // Store the response for survey completion tracking
-                currentSurveyResponses[getResponseKey(questionIndex)] = response
+                currentSurveyResponses[getLegacyResponseKey(questionIndex)] = response
```
ioannisj, line 288: This was pre-existing but I think `currentSurveyResponses` should be synchronized with activeSurveyLock

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
ioannisj: I'm not 100% sure if this will generate "null" string either, but doesn't hurt to be covered in a test?

ioannisj (approved):

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
ioannisj: Yeah looks like it will generate "null" 

We should probably change [this](https://github.com/PostHog/posthog-android/blob/e5b7d79913e3c41d38f8f258a0da9340c98f08e8/posthog/src/main/java/com/posthog/surveys/PostHogSurveyResponse.kt#L52) from `is Rating -> rating.toString()` to `is Rating -> rating?.toString()`

lucasheriques: Yep, fixed this. Rating(null) now returns null instead of the string "null", and I added both a core response test and a dismissed payload test for that case.

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
lucasheriques: Good call, fixed. I wrapped the active survey response/completion reads and writes with activeSurveyLock and snapshot the responses before sending the event.

PR #498, opened 2026-04-27
feat(server): add evaluateFlags() API for single-call flag evaluation

dmarticus:
## Problem

Phase 1 + Phase 2 of the Server SDK Feature Flag Evaluations RFC for the JVM `posthog-server` package. Companion to the Node SDK PR ([PostHog/posthog-js#3476](https://github.com/PostHog/posthog-js/pull/3476)) and the Python SDK PR ([PostHog/posthog-python#539](https://github.com/PostHog/posthog-python/pull/539)).

Today every flag check on `posthog-server` fires its own `/flags` request, and `capture(appendFeatureFlags = true)` silently fires yet another on every captured event. The flag values on a captured event can diverge from the ones the code actually branched on when person/group properties differ between calls. `appendFeatureFlags` also attaches every evaluated flag to every event, which bloats properties on high-volume events.

Tracking RFC: [requests-for-comments-internal#1020](https://github.com/PostHog/requests-for-comments-internal/pull/1020).

### New API (Phase 1)

`postHog.evaluateFlags(distinctId, …)` returns a `PostHogFeatureFlagEvaluations` snapshot:

```kotlin
val flags = postHog.evaluateFlags(distinctId, personProperties = mapOf("plan" to "enterprise"))
if (flags.isEnabled("new-dashboard")) {
    renderNewDashboard()
}
postHog.capture(distinctId, "page_viewed", flags = flags)
```

A single `/flags` request powers both branching and event enrichment. `isEnabled()` and `getFlag()` fire `$feature_flag_called` events (deduped through the existing per-distinct-id LRU) with the full metadata — `$feature_flag_id`, `$feature_flag_version`, `$feature_flag_reason`, `$feature_flag_request_id`, `$feature_flag_error`, plus `$feature_flag_definitions_loaded_at` for locally-evaluated flags — so experiment exposure tracking keeps working.

Java callers get a `PostHogEvaluateFlagsOptions.builder()` analogue alongside the Kotlin named-args entry point:

```java
PostHogFeatureFlagEvaluations flags = postHog.evaluateFlags(
    "user-123",
    PostHogEvaluateFlagsOptions.builder()
        .personProperty("plan", "enterprise")
        .flagKeys(List.of("new-dashboard"))
        .build()
);
```

### Two layers of scoping

- **Network-level (`flagKeys` option):** scopes the underlying `/flags` request itself. Goes into the request body as `flag_keys_to_evaluate`.

  ```kotlin
  val flags = postHog.evaluateFlags(distinctId, flagKeys = listOf("new-dashboard", "checkout-flow"))
  ```

- **Event-level (filter helpers):** narrow which flags get attached to a captured event without re-fetching.

  ```kotlin
  postHog.capture(distinctId, "page_viewed", flags = flags.onlyAccessed())
  postHog.capture(distinctId, "page_viewed", flags = flags.only(listOf("new-dashboard")))
  ```

Both `onlyAccessed()` and `only(...)` clone the snapshot with their own `accessed` set so filtered views don't back-propagate into the parent.

### Deprecations (Phase 2)

The legacy single-flag surface keeps working but is now `@Deprecated`:

- `isFeatureEnabled(...)`
- `getFeatureFlag(...)`
- `getFeatureFlagPayload(...)`
- `getFeatureFlagResult(...)`
- `capture(appendFeatureFlags = true)` — emits a one-line deprecation log at runtime when truthy (Kotlin can't deprecate a single parameter value at compile time)

Each `@Deprecated` annotation includes a message pointing at `evaluateFlags(...)`. Phase 3 (removal in next major) ships separately.

### New config option

`PostHogConfig.featureFlagsLogWarnings` (default `true`) — set to `false` to silence the warnings emitted by the `onlyAccessed()` (empty-access fallback) and `only(...)` (unknown-key drop) filter helpers. Useful for callers who use those helpers conditionally.

### Other request-body additions

`evaluateFlags(...)` also takes `disableGeoip = true` to forward `geoip_disable` into the `/flags` request body, parallel to the option that already exists in posthog-node and posthog-python.

### Local evaluation

Transparent. When the poller resolves a flag, the snapshot carries `locally_evaluated = true`, reason `"Evaluated locally"`, and `$feature_flag_definitions_loaded_at` is plumbed through, matching what the per-flag local path emits today.

### Backwards compatibility

**No breaking changes.** All existing call paths return the same values they did before. Kotlin callers see a `@Deprecated` compile-time warning (silenceable with `@Suppress("DEPRECATION")`); the `appendFeatureFlags = true` runtime log goes through the standard config logger.

### Internals

`PostHogStateless.sendFeatureFlagCalled` is split into `captureFeatureFlagCalledEvent`, which is shared between the per-flag accessor path and the new snapshot. Both paths now dedupe identically against the same `PostHogFeatureFlagCalledCache` LRU. The single-flag path also picks up `$feature_flag_id` / `$feature_flag_version` / `$feature_flag_reason` here — previously only the Android client emitted those, server SDK was missing them.

A small `EvaluationsHost` interface is what the snapshot calls back into, instead of holding a reference to the full client — keeps the snapshot easy to test in isolation with a fake host.

Response-level errors (`errors_while_computing_flags`, `quota_limited`) are propagated into snapshot `$feature_flag_called` events via `EvaluateFlagsResult.responseError`, matching the granularity of the per-flag path.

A new `getFeatureFlagDetails(...)` default method on `PostHogFeatureFlagsInterface` (returns `null` by default) lets the server SDK expose the cached `FeatureFlag` to the per-flag path without changing existing implementations.

Docs: `posthog-server/USAGE.md` updated with a Kotlin + Java example of the snapshot flow and the `flagKeys` vs `only(...)` distinction.

## Tests

Two test files cover snapshot semantics and end-to-end flow:

- `PostHogFeatureFlagEvaluationsTest` — snapshot accessors, full metadata on captures, filter helpers, `onlyAccessed()` empty-fallback warning, `only(...)` unknown-key warning, parent/child filter isolation, blank-distinctId no-op, locally-evaluated tagging, response-error propagation, `featureFlagsLogWarnings = false` host suppression.
- `PostHogEvaluateFlagsTest` — end-to-end via `MockWebServer`: single `/flags` round-trip, no events fire before access, dedup across access, `getFlagPayload` no-event, `capture(flags=…)` doesn't issue a second request, `flag_keys_to_evaluate` body forwarding, blank-distinctId short-circuit, local evaluation tagging, `quota_limited` propagation, `appendFeatureFlags = true` deprecated path still works end-to-end.

`./gradlew :posthog-server:check :posthog:check` (tests + lint + apiCheck + animalsniffer) clean.

---


posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
```diff
@@ -0,0 +1,177 @@
...
+        val accessedKeys =
+            synchronized(accessLock) { accessed.toSet() }
+        if (accessedKeys.isEmpty()) {
+            if (host.warningsEnabled) {
+                host.logWarning(
+                    "PostHogFeatureFlagEvaluations.onlyAccessed() called before any flag was " +
+                        "accessed; returning all $${flagMap.size} flags instead of an empty snapshot.",
+                )
+            }
+            return cloneWith(flagMap.keys)
```
dustinbyrne: This should be returning an empty array

posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
```diff
@@ -665,6 +679,113 @@ internal class PostHogFeatureFlags(
...
+        }
+
+        val localFlags =
+            getFeatureFlagsFromLocalEvaluation(
+                distinctId,
+                groups,
+                personProperties,
+                groupProperties,
+                onlyEvaluateLocally,
+            )
```
dustinbyrne, line 746: Probably out of scope for this PR, but noting that we don't support `flagKeys` in local evaluation

posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
```diff
@@ -0,0 +1,177 @@
...
+            initialAccessed = emptySet(),
+        )
+    }
+
+    private fun recordAccess(
+        key: String,
+        flag: FeatureFlag?,
+    ) {
+        synchronized(accessLock) { accessed.add(key) }
+        if (flag == null) return
```
dustinbyrne: In Python we'd continue to emit a`$feature_flag_called` for the missing flag (with `$feature_flag_error: flag_missing`)

posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
```diff
@@ -0,0 +1,177 @@
...
+        if (flag == null) return null
+        return flag.variant ?: flag.enabled
+    }
+
+    /**
+     * Returns the flag payload. Does not fire any event and does not record the access — payloads
+     * are an inert read.
+     */
+    public fun getFlagPayload(key: String): Any? {
+        return flagMap[key]?.metadata?.payload
```
dustinbyrne, line 86: `FeatureFlagMetadata.payload` is typed as a `String?`, so either the return type is wrong or this isn't returning what we expect it to return.

Perhaps we deserialize and copy what `FeatureFlagResult.getPayloadAs<T>` does

dustinbyrne: also, is the payload always guaranteed to be a string now? I thought the API response could also be a deserialized JSON value, but maybe this has changed

posthog-server/src/main/java/com/posthog/server/PostHog.kt
```diff
@@ -187,38 +245,91 @@ public class PostHog : PostHogStateless(), PostHogInterface {
...
+        snapshot: PostHogFeatureFlagEvaluations,
+    ): Map<String, Any> {
+        return appendFlagPropertiesFromMap(properties, snapshot.flags)
+    }
+
+    private fun appendFlagPropertiesFromMap(
+        properties: Map<String, Any>?,
+        flags: Map<String, FeatureFlag>?,
+    ): Map<String, Any> {
+        val props = properties?.toMutableMap() ?: mutableMapOf()
```
dustinbyrne, line 262: nit: in Python we prefer user defined properties over generated properties. In this method we override user properties

posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
dustinbyrne: Also `flagKeys` and `disableGeoip` is not considered in the cache key

posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
```diff
@@ -0,0 +1,177 @@
...
+     * clone containing every flag instead of silently dropping data.
+     */
+    public fun onlyAccessed(): PostHogFeatureFlagEvaluations {
+        val accessedKeys =
+            synchronized(accessLock) { accessed.toSet() }
+        if (accessedKeys.isEmpty()) {
+            if (host.warningsEnabled) {
+                host.logWarning(
+                    "PostHogFeatureFlagEvaluations.onlyAccessed() called before any flag was " +
+                        "accessed; returning all $${flagMap.size} flags instead of an empty snapshot.",
```
dustinbyrne:
```suggestion
                        "accessed; returning all ${flagMap.size} flags instead of an empty snapshot.",
```

posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
```diff
@@ -0,0 +1,177 @@
...
+    private val accessLock = Any()
+    private val accessed: MutableSet<String> = HashSet(initialAccessed)
+
+    /** Returns the snapshotted flag keys in iteration order of the underlying map. */
+    public val keys: List<String>
+        get() = flagMap.keys.toList()
+
+    /** Returns the immutable flag map this snapshot was built from. */
+    public val flags: Map<String, FeatureFlag>
+        get() = flagMap
```
dustinbyrne: There's no guarantee that this is immutable when called from Java

posthog/src/main/java/com/posthog/PostHogStateless.kt
```diff
@@ -442,28 +442,47 @@ public open class PostHogStateless protected constructor(
...
+     * (after applying its `sendFeatureFlagEvent` override) and the new feature-flag-evaluations
+     * snapshot (after checking its own config) funnel through here so dedup stays uniform.
+     */
+    @PostHogInternal
+    public fun captureFeatureFlagCalledEvent(
+        distinctId: String,
+        key: String,
+        value: Any?,
+        properties: Map<String, Any>,
+    ) {
```
dustinbyrne, line 476: This can be `protected` instead of `public`

posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
```diff
@@ -0,0 +1,177 @@
...
+) {
+    private val accessLock = Any()
+    private val accessed: MutableSet<String> = HashSet(initialAccessed)
+
+    /** Returns the snapshotted flag keys in iteration order of the underlying map. */
+    public val keys: List<String>
+        get() = flagMap.keys.toList()
+
+    /** Returns the immutable flag map this snapshot was built from. */
+    public val flags: Map<String, FeatureFlag>
```
dustinbyrne: Does this need to be public? It leaks `com.posthog.internal.FeatureFlag` into the public API space

posthog-server/src/main/java/com/posthog/server/PostHogConfig.kt
```diff
@@ -137,6 +137,13 @@ public open class PostHogConfig constructor(
      * Defaults to null
      */
     public var evaluationContexts: List<String>? = null,
+    /**
+     * Whether the SDK logs warnings emitted by `PostHogFeatureFlagEvaluations` filter helpers
+     * (`onlyAccessed()` with no recorded access, `only([...])` with unknown keys). Set to false
+     * to silence those messages.
+     * Defaults to true.
+     */
+    public var featureFlagsLogWarnings: Boolean = true,
```
dustinbyrne: Do we need this? The `onlyAccessed` case should be fixed, and it doesn't seem worth adding an additional configuration option just for the latter case

posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
dmarticus: Fixed in 1fabff1 — `onlyAccessed()` now returns an empty snapshot when nothing has been accessed (no warning, no fall back to all flags).

posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
dmarticus: Done in 1fabff1 — local evaluation evaluates every defined flag (no way around that without partial flag-definition support), but the result is now filtered by `flagKeys` post-hoc so callers see the same scoped set regardless of which tier resolved them. Added a comment in the code calling that out.

dmarticus: Fixed in 1fabff1 — `flagKeys` and `disableGeoip` are now part of `FeatureFlagCacheKey`, so successive `evaluateFlags` calls with different scopes don't return stale cached results. Added an integration test (`evaluateFlags caches per (distinctId, flagKeys, disableGeoip) tuple`) that verifies different `flagKeys` miss the cache.

posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
dmarticus: Fixed in 1fabff1 — `recordAccess` now still fires for `flag == null`, builds props with `$feature_flag_error: flag_missing`, and combines with the response-level error when both apply (e.g. `errors_while_computing_flags,flag_missing`). Two new tests cover this.

posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
dmarticus: Fixed in 1fabff1 — split into two methods to make the contract clear:

- `getFlagPayload(key): String?` returns the raw JSON-encoded payload string (matches `FeatureFlagMetadata.payload`).
- `getFlagPayloadAs<T>(key)` (and the `Class<T>` overload) Gson-parses the raw string, mirroring `FeatureFlagResult.getPayloadAs`.

I couldn't delegate to `FeatureFlagResult.getPayloadAs` directly because that path treats `payload` as already-deserialized — when you pass the raw JSON string `"\"hello\""` it `toJsonTree`s it as a primitive instead of parsing. The new method always Gson-parses so quoted JSON literals deserialize correctly.

dmarticus: `FeatureFlagMetadata.payload` is typed as `String?` today, and the `/flags` v2 deserializer (in `posthog/internal`) keeps it as a string. So in this PR the snapshot stores it as the raw JSON string and `getFlagPayloadAs<T>` does the Gson parse. If the wire format is ever broadened to deserialize JSON values upfront we'd revisit, but matched the current behavior here.

posthog-server/src/main/java/com/posthog/server/PostHog.kt
dmarticus: Fixed in 1fabff1 — `appendFlagPropertiesFromMap` now uses `putIfAbsent` for both `$feature/<key>` and `$active_feature_flags`, so user-supplied values win. Added an integration test (`capture preserves user-supplied feature properties over snapshot values`).

posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
dmarticus: Moot now — the warning is gone (`onlyAccessed()` returns an empty snapshot, no warning). Thanks for catching the typo.

posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
dmarticus: Fixed in 1fabff1 — `flagMap` and `locallyEvaluated` are now wrapped in `Collections.unmodifiableMap` at construction. Java callers that try to mutate get an `UnsupportedOperationException`.

posthog/src/main/java/com/posthog/PostHogStateless.kt
dmarticus: Done in 1fabff1 — changed to `protected`. The inner `EvaluationsHost` anonymous object in `PostHog` calls it via `this@PostHog.captureFeatureFlagCalledEvent(...)`, which is a subclass-internal access so the visibility constraint holds.

posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
dmarticus: Good catch — fixed in 1fabff1, the `flags` getter is now `internal`. Same-package internal callers (`PostHog.appendFlagPropertiesFromMap`) still read it; tests use the constructor's parameter directly.

posthog-server/src/main/java/com/posthog/server/PostHogConfig.kt
dmarticus: Removed in 1fabff1. With `onlyAccessed()` now returning an empty snapshot, the only remaining warning is the rare `only(...)` typo case, which doesn't justify a dedicated config.

dustinbyrne (approved):

PR #503, opened 2026-04-30
fix: no-op setup without api key

marandaneto:
Setup currently continued initializing after the API key trimmed to an empty string. Java callers could also pass a null Android API key and hit SDK init unexpectedly.

This changes null API keys to trim to an empty string and makes core/stateless setup return before initialization when the key is empty after trimming. Android setup skips Android-specific configuration for invalid keys.

- `./gradlew :posthog:test --tests "com.posthog.PostHogConfigTest" --tests "com.posthog.PostHogStatelessTest" --tests "com.posthog.PostHogTest"`
- `./gradlew :posthog-android:testDebugUnitTest --tests "com.posthog.android.PostHogAndroidConfigTest" --tests "com.posthog.android.PostHogAndroidConfigJavaTest" --tests "com.posthog.android.PostHogAndroidTest"`
- `make checkFormat`
- `./gradlew :posthog:apiCheck :posthog-android:apiCheck`

## :pencil: Checklist

### If releasing new changes

- [x] Added a changeset file for the affected packages


dustinbyrne (approved):

posthog/src/main/java/com/posthog/PostHogStateless.kt
```diff
@@ -54,7 +54,8 @@ public open class PostHogStateless protected constructor(
                     if (config.logger is PostHogNoOpLogger) PostHogPrintLogger(config) else config.logger
 
                 if (config.apiKey.isEmpty()) {
-                    config.logger.log("apiKey is empty after trimming whitespace; check your project API key")
+                    config.logger.log("PostHog SDK is disabled because the API key is required and was empty after trimming whitespace.")
```
marandaneto, line 57: i think this is wrong, working on it
reason is that the SDK might not have an api key, but might have a personal api key, and the features that depend on the personal api key should still work

marandaneto:
@dustinbyrne can you check again, i changed a few things most related to personal api key checks

dustinbyrne (approved):
👍 👍

PR #509, opened 2026-05-04
ci: caller for PostHog/.github changeset-hygiene workflow

turnipdabeets:
Updated to use merged workflow in https://github.com/PostHog/.github/pull/39

Adds a tiny caller workflow for the shared `changeset-hygiene` reusable workflow at [PostHog/.github](https://github.com/PostHog/.github/blob/main/.github/workflows/changeset-hygiene.yml). Catches the class of mistake that caused #491 → #507: a changeset frontmatter pointing at the wrong workspace package, plus a few related failure modes:

- changeset declares a package the PR didn't modify
- PR modifies a package no changeset declares (the #491 case)
- changeset declares a name not in the workspace (typo)
- PR modifies workspace package code but adds no changeset at all

This PR carries no logic of its own — the workflow + script live in PostHog/.github so all SDKs share them. This file just delegates.

## What it does

- Triggers on `pull_request` against `main`.
- Delegates to the reusable workflow via `uses: PostHog/.github/.github/workflows/changeset-hygiene.yml@main`.
- Posts a single sticky comment under `<!-- changeset-hygiene -->` when there's a coverage issue. Auto-deletes when fixed. **Never blocks merging.**

## Caller is 18 lines

The whole change is one file: `.github/workflows/changeset-hygiene.yml`. The actual workflow + script are at `PostHog/.github/.github/workflows/changeset-hygiene.yml` and `PostHog/.github/.github/scripts/check-changeset-coverage.mjs`.

## Test plan

Verified end-to-end on the parallel test PR #510, which uses the same caller plus deliberate simulation files (source change in `posthog-android-gradle-plugin/` paired with a changeset declaring `posthog-android`). The workflow correctly posted the expected mismatch comment.

This PR's own CI run (no workspace package changes, no changeset) correctly stays silent — the workflow only comments when there's an issue.

## :pencil: Checklist

### If releasing new changes

(CI-only change — no workspace package code is touched, so no changeset is needed.)


marandaneto:
does it also cover the use case where maybe we have modinified packages but no changeset at all? i think this would be great as well

also take a look at `posthog-js/.github/workflows/check-posthog-major-version.yml` (Comment about missing changeset) since theres something in there as well
we probably want to cover any other use case where we failed before and then port across all sdks
if possible, this would live in https://github.com/PostHog/.github/tree/main/.github/workflows and then we just reuse across repos

.github/workflows/changeset-coverage.yml
```diff
@@ -0,0 +1,49 @@
+name: 'Changeset coverage'
+
+on:
+    pull_request:
+        branches: [main]
```
marandaneto: probably bettero only for types: [opened, synchronize, reopened]

.github/workflows/changeset-coverage.yml
```diff
@@ -0,0 +1,49 @@
...
+            - name: Post sticky PR comment
+              if: steps.coverage.outputs.body != ''
+              uses: marocchino/sticky-pull-request-comment@0ea0beb66eb9baf113663a64ec522f60e49231c0 # v3.0.4
+              with:
+                  header: changeset-coverage
+                  message: ${{ steps.coverage.outputs.body }}
+
+            - name: Delete stale sticky PR comment
+              if: steps.coverage.outputs.body == ''
+              uses: marocchino/sticky-pull-request-comment@0ea0beb66eb9baf113663a64ec522f60e49231c0 # v3.0.4
```
marandaneto: actions/github-script is an alternative to not use a 3rd party action but its much simpler this way and this action looks well maintained so all good =)

.github/workflows/changeset-coverage.yml
turnipdabeets: added types

turnipdabeets:
> does it also cover the use case where maybe we have modinified packages but no changeset at all? i think this would be great as well
> 
> also take a look at `posthog-js/.github/workflows/check-posthog-major-version.yml` (Comment about missing changeset) since theres something in there as well we probably want to cover any other use case where we failed before and then port across all sdks if possible, this would live in [PostHog/.github@`main`/.github/workflows](https://github.com/PostHog/.github/tree/main/.github/workflows?rgh-link-date=2026-05-04T15%3A34%3A47Z) and then we just reuse across repos

@marandaneto  I moved this to https://github.com/PostHog/.github/pull/39. Considered joining with `check-posthog-major-version.yml` but they seem to be a bit different and js want to block on major versions. I don't think we want that in other repos so that can stay in js repo and we can share this across other repos.

marandaneto (approved):

marandaneto:
Nice!!!

PR #512, opened 2026-05-05
fix(queue): 413 no config mutation

turnipdabeets:
The 413 adaptive batch sizing currently mutates the user-provided PostHogConfig, halving config.maxBatchSize and config.flushAt directly. That makes the user's configured values silently change at runtime — surprising for callers that read those fields back, and it leaks internal queue state across SDK boundaries.

It also halves from the *configured* cap rather than the *actual* batch size that triggered the 413. If a 413 fires on a batch of 10 with maxBatchSize=50, we halve from 50 → 25 → 12 → ... wasting halvings on sizes the server already proved (or is about to prove) too large.

Move the cap and flushAt into a private BatchLimits class owned by PostHogQueue, initialized from config once at construction. Halve from min(cap, actualBatchSize) so a partial batch's 413 doesn't waste cycles on larger caps the server hasn't accepted, and clamp flushAt to cap on each halve so we never buffer more events than a single batch can drain (e.g. cap=1 + flushAt=10 would otherwise pile 10 events to send 1 at a time). config.maxBatchSize and config.flushAt are now read-only as far as the queue is concerned.

Matches the equivalent iOS fix in posthog-ios PR #585

## :pencil: Checklist

### If releasing new changes

- [X] Ran `pnpm changeset` to generate a changeset file

### DIFF
--- a/posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
@@ -12,7 +12,6 @@ import java.util.TimerTask
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.atomic.AtomicBoolean
 import kotlin.concurrent.schedule
-import kotlin.math.max
 import kotlin.math.min
 import kotlin.math.pow
 
@@ -36,7 +35,7 @@ internal class PostHogQueue(
     private val timerLock = Any()
     private var pausedUntil: Date? = null
     private var retryCount = 0
-    private val batchLimits = BatchLimits.initial(config)
+    private val batchLimits = initialBatchLimits(config)
     private val initialRetryDelaySeconds = 1
     private val maxRetryDelaySeconds = 30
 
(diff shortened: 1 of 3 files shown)


posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
```diff
@@ -463,35 +464,59 @@ internal class PostHogQueue(
...
+    var cap: Int,
+    var flushAt: Int,
+) {
+    companion object {
+        fun initial(config: PostHogConfig): BatchLimits =
+            BatchLimits(
+                cap = max(config.maxBatchSize, 1),
+                flushAt = max(config.flushAt, 1),
+            )
+    }
```
marandaneto:
```
internal fun initialBatchLimits(config: PostHogConfig) = BatchLimits(
    cap = config.maxBatchSize.coerceAtLeast(1),
    flushAt = config.flushAt.coerceAtLeast(1),
)
```

would be more idiomatic i think

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
```diff
@@ -463,35 +464,59 @@ internal class PostHogQueue(
...
+
+    fun halve(actualBatchSize: Int) {
+        cap = max(min(cap, actualBatchSize) / 2, 1)
+        // keep flushAt <= cap so we don't pile up events larger than a single batch
+        flushAt = max(min(flushAt / 2, cap), 1)
+    }
```
marandaneto, line 491:
```
fun halve(actualBatchSize: Int) {
    cap = minOf(cap, actualBatchSize)
        .div(2)
        .coerceAtLeast(1)

    // keep flushAt <= cap so we don't pile up events larger than a single batch
    flushAt = (flushAt / 2)
        .coerceAtMost(cap)
        .coerceAtLeast(1)
}
```

prob also better, more readable instead of nested max/min etc

ioannisj: Are there any lint rules we can use for this?

ioannisj (approved):

PR #513, opened 2026-05-14
feat(surveys): add survey translations support

lucasheriques:
## Summary
Brings the posthog-js / posthog-react-native survey translations feature to the native SDK. Surveys can carry per-language overrides for user-visible strings via a `translations` map keyed by language code. At display time the SDK resolves a language (init override → person property `language` → device locale), applies any matching translation onto the display model, and stamps the matched key as `$survey_language` on every survey event when a translation actually took effect.

- Wire model: `translations` on `Survey` and on each `SurveyQuestion` subtype.
- Config: `PostHogSurveysConfig.overrideDisplayLanguage` (init option).
- Matching: case-insensitive exact match with base-language fallback (`pt-BR` → `pt`); returns the original-cased dict key for event reporting.
- Resolution: only stamps `matchedKey` when at least one user-visible field actually changed.
- Event threading: `survey shown`, `survey sent`, `survey dismissed` all carry `$survey_language` when matched; `$survey_questions` reflects translated text.

Companion PR: PostHog/posthog-ios#601

## Test plan
All tests pass in CI. New coverage:
- Wire-format decoding round-trip (Gson + serializer).
- Matching algorithm: exact, case-insensitive, base-language fallback, no-match, no-fallback-without-hyphen.
- Priority chain: override → person prop → device locale.
- Display model translation with field-level fallback to original.
- Integration: `$survey_language` present on all three events when matched, absent otherwise; translated text in `$survey_questions`.

---

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
@@ -243,8 +243,6 @@ public class PostHogSurveysIntegration(
             return
         }
 
-        // Resolve the survey display language and apply translations once. The resulting
-        // matchedKey is stamped onto all three survey events as $survey_language.
         val displayLanguage = resolveDisplayLanguage()
         val translations = resolveSurveyTranslations(survey, displayLanguage)
         val resolvedLanguage = translations.matchedKey
@@ -598,12 +596,6 @@ public class PostHogSurveysIntegration(
     // Lifecycle management
     private var isStarted: Boolean = false
 
-    /**
-     * Resolves the language to use for the next survey display. Priority chain:
-     *   1. `PostHogSurveysConfig.overrideDisplayLanguage`
-     *   2. The `"language"` person property persisted by identify()
-     *   3. The device locale (BCP-47 tag, e.g. "en-US")
-     */
     private fun resolveDisplayLanguage(): String? {
         val override = config.surveysConfig.overrideDisplayLanguage
         val personProperties = readPersistedPersonProperties(config)
--- a/posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
@@ -11,8 +11,11 @@ import com.posthog.internal.formatISO8601Date
 import com.posthog.internal.parseISO8601Date
 import com.posthog.internal.surveys.PostHogSurveysHandler
 import com.posthog.internal.surveys.canActivateRepeatedly
+import com.posthog.internal.surveys.detectSurveyLanguage
 import com.posthog.internal.surveys.hasEvents
 import com.posthog.internal.surveys.hasWaitPeriodPassed
+import com.posthog.internal.surveys.readPersistedPersonProperties
+import com.posthog.internal.surveys.resolveSurveyTranslations
 import com.posthog.surveys.OnPostHogSurveyClosed
 import com.posthog.surveys.OnPostHogSurveyResponse
 import com.posthog.surveys.OnPostHogSurveyShown
@@ -29,6 +32,7 @@ import com.posthog.surveys.SurveyQuestion
 import com.posthog.surveys.SurveyQuestionBranching
 import com.posthog.surveys.SurveyType
 import java.util.Date
+import java.util.Locale
 
 public class PostHogSurveysIntegration(
     context: Context,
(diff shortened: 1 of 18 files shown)


posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
```diff
@@ -572,10 +590,24 @@ public class PostHogSurveysIntegration(
...
 
     // Lifecycle management
     private var isStarted: Boolean = false
 
+    private fun resolveDisplayLanguage(): String? {
+        val override = config.surveysConfig.overrideDisplayLanguage
+        val personProperties = readPersistedPersonProperties(config)
+        val deviceLocale =
+            try {
+                Locale.getDefault().toLanguageTag()
```
turnipdabeets: if neither Locale.getDefault() nor toLanguageTag() throws, could we drop the try/catch?

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
```diff
@@ -572,10 +590,24 @@ public class PostHogSurveysIntegration(
 
     // Active survey tracking
     private var activeSurvey: Survey? = null
+    private var activeSurveyLanguage: String? = null
+    private var activeSurveyQuestionTranslations: List<com.posthog.surveys.SurveyQuestionTranslation?>? = null
```
turnipdabeets: could we add import com.posthog.surveys.SurveyQuestionTranslation and then drop the full name here?

posthog/src/main/java/com/posthog/internal/surveys/SurveyLanguageDetection.kt
```diff
@@ -0,0 +1,76 @@
...
+ * Reads the persisted person properties dictionary written by `identify()` and
+ * `setPersonPropertiesForFlags()`. Returns an empty map when nothing is persisted.
+ *
+ * NOTE: the storage key here intentionally duplicates the value of the (internal)
+ * `PostHogPreferences.PERSON_PROPERTIES_FOR_FLAGS` constant. Keeping it local avoids
+ * widening the SDK's internal preference key API surface.
+ */
+@PostHogInternal
+public fun readPersistedPersonProperties(config: PostHogConfig): Map<String, Any?> {
+    val raw = config.cachePreferences?.getValue(PERSON_PROPERTIES_FOR_FLAGS_KEY) ?: return emptyMap()
```
turnipdabeets: I think PostHogRemoteConfig keeps personPropertiesForFlags as an in-memory map, like iOS

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
lucasheriques: yep good catch — dropped in 1b82274. neither call throws so the defensive guard was just noise

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
lucasheriques: done in 1b82274 — added the import and switched all sites to the short name

posthog/src/main/java/com/posthog/internal/surveys/SurveyLanguageDetection.kt
lucasheriques: yep good call — done in 1b82274. exposed `PostHogRemoteConfig.getPersonPropertiesForFlags()` as `@PostHogInternal public` and the integration reads from it directly now. dropped the helper that was reading from cachePreferences

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
```diff
@@ -572,10 +590,19 @@ public class PostHogSurveysIntegration(
 
     // Active survey tracking
     private var activeSurvey: Survey? = null
+    private var activeSurveyLanguage: String? = null
+    private var activeSurveyQuestionTranslations: List<SurveyQuestionTranslation?>? = null
```
turnipdabeets: are these two fields ever read, it looks like just written but not read so maybe they are dead code?

posthog/src/main/java/com/posthog/surveys/PostHogDisplaySurveyQuestion.kt
```diff
@@ -19,15 +19,14 @@ public open class PostHogDisplaySurveyQuestion(
     public val buttonText: String?,
 ) {
     internal companion object {
-        /**
-         * Creates a display question from a survey question
-         *
-         * @param question The survey question to convert
-         * @return A display question or null if the question type is not supported
-         */
```
ioannisj, line 27: we should maybe keep+update these doc comments

ioannisj (approved):
lgtm, just a very small comment

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
lucasheriques: great catch — those were just leftover from mirroring the iOS state-storage pattern. on android the callback closures capture `resolvedLanguage` and `resolvedQuestionTranslations` directly so the instance fields were never read. dropped in e2e1a56

posthog/src/main/java/com/posthog/surveys/PostHogDisplaySurveyQuestion.kt
lucasheriques: restored + updated in e2e1a56 to cover the new translation params on `fromSurveyQuestion`, `fromSurveyAppearance`, and `toDisplaySurvey`

PR #519, opened 2026-05-19
refactor(queue): generic PostHogQueue<Record> + EndpointSpec

turnipdabeets:

Preparing the SDK to host additional queue instances without duplicating the
disk/retry/timer plumbing that events and session replay already share. Today
`PostHogQueue` is hardcoded to `PostHogEvent` and branches on a
`PostHogApiEndpoint` enum (BATCH / SNAPSHOT) inside `batchEvents()`. A new
endpoint with a different record type, wire format, retry set, or rate cap
would otherwise become scattered `if (endpoint == ...)` branches throughout
the queue.

This PR lifts those variations into a single per-endpoint spec object so the
queue itself stays record-type-agnostic. Matches iOS's `PostHogQueue<Record>`
+ `QueueEndpoint<Record>` shape.

- All 31 existing `PostHogQueueTest` cases pass **unchanged** (byte-identical test source) — strongest behavior-preservation signal.
- `make test` green across `posthog`, `posthog-android`, `posthog-server`.
- `make checkFormat` green.
- New `EndpointSpecTest` with 6 cases proving the genericization works for a non-event synthetic record type (not just events).
- `make api` confirms the public surface delta is exactly: `EndpointSpec` + factories; `PostHogQueueInterface` becomes `<Record>`-generic; `PostHogQueue` constructor takes a spec instead of the 5 legacy args. All touched symbols are `@PostHogInternal`.
- Live end-to-end test on Android emulator with the sample app — replay snapshots queue, flush, and POST to `/s/` with `200 OK`. Verified the `recordUuid` spec lambda preserves the existing "filename matches `event.uuid`" behavior (on-disk filename `<uuid>.event` matches the wire payload `"uuid":"<uuid>"`).

## :pencil: Checklist

### If releasing new changes

- [x] Ran \`pnpm changeset\` to generate a changeset file

### DIFF
--- a/posthog/src/main/java/com/posthog/PostHogConfig.kt
+++ b/posthog/src/main/java/com/posthog/PostHogConfig.kt
@@ -1,6 +1,7 @@
 package com.posthog
 
 import com.posthog.errortracking.PostHogErrorTrackingConfig
+import com.posthog.internal.EndpointSpec
 import com.posthog.internal.PostHogApi
 import com.posthog.internal.PostHogApiEndpoint
 import com.posthog.internal.PostHogContext
@@ -258,8 +259,21 @@ public open class PostHogConfig(
     /**
      * Factory to instantiate a custom queue implementation.
      */
-    public val queueProvider: (PostHogConfig, PostHogApi, PostHogApiEndpoint, String?, ExecutorService) -> PostHogQueueInterface =
-        { config, api, endpoint, storagePrefix, executor -> PostHogQueue(config, api, endpoint, storagePrefix, executor) },
+    public val queueProvider: (
+        PostHogConfig,
+        PostHogApi,
+        PostHogApiEndpoint,
+        String?,
+        ExecutorService,
+    ) -> PostHogQueueInterface<PostHogEvent> =
+        { config, api, endpoint, storagePrefix, executor ->
+            val spec =
+                when (endpoint) {
+                    PostHogApiEndpoint.BATCH -> EndpointSpec.batch(config, api, storagePrefix)
+                    PostHogApiEndpoint.SNAPSHOT -> EndpointSpec.snapshot(config, api, storagePrefix)
+                }
+            PostHogQueue(config, spec, executor)
+        },
     /**
      * Configuration for PostHog Error Tracking feature.
      */
--- a/posthog/src/main/java/com/posthog/PostHogConfig.kt
+++ b/posthog/src/main/java/com/posthog/PostHogConfig.kt
@@ -259,6 +259,7 @@ public open class PostHogConfig(
     /**
      * Factory to instantiate a custom queue implementation.
      */
+    @PostHogInternal
     public val queueProvider: (
         PostHogConfig,
         PostHogApi,
--- a/posthog/src/main/java/com/posthog/internal/EndpointSpec.kt
+++ b/posthog/src/main/java/com/posthog/internal/EndpointSpec.kt
@@ -20,10 +20,10 @@ import java.util.UUID
 public class EndpointSpec<Record> internal constructor(
     internal val name: String,
     internal val storagePrefix: String?,
-    internal val initialCap: Int,
-    internal val initialFlushAt: Int,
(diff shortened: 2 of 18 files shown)


posthog/src/main/java/com/posthog/PostHogConfig.kt
```diff
@@ -258,8 +259,21 @@ public open class PostHogConfig(
...
      */
-    public val queueProvider: (PostHogConfig, PostHogApi, PostHogApiEndpoint, String?, ExecutorService) -> PostHogQueueInterface =
-        { config, api, endpoint, storagePrefix, executor -> PostHogQueue(config, api, endpoint, storagePrefix, executor) },
+    public val queueProvider: (
+        PostHogConfig,
+        PostHogApi,
+        PostHogApiEndpoint,
+        String?,
+        ExecutorService,
+    ) -> PostHogQueueInterface<PostHogEvent> =
```
dustinbyrne, line 269: this is source incompatible for anyone implementing a custom `queueProvider`

also an existing leak of a `@PostHogInternal` construct

it looks like we made a similar change in the last PR too, so maybe not the worst but not ideal. If this is truly public API, we should consider adding a stable compatibility layer. It seems like we'll have to break this at least one more time, but maybe we can avoid doing it a second time?

posthog/src/main/java/com/posthog/internal/EndpointSpec.kt
```diff
@@ -0,0 +1,92 @@
...
+            api: PostHogApi,
+            storagePrefix: String?,
+        ): EndpointSpec<PostHogEvent> =
+            EndpointSpec(
+                name = "events",
+                storagePrefix = storagePrefix,
+                initialCap = config.maxBatchSize,
+                initialFlushAt = config.flushAt,
+                maxQueueSize = config.maxQueueSize,
+                flushIntervalSeconds = config.flushIntervalSeconds,
```
dustinbyrne: some of these properties are now copied at construction time, whereas before they'd be read from config directly (and, as a result, could be modified at runtime)

just noting this as a behavioral change, no action needed

posthog/src/main/java/com/posthog/internal/PostHogQueue.kt
```diff
@@ -96,57 +90,57 @@ public class PostHogQueue(
...
             }
         }
     }
 
     /**
-     * Ensures cached events from disk are loaded into the deque exactly once.
+     * Ensures cached records from disk are loaded into the deque exactly once.
      * Must be called on the executor thread (single-threaded executor, no lock needed).
      */
     private fun ensureCachedEventsLoaded() {
```
dustinbyrne: should we rename these too?

`ensureCachedEventsLoaded`
`cachedEventsLoaded`
`loadCachedEvents`

dustinbyrne (approved):
great work on this! I'd like to maybe see `queueProvider` stabilize, but I'll leave it up to you as to how you'd like to handle that.

posthog/src/main/java/com/posthog/internal/EndpointSpec.kt
turnipdabeets: I'll update these to use lambdas

posthog/src/main/java/com/posthog/PostHogConfig.kt
turnipdabeets: Is there a case we know a custom `queueProvider` is used / do we want to support that? I don't think RN or Flutter use this. Doesn't look like we support this on iOS. I made queueProvider as @PostHogInternal since I don't think people should be implementing a custom queueProvider. Let me know if I'm wrong about this. ba88cb937dd8463cdd73c912db6e860fe2712fb4 @dustinbyrne

posthog/src/main/java/com/posthog/internal/EndpointSpec.kt
turnipdabeets: updated!

PR #520, opened 2026-05-19
chore(logs): internal plumbing for upcoming logs feature

turnipdabeets:

Sets up the internal plumbing for the upcoming logs feature so that the
public-facing PR (next) is purely additive on top of a proven wire path.

Adds:

- `PostHogLogRecord` and `PostHogLogSeverity` (both `internal`)
- OTLP/JSON payload builder (`PostHogLogsOTLP`)
- `PostHogApi.sendLogs(...)` — POSTs to `/i/v1/logs?token=<apiKey>`
- `EndpointSpec.logs(...)` factory — wires a third `PostHogQueue` into
  `PostHog.setup` alongside events and replay

**No public `log()` API in this PR** — that lands in the follow-up,
along with `PostHogLogsConfig` (service.name / environment / beforeSend
hook), capture-time context snapshot (distinctId / sessionId / screen
name), and screen-name tracking via `Application.ActivityLifecycleCallbacks`.
That follow-up will carry the changeset and the version bump; this PR
ships zero runtime behavior change.

The logs queue is constructed but **not started** in `setup()` — nothing
can enqueue records yet, so spinning up the periodic flush timer would
waste a daemon thread. A one-line `// logsQueue.start()` marker sits in
`PostHog.setup` where it'll be uncommented when the public capture API
ships.

Worth noting for review:

- Retry policy for `/i/v1/logs` is narrower than `/batch` /
  `/snapshot`: 408, 429, and 5xx only. 3xx redirects aren't retried
  (would loop on a misconfigured host). Regression-guarded so events
  don't accidentally inherit the broader set.
- Wire format follows OTLP `LogsService` JSON
  (`resourceLogs[].scopeLogs[].logRecords[]`). Multi-record batches
  share a single resource + scope envelope, locked in by test.
- ABI delta: one synthetic constructor parameter on `private PostHog(...)`
  (unreachable from outside the class) plus two `@PostHogInternal`
  accessors for `logsStoragePrefix`. No public surface change.

- `./gradlew :posthog:test` — OTLP shape, storage round-trip (both
  direct and through `config.serializer`), retry policy, multi-record
  batching, `sendLogs` 408 / 500, per-key collision rules,
  parameterized severity coverage
- `./gradlew :posthog-android:testDebugUnitTest`
- `./gradlew :posthog:apiCheck` — confirms ABI is additive
- `./gradlew spotlessCheck detekt`

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/posthog/src/main/java/com/posthog/internal/logs/PostHogLogsOTLP.kt
+++ b/posthog/src/main/java/com/posthog/internal/logs/PostHogLogsOTLP.kt
@@ -1,6 +1,8 @@
 package com.posthog.internal.logs
 
+import com.posthog.internal.formatISO8601Date
 import com.posthog.logs.PostHogLogRecord
+import java.util.Date
 
 /**
  * OpenTelemetry / OTLP-JSON serialization for log records.
@@ -42,6 +44,7 @@ internal object PostHogLogsOTLP {
                 }
                 mapOf("kvlistValue" to mapOf("values" to toKeyValueList(safe)))
             }
+            is Date -> mapOf("stringValue" to formatISO8601Date(value))
             // Last resort: stringify so the user gets something rather than a
             // silently dropped attribute.
             else -> mapOf("stringValue" to value.toString())
@@ -103,10 +106,6 @@ internal object PostHogLogsOTLP {
      * Layers SDK-managed resource attributes (`telemetry.sdk.name`,
      * `telemetry.sdk.version`) on top of the user-supplied
      * `resourceAttributes` so the SDK identification can't be shadowed.
-     *
-     * Other resource attributes (`service.name`, `os.*`,
-     * `deployment.environment`) are populated by the caller through
-     * `resourceAttributes` once a logs config / context bridge exists.
      */
     fun buildResourceAttributes(
         userResourceAttributes: Map<String, Any>,
--- a/posthog/src/main/java/com/posthog/internal/logs/PostHogLogsOTLP.kt
+++ b/posthog/src/main/java/com/posthog/internal/logs/PostHogLogsOTLP.kt
@@ -1,0 +1,157 @@
+package com.posthog.internal.logs
+
+import com.posthog.logs.PostHogLogRecord
+
+/**
+ * OpenTelemetry / OTLP-JSON serialization for log records.
+ *
+ * Emits the OTLP `LogsService` request shape
+ * (`resourceLogs[].scopeLogs[].logRecords[]`) — distinct from PostHog's
+ * internal events batch shape because the logs ingestion endpoint expects OTLP.
+ */
+internal object PostHogLogsOTLP {
+    /**
+     * Wraps a value as an OTLP `AnyValue`. Returns `null` for values that
+     * have no representable OTLP form (which the caller should drop).
+     *
+     * `Int` / `Long` / `Short` / `Byte` map to `intValue` (encoded as a
+     * string per proto3 JSON int64 rules). `Double` / `Float` map to
+     * `doubleValue` for finite numbers and `stringValue` ("NaN" | "Infinity"
+     * | "-Infinity") otherwise — JSON cannot represent those floats directly.
+     */
+    fun toAnyValue(value: Any?): Map<String, Any>? {
+        if (value == null) return null
+        return when (value) {
+            is String -> mapOf("stringValue" to value)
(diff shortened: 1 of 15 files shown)


posthog/src/main/java/com/posthog/logs/PostHogLogRecord.kt
```diff
@@ -0,0 +1,79 @@
...
+                sessionId = map["sessionId"] as? String,
+                screenName = map["screenName"] as? String,
+                featureFlagKeys = featureFlagKeys,
+                appState = map["appState"] as? String,
+                timeUnixNano = timeUnixNano,
+                observedTimeUnixNano = observedTimeUnixNano,
+            )
+        }
+
+        internal fun nanosNow(): String = (System.currentTimeMillis() * 1_000_000L).toString()
```
ioannisj: Should we run through PostHogDateProvider?

posthog/src/main/java/com/posthog/internal/logs/PostHogLogsOTLP.kt
```diff
@@ -0,0 +1,157 @@
...
+    fun buildResourceAttributes(
+        userResourceAttributes: Map<String, Any>,
+        sdkName: String,
+        sdkVersion: String,
+    ): Map<String, Any> {
+        val attrs = LinkedHashMap<String, Any>()
+        attrs.putAll(userResourceAttributes)
+        attrs["telemetry.sdk.name"] = sdkName
+        attrs["telemetry.sdk.version"] = sdkVersion
+        return attrs
```
ioannisj, line 119: Looks like we add a few more props on [iOS](https://github.com/PostHog/posthog-ios/blob/cf6155963a6532c7b88e0cfa2b1d07595d6b6db6/PostHog/Logs/PostHogLogsOTLP.swift#L138) side?

posthog/src/main/java/com/posthog/internal/logs/PostHogLogsOTLP.kt
```diff
@@ -0,0 +1,157 @@
...
+            is Map<*, *> -> {
+                val safe = LinkedHashMap<String, Any>()
+                for ((k, v) in value) {
+                    if (k is String && v != null) safe[k] = v
+                }
+                mapOf("kvlistValue" to mapOf("values" to toKeyValueList(safe)))
+            }
+            // Last resort: stringify so the user gets something rather than a
+            // silently dropped attribute.
+            else -> mapOf("stringValue" to value.toString())
```
ioannisj, line 50: Dates will fall through here. Should we have a case for converting to ISO8601 string?

posthog/src/main/java/com/posthog/logs/PostHogLogRecord.kt
turnipdabeets: Good call. Pushed in 9acef29.

posthog/src/main/java/com/posthog/internal/logs/PostHogLogsOTLP.kt
turnipdabeets: Was planning to add these in the next PR but I can add os.name / os.version from PostHogContext here. service.* and deployment.environment need to wait for PostHogLogsConfig in next PR. Pushed in 9acef29.

posthog/src/main/java/com/posthog/internal/logs/PostHogLogsOTLP.kt
turnipdabeets: Added! Pushed in 9acef29.

posthog/src/main/java/com/posthog/logs/PostHogLogRecord.kt
```diff
@@ -0,0 +1,89 @@
...
+            )
+        }
+
+        /**
+         * Default timestamp source. Production capture should call this with
+         * `config.dateProvider` so tests using `FakePostHogDateProvider` can
+         * drive log timestamps deterministically; the no-arg default falls
+         * back to the device clock for callers without an SDK instance.
+         */
+        internal fun nanosNow(dateProvider: PostHogDateProvider = PostHogDeviceDateProvider()): String =
```
ioannisj: Shouldn't we use config.dateProvider instead of new instance on each call?

ioannisj (approved):
LG, left one last comment but I'll leave that up to you.

PR #524, opened 2026-05-21
fix: stop screenshot frames being dropped during heavy animations

ioannisj:

Ticket :https://posthoghelp.zendesk.com/agent/tickets/55153 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/59818)

### Problem

In screenshot mode, session recordings showed blank frames whenever a continuously animating view (e.g. a Lottie animation) was on screen. 

The `isOnDrawnCalled` flag was reset just before initiating PixelCopy, then checked again in the async callback. If the flag had been set again by the time the callback fired, the screenshot was discarded. Because a 60fps animation fires onDraw every ~16ms and PixelCopy takes longer than that, the flag was always true on callback causing every frame to be silently dropped.

The original intent of `isOnDrawnCalled` guard was to protect against PII leaking through mask drift

### Fix

Collect maskable widget positions on the main thread at the moment of the draw callback, **before** submitting to the background executor, then pass the pre-collected list into PixelCopy. The callback simply applies the masks and encodes. For some reason, timing seems to be better with this new order of execution. Validated that masking does not seem to drift on fast scrolling ([here](https://us.posthog.com/shared/0oCasw2RpoZ-aVQ5hjQkhf8O97p52g?t=2465)) and that views with Lottie animations that would drop frame render okay now ([here](https://us.posthog.com/shared/ErRbDYvclFGyF8cfwzigcYm_G9GvgQ?t=51))

The isOnDrawnCalled flag and all its reset/check sites are removed. 

### Changes

- PostHogReplayIntegration: collect masks in the throttled draw callback (main thread) before handing off to the executor
- NextDrawListener: remove the separate onDrawCallback parameter — it only existed to set the flag

Manual testing with a provided MRE

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -190,13 +190,6 @@ public class PostHogReplayIntegration(
             }
         }
 
-    @Volatile
-    private var isOnDrawnCalled: Boolean = false
-
-    private fun onDrawCallback() {
-        isOnDrawnCalled = true
-    }
-
     private fun addView(
         view: View,
         added: Boolean = true,
@@ -219,15 +212,24 @@ public class PostHogReplayIntegration(
                                         mainHandler,
                                         config.dateProvider,
                                         config.sessionReplayConfig.throttleDelayMs,
-                                        ::onDrawCallback,
                                     ) {
                                         if (!isActive() || !isNativeSdk) {
                                             return@onNextDraw
                                         }
 
+                                        // Collect masks on the main thread so they're in sync with PixelCopy pixel data.
+                                        val maskableWidgets =
+                                            if (config.sessionReplayConfig.screenshot && decorView.isAliveAndAttachedToWindow()) {
+                                                val masks = mutableListOf<Rect>()
+                                                findMaskableWidgets(decorView, masks)
+                                                masks
+                                            } else {
+                                                emptyList()
+                                            }
+
                                         executor.submit {
                                             try {
-                                                generateSnapshot(WeakReference(decorView), WeakReference(window))
+                                                generateSnapshot(WeakReference(decorView), WeakReference(window), maskableWidgets)
                                             } catch (e: Throwable) {
                                                 config.logger.log("Session Replay generateSnapshot failed: $e.")
                                             }
@@ -448,7 +450,6 @@ public class PostHogReplayIntegration(
             }
 
             isSessionReplayActive = false
-            isOnDrawnCalled = false
 
             pixelCopyThread?.quitSafely()
             pixelCopyThread = null
@@ -477,6 +478,7 @@ public class PostHogReplayIntegration(
     private fun generateSnapshot(
         viewRef: WeakReference<View>,
         windowRef: WeakReference<Window>,
+        maskableWidgets: List<Rect> = emptyList(),
     ) {
         // Early bail if stopped and this is processing previous generateSnapshot() from executor.submit
         if (!isActive()) return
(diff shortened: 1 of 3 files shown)


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -219,15 +212,24 @@ public class PostHogReplayIntegration(
...
                                     ) {
                                         if (!isActive() || !isNativeSdk) {
                                             return@onNextDraw
                                         }
 
+                                        // Collect masks on the main thread so they're in sync with PixelCopy pixel data.
+                                        val maskableWidgets =
+                                            if (config.sessionReplayConfig.screenshot && decorView.isAliveAndAttachedToWindow()) {
+                                                val masks = mutableListOf<Rect>()
+                                                findMaskableWidgets(decorView, masks)
```
turnipdabeets, line 224: Quick threading question — Can we inline the semantics walk now that we're already on main?

findMaskableWidgets now runs on the main thread, but findMaskableComposeWidgets still posts to the main handler and awaits a latch which I think might deadlock until the 1s timeout and return empty masks. On a screen with a ComposeView, I'd expect a ~1s UI freeze per snapshot plus dropped Compose masks (PII the old isOnDrawnCalled guard used to catch). Did you test this case?

ioannisj: Ah, right I think you are completely right. Did not test that path

ioannisj:
@turnipdabeets if you get some time, mind testing for masking drift as well in recordings? I'm not super confident that we are not bringing back an old with these changes 🙏 In the meantime I'll address the Compose issue you mentioned above

marandaneto:
im not sure removing `isOnDrawnCalled` is the right approach here, it was added for increasing performance, so screenshots midflight would be dropped and masking would be more accurate, if this is now gone, we are back to the other issue
using animations or screens that are drawing 24/7 is an edge case so i guess a better option would be to offer a sdk option to disable the isOnDrawnCalled behaviour for those people?

marandaneto:
> Collect maskable widget positions on the main thread at the moment of the draw callback, before submitting to the background executor, then pass the pre-collected list into PixelCopy. The callback simply applies the masks and encodes. For some reason, timing seems to be better with this new order of execution. Validated that masking does not seem to drift on fast scrolling ([here](https://us.posthog.com/shared/0oCasw2RpoZ-aVQ5hjQkhf8O97p52g?t=2465)) and that views with Lottie animations that would drop frame render okay now ([here](https://us.posthog.com/shared/ErRbDYvclFGyF8cfwzigcYm_G9GvgQ?t=51))

ok just read it, down to try but i'm afraid we'd be doing more work in the main thread which was the reason to do masking off of the main thread, so performance wise its not the right trade off here

turnipdabeets:
> @turnipdabeets if you get some time, mind testing for masking drift as well in recordings? I'm not super confident that we are not bringing back an old with these changes 🙏 In the meantime I'll address the Compose issue you mentioned above

Tested (fast-flinging masked EditTexts, masked field next to Lottie, auto-keyboard toggle) on device. Seeing some masking drift during fast scroll — masked email leaks through in some frames. Recording [here](https://us.posthog.com/shared/y6SweYMV72tYumws0HylhIlZl_YTbg?t=52)

ioannisj:
> > @turnipdabeets if you get some time, mind testing for masking drift as well in recordings? I'm not super confident that we are not bringing back an old with these changes 🙏 In the meantime I'll address the Compose issue you mentioned above
> 
> Tested (fast-flinging masked EditTexts, masked field next to Lottie, auto-keyboard toggle) on device. Seeing some masking drift during fast scroll — masked email leaks through in some frames. Recording [here](https://us.posthog.com/shared/y6SweYMV72tYumws0HylhIlZl_YTbg?t=52)

Thank you @turnipdabeets. Yeah seems like this approach is not the way to go, agreed with all concerns above. I'll close this and see if I can take a different approach here

PR #525, opened 2026-05-21
feat(logs): public capture API (PostHog.logger + PostHogLogsConfig)

turnipdabeets:

Customers asked for the same Logs SDK posthog-ios and posthog-react-native ship. This adds the Android equivalent: a `posthog.logger` you can call from anywhere, with batching, offline persistence, redaction hooks, and a rate cap. Stacked on #520 (internal plumbing, already merged).

Unit tests cover capture, severity routing, OTLP wire format, rate cap, and beforeSend hooks. End-to-end tested on a physical device — 7 scenarios including all severities, app background/foreground, offline → online recovery, force-stop persistence, manual flush, and 20+ tap stress. Records land on `/i/v1/logs` with the expected resource attributes.

## :pencil: Checklist

### If releasing new changes


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -662,6 +751,9 @@ public class PostHog private constructor(
             return
         }
 
+        // Cache for capture-time context snapshot on log records.
+        this.lastScreenName = screenTitle
```
marandaneto, line 828: now you can do this https://github.com/PostHog/posthog-android/issues/119 as a follow up
we need to make this clear that this is only if they call `screen` or if they use screen auto capture

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -622,6 +636,81 @@ public class PostHog private constructor(
...
+        message: String,
+        severity: PostHogLogSeverity,
+        attributes: Map<String, Any>?,
+    ) {
+        try {
+            if (!isEnabled()) return
+            val cfg = config ?: return
+            if (cfg.optOut) return
+            if (message.isBlank()) return
+            if (!acquireLogsRateCap(cfg)) return
```
ioannisj: I feel we should be running this after runBeforeSend below since people may anw be dropping records there, but we are incrementing window count regardless. wdyt?

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -622,6 +636,81 @@ public class PostHog private constructor(
...
+            // to avoid an unnecessary value-map copy on the hot capture path.
+            val featureFlagKeys = remoteConfig?.getActiveFeatureFlagKeys() ?: emptyList()
+
+            val record =
+                PostHogLogRecord(
+                    body = message,
+                    level = severity,
+                    // Defensive copy — caller may reuse the map; serializer
+                    // reads it later on the logs executor thread.
+                    attributes = attributes?.let { LinkedHashMap(it) } ?: emptyMap(),
```
ioannisj: This is a top-level copy only, right? What if attributes has nested maps, would we be running into issues when serializing async?

posthog/src/main/java/com/posthog/logs/PostHogLogger.kt
```diff
@@ -0,0 +1,181 @@
...
+        attributes: Map<String, Any>? = null,
+    ): Unit = log(message, PostHogLogSeverity.FATAL, attributes)
+
+    internal companion object {
+        // No-op singleton used by [PostHogInterface]'s default `logger`
+        // getter. The default is compiled into `PostHogInterface$DefaultImpls`,
+        // which lives in this module — so the call to NO_OP is intra-module
+        // and `internal` access is sufficient. Not part of the public API
+        // (iOS doesn't expose an equivalent either).
+        @JvmField
```
ioannisj: why do we need this here?

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -622,6 +636,81 @@ public class PostHog private constructor(
...
+
+            val sendable =
+                cfg.logs.runBeforeSend(record) { e ->
+                    cfg.logger.log("Error in beforeSend function: ${e.javaClass.simpleName}")
+                } ?: return
+            logsQueue?.add(sendable)
+        } catch (e: Throwable) {
+            // Only the throwable class — a hook's exception message can embed
+            // user log bodies / attributes (PII).
+            config?.logger?.log("captureLog failed: ${e.javaClass.simpleName}")
```
ioannisj: This could be a custom provided logger that could also throw right? I feel we should have a generic way of calling config.logger safely

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -622,6 +636,81 @@ public class PostHog private constructor(
...
+    }
+
+    /**
+     * Tumbling-window rate cap for log capture. Returns `true` if the call
+     * is allowed, `false` if the window's budget is exhausted. Non-positive
+     * `rateCapMaxLogs` or `rateCapWindowSeconds` disable the cap.
+     */
+    private fun acquireLogsRateCap(cfg: PostHogConfig): Boolean {
+        val max = cfg.logs.rateCapMaxLogs
+        val windowSeconds = cfg.logs.rateCapWindowSeconds
```
ioannisj: shouldn't we be synchronizing reads here with logsRateCapLock?

ioannisj (approved):
Left just a couple of comments for you, otherwise lgtm

posthog/src/main/java/com/posthog/PostHog.kt
turnipdabeets: agreed, will update

posthog/src/main/java/com/posthog/PostHog.kt
turnipdabeets: will do deep copy

posthog/src/main/java/com/posthog/logs/PostHogLogger.kt
turnipdabeets: it was to help with test boilerplate but I'm going to redo this so its only accessible to the tests

PR #529, opened 2026-05-22
fix: stop screenshot frames being dropped during animations

ioannisj:

Follow-up to #524

- Adds isOnlyAnimationRedraw to distinguish animation-driven redraws from structural layout changes
- Uses View.hasTransientState() on the decor view as the signal — Android propagates transient state up from any animating descendant, so a single check at the root covers deeply nested LottieAnimationView instances without walking the tree
- The PII concern is preserved: if a structural change occurs (mask positions could shift), isOnlyAnimationRedraw is false and the screenshot is still discarded. Only pure animation redraws (where view geometry is stable) are allowed through

Tested on a sample project with lottie animations ([here](https://us.posthog.com/shared/wSoFrC48qKqAuaq-CkJD2HyZRcrMjw?t=5))
 - snapshots don't get dropped because of animated view
 - fast scrolling does not cause masking drift

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -201,9 +201,10 @@ public class PostHogReplayIntegration(
 
     private fun onDrawCallback(decorView: View) {
         isOnDrawnCalled = true
-        // hasTransientState() propagates up from any descendant that is actively animating,
-        // so checking the decor view tells us whether the invalidation came from an animation.
-        isOnlyAnimationRedraw = decorView.hasTransientState()
+        // hasTransientState() propagates up from any descendant using a ValueAnimator (e.g. Lottie),
+        // indicating pixel-only changes with stable view geometry. We additionally exclude legacy
+        // view.animation which mutates the transformation matrix and can shift mask positions.
+        isOnlyAnimationRedraw = decorView.hasTransientState() && !decorView.isAnimationRunning()
     }
 
     private fun addView(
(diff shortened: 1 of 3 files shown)


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -193,8 +193,17 @@ public class PostHogReplayIntegration(
...
+    // than a structural layout change. When set, the isOnDrawnCalled guard is relaxed so that
+    // continuously-animating screens can still be captured (mask geometry remains valid).
+    @Volatile
+    private var isOnlyAnimationRedraw: Boolean = false
+
+    private fun onDrawCallback(decorView: View) {
         isOnDrawnCalled = true
+        // hasTransientState() propagates up from any descendant that is actively animating,
+        // so checking the decor view tells us whether the invalidation came from an animation.
+        isOnlyAnimationRedraw = decorView.hasTransientState()
```
marandaneto: we also check some other flags eg https://github.com/PostHog/posthog-android/blob/9165c3d9fa2a0607cdace567917b8d3afba823e1/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt#L717
not sure if worth adding?

marandaneto:
looks like y ou have to update the sha here https://github.com/PostHog/posthog-android/actions/runs/26275242546/job/77337802632?pr=529

marandaneto (approved):

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
marandaneto: probably worth testing on react native as well

ioannisj: Yeah, checking RN right now. I'll have a look at those additional checks as well

ioannisj: RN looks good. Added isAnimationRunning() check to tighten the check and retested MRE. Looks good

PR #530, opened 2026-05-22
feat: auto-attach $screen_name to all events

### LINKED ISSUE #119: Set `$screen_name` to all events

turnipdabeets:
> **Stacked on #525.** Base will auto-update to `main` when that lands.

Closes #119. RN and Flutter already attach `$screen_name` to every event once `screen()` has been called; Android did not. Suggested as a follow-up in #525 since that PR introduced the `lastScreenName` cache on `PostHog` (originally for log records).

Six unit tests in `PostHogScreenNameTest`: before-screen baseline, after-screen attaches, caller override wins, `reset()` clears, `$exception` carries it (locks in the cross-event scope decision), `$snapshot` does not (verifies the `appendSharedProps = false` gate).

End-to-end on a Pixel 4 across all five customer-visible scenarios — confirmed in device logcat and by querying the `events` table on `us.posthog.com`:

| Scenario | Event | `$screen_name` |
|---|---|---|
| Auto-capture (NormalActivity) | `test_auto_capture` | `NormalActivity` |
| Manual `screen("ManualScreen")` | `test_after_manual` | `ManualScreen` |
| Caller override | `test_override` | `OverrideValue` |
| After `reset()` | `test_after_reset` | _absent_ |
| Exception after `screen("ExceptionScreen")` | `$exception` | `ExceptionScreen` |

**Behavior change** to call out for reviewers: existing customers who have not disabled `captureScreenViews` will start seeing `$screen_name` attached to `$exception`, `$feature_view`, `$identify`, etc. — not only `$screen` events. Opt-out is `captureScreenViews = false` plus avoiding manual `PostHog.screen()` calls.

## :pencil: Checklist

### If releasing new changes


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -751,7 +775,10 @@ public class PostHog private constructor(
             return
         }
 
-        // Cache for capture-time context snapshot on log records.
+        if (screenTitle.isBlank()) {
```
marandaneto: better to trim before checking, and assign the lastScreenName already trimmed

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -502,6 +502,18 @@ public class PostHog private constructor(
             props.putAll(it)
         }
 
+        // Stamp $screen_name from the cache only if the caller didn't supply a
+        // meaningful value. Treat caller empty/blank as absent (almost always
+        // accidental) so the cache wins. Runs after putAll so a real caller
+        // override still takes precedence.
+        if (appendSharedProps) {
+            val cachedName = lastScreenName
+            val callerName = (props["\$screen_name"] as? String)?.trim()
```
marandaneto: see https://github.com/PostHog/posthog-android/pull/530/changes#r3292518529

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -502,6 +502,18 @@ public class PostHog private constructor(
...
+            if (!cachedName.isNullOrEmpty() && callerName.isNullOrEmpty()) {
+                props["\$screen_name"] = cachedName
```
marandaneto: flutter also captures $screen_name, and uses the android sdk, so its better to check if we are not overwriting the flutter value here

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -502,6 +502,19 @@ public class PostHog private constructor(
             props.putAll(it)
         }
 
+        // Stamp $screen_name from the cache only if the caller didn't supply a
+        // meaningful value. Runs after putAll so a real caller override (incl.
+        // posthog-flutter, which sets $screen_name in properties on capture)
+        // still takes precedence. Treat caller empty/blank as absent (almost
+        // always accidental) so the cache wins.
+        if (appendSharedProps) {
```
marandaneto: we already have a appendSharedProps check above, should we move this block inside that check?

marandaneto (approved):

.changeset/screen-name-on-all-events.md
```diff
@@ -0,0 +1,17 @@
...
+- **`PostHog.screen("")` (and whitespace-only titles) are silently dropped.** No `$screen` event is emitted and the cached value is untouched, so the last useful screen name survives. Previously these emitted a `$screen` event with an empty/whitespace `$screen_name`. Customers using empty-string as a sentinel in dashboards will see those rows disappear.
+- **Empty/whitespace `$screen_name` in `properties` falls back to the cache.** Passing `properties = mapOf("$screen_name" to "")` (or whitespace-only) on a `capture(...)` call no longer ships an empty `$screen_name` — the cached value wins. A meaningful caller-supplied value still wins.
+- **No change for `screen()` override semantics.** `PostHog.screen("Home", properties = mapOf("$screen_name" to "Override"))` continues to ship `$screen_name = "Override"` on the `$screen` event itself (Kotlin's `putAll` already let the caller override).
```
marandaneto: we are just adding a new prop $screen_name so i dont think we need to call this out that explicitly

ioannisj:
Hard to review cause PR diff is not clean here. I think this may need to be rebased?

turnipdabeets:
@Piccirello this PR seems stuck on `ensure-pinned-actions`. Could it be that it previous was pointing to a non main branch and now it is? Could you help me run this so I can merge 🙏

PR #533, opened 2026-05-25
fix(server): include group context in $feature_flag_called dedupe key

### LINKED ISSUE #3651: $feature_flag_called dedupe does not include group context for group-scoped flags
## Summary

In `posthog-node`, `$feature_flag_called` events appear to be deduped by `(distinctId, flag key, response)` only. The group context passed to feature flag evaluation is attached to the captured event, but does not appear to participate in the SDK-level dedupe key.

For group-scoped feature flags / experiments, this can suppress exposure events for a second group when the same user evaluates the same flag and receives the same response in both groups.

I'm not sure if this is intended behavior or if there's a recommended usage pattern we're missing, so filing this as a question / possible bug.

## Why this matters

For a group-scoped experiment, assignment can be based on the group key, and experiment metrics may join exposure events using a group property like `$group_0`.

If the same user visits two different groups and receives the same variant in both, the second group may not get a `$feature_flag_called` event. That means group-level exposure counts and warehouse-backed experiment joins can undercount exposed groups.

## Reproduction

Using `posthog-node@5.34.3`:

```ts
import { PostHog } from 'posthog-node'

const client = new PostHog('ph_test_key', {
  host: 'http://127.0.0.1:9',
  flushAt: 9999,
  flushInterval: 0,
})

const captured: any[] = []
client.capture = (event: any) => captured.push(event)

function fire({
  distinctId,
  response,
  org,
}: {
  distinctId: string
  response: string
  org: string
}) {
  // Calling the internal method here only to isolate the dedupe behavior.
  ;(client as any)._captureFlagCalledEventIfNeeded({
    distinctId,
    key: 'group-scoped-experiment',
    response,
    groups: { org },
    properties: {
      $feature_flag: 'group-scoped-experiment',
      $feature_flag_response: response,
      '$feature/group-scoped-experiment': response,
    },
  })
}

fire({ distinctId: 'user_123', response: 'control', org: 'org_A' })
fire({ distinctId: 'user_123', response: 'control', org: 'org_B' })
fire({ distinctId: 'user_123', response: 'test', org: 'org_B' })
fire({ distinctId: 'user_456', response: 'control', org: 'org_B' })

console.log(
  captured.map(({ distinctId, event, groups, properties }) => ({
    distinctId,
    event,
    groups,
    response: properties.$feature_flag_response,
  })),
)
```

## Actual behavior

The second call is skipped:

```json
[
  {
    "distinctId": "user_123",
    "event": "$feature_flag_called",
    "groups": { "org": "org_A" },
    "response": "control"
  },
  {
    "distinctId": "user_123",
    "event": "$feature_flag_called",
    "groups": { "org": "org_B" },
    "response": "test"
  },
  {
    "distinctId": "user_456",
    "event": "$feature_flag_called",
    "groups": { "org": "org_B" },
    "response": "control"
  }
]
```

`user_123 + control + org_B` is not emitted because `user_123 + group-scoped-experiment + control` was already emitted for `org_A`.

## Expected / question

For group-scoped flags, should the dedupe key include group context, at least the aggregation group used by the flag?

For example:

```ts
distinctId + flag key + response + relevant group type/key
```

Or is the intended usage for group-scoped experiments to pass the group key as `distinctId`, e.g. `evaluateFlags(orgId, { groups: { org: orgId } })`?

## Relevant SDK code

In current `posthog-js` main, the dedupe key appears to ignore groups:

```ts
const { distinctId, key, response, groups, disableGeoip, properties } = params
const featureFlagReportedKey = `${key}_${response}`

if (
  distinctId in this.distinctIdHasSentFlagCalls &&
  this.distinctIdHasSentFlagCalls[distinctId].includes(featureFlagReportedKey)
) {
  return
}
```

`groups` is then attached only when the event is captured:

```ts
this.capture({
  distinctId,
  event: '$feature_flag_called',
  properties,
  groups,
  disableGeoip,
})
```

## Environment

- Observed with `posthog-node@5.34.3`
- Current npm latest checked: `posthog-node@5.34.10`
- Current `PostHog/posthog-js` main appears to have the same dedupe shape

## Context

We hit this while instrumenting a group-scoped A/A experiment where:

- the flag is scoped to an `org` group
- the exposure event is `$feature_flag_called`
- the experiment metric joins exposure using the org group key
- the same authenticated user can switch between multiple orgs in the same app/server process

The local workaround is to use a fresh PostHog client for this specific exposure path so the SDK-level dedupe cache does not span orgs, while keeping a singleton client for other server-side usage.

gustavohstrassburger:

In `PostHogFeatureFlagCalledCache.add`, the per-`distinctId` LRU dedupe key only included `(distinctId, flagKey, value)`. For group-scoped flags this meant that when the same user was evaluated under a different group, no new `$feature_flag_called` event was fired — causing per-group exposure undercount for experiments scoped to a group key.

This affects two modules:
- **`posthog/`** (PostHogStateless / cache) — `captureFeatureFlagCalledEvent` now takes an optional `groups` parameter (added via `@JvmOverloads` so the existing Java-callable signature still exists). The cache key gets a canonical `groupsRepr` built from the sorted entries so two equal maps with keys in a different insertion order still dedupe.
- **`posthog-server/`** — `PostHogFeatureFlagEvaluations` now carries the `groups` it was evaluated under and passes them to the host's dedup-aware capture call.

Mirrors the posthog-node fix in PostHog/posthog-js#3658 (which closes PostHog/posthog-js#3651). Both SDKs share the same dedupe shape, so backend evaluation needs the same change.

Added 4 new tests in `PostHogFeatureFlagCalledCacheTest` and 1 in `PostHogFeatureFlagEvaluationsTest`:

Cache-level (`posthog/`):
- `same flag and value with different group contexts are tracked separately` — two cache entries.
- `same flag and value with same group context dedupe` — one cache entry.
- `same flag and value with same groups in different key order dedupe` — one cache entry (canonicalized).
- `null vs empty vs unset groups all dedupe to no-group bucket` — single entry covers all three.

Snapshot-level (`posthog-server/`):
- `isEnabled propagates groups to the host so dedup can include group context` — confirms the snapshot threads its evaluation groups through to the host's capture call.

`./gradlew :posthog:test :posthog-server:test` is green; `./gradlew :posthog:apiDump :posthog-server:apiDump` regenerated the public `.api` files.

## :pencil: Checklist

### If releasing new changes

---

### DIFF
--- a/posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
+++ b/posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
@@ -32,6 +32,7 @@ public class PostHogFeatureFlagEvaluations internal constructor(
     private val responseError: String?,
     private val host: EvaluationsHost,
     initialAccessed: Set<String> = emptySet(),
+    private val groups: Map<String, String>? = null,
 ) {
     private val flagMap: Map<String, FeatureFlag> = Collections.unmodifiableMap(LinkedHashMap(flagMap))
     private val locallyEvaluated: Map<String, Boolean> = Collections.unmodifiableMap(HashMap(locallyEvaluated))
@@ -158,6 +159,7 @@ public class PostHogFeatureFlagEvaluations internal constructor(
             responseError = responseError,
             host = host,
             initialAccessed = emptySet(),
+            groups = groups,
         )
     }
 
@@ -191,7 +193,7 @@ public class PostHogFeatureFlagEvaluations internal constructor(
             props["\$feature_flag_error"] = error.joinToString(",")
         }
 
-        host.captureFeatureFlagCalled(distinctId, key, value, props)
+        host.captureFeatureFlagCalled(distinctId, key, value, props, groups)
     }
 
     public companion object {
--- a/posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
+++ b/posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
@@ -32,10 +32,11 @@ public class PostHogFeatureFlagEvaluations internal constructor(
     private val responseError: String?,
     private val host: EvaluationsHost,
     initialAccessed: Set<String> = emptySet(),
-    private val groups: Map<String, String>? = null,
+    groups: Map<String, String>? = null,
 ) {
     private val flagMap: Map<String, FeatureFlag> = Collections.unmodifiableMap(LinkedHashMap(flagMap))
     private val locallyEvaluated: Map<String, Boolean> = Collections.unmodifiableMap(HashMap(locallyEvaluated))
+    private val groups: Map<String, String>? = groups?.let { Collections.unmodifiableMap(HashMap(it)) }
 
     private val accessLock = Any()
     private val accessed: MutableSet<String> = HashSet(initialAccessed)
--- a/.changeset/include-group-context-in-flag-called-dedupe.md
+++ b/.changeset/include-group-context-in-flag-called-dedupe.md
@@ -1,0 +1,6 @@
+---
+"posthog": patch
+"posthog-server": patch
+---
+
+Include group context in the `$feature_flag_called` LRU dedupe key so group-scoped flags fire a separate event for each group a user is evaluated under, instead of being dedup-ed against the first group context the same `(distinctId, flagKey, value)` was seen under. The groups are canonicalized order-independently so two equal maps built in different insertion orders still dedupe to one event.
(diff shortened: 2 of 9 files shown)


posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
```diff
@@ -32,6 +32,7 @@ public class PostHogFeatureFlagEvaluations internal constructor(
     private val responseError: String?,
     private val host: EvaluationsHost,
     initialAccessed: Set<String> = emptySet(),
+    private val groups: Map<String, String>? = null,
```
ioannisj: For flagMap and locallyEvaluated below we ensure immutable maps through `Collections.unmodifiableMap(HashMap(locallyEvaluated))`. Should we do the same for groups?

ioannisj (approved):
LG, left just one comment

posthog-server/src/main/java/com/posthog/server/PostHogFeatureFlagEvaluations.kt
gustavohstrassburger: Fair, fixing that.

Issue #448, opened 2026-03-03
Expose surveyPopupDelaySeconds on PostHogDisplaySurveyAppearance


adboio:
## 🚨 IMPORTANT

This issue is likely user-facing in the main PostHog app, see [`surveyVersionRequirements.ts`](https://github.com/PostHog/posthog/blob/master/frontend/src/scenes/surveys/surveyVersionRequirements.ts). If you delete or close this issue, be sure to update the version requirements list here.

## Summary

The posthog-android SDK parses `surveyPopupDelaySeconds` from the survey JSON but does not expose it on the `PostHogDisplaySurveyAppearance` delegate model, so developers implementing their own survey UI cannot access this value.

## Current State

- The `surveyPopupDelaySeconds` field exists on the internal `SurveyAppearance` model (`posthog/src/main/java/com/posthog/surveys/SurveyAppearance.kt:27`)
- The `PostHogDisplaySurveyAppearance` delegate model (what developers use) does NOT include `surveyPopupDelaySeconds`
- The `fromSurveyAppearance()` conversion method does not map this field

## Expected Behavior

The `PostHogDisplaySurveyAppearance` model should include `surveyPopupDelaySeconds` so developers implementing custom survey UIs can apply the delay themselves.

## Reference Implementation

See posthog-js browser: `packages/browser/src/extensions/surveys.tsx` — `_handlePopover()` method for how the delay is applied.

## Tracking

This is tracked in the survey SDK feature parity issue: https://github.com/PostHog/posthog/issues/45658

_This issue was generated by Claude using the `/survey-sdk-audit` skill._

ioannisj:
Closed with https://github.com/PostHog/posthog-android/pull/541

PR #541, opened 2026-05-29
feat(surveys): add posthog-android-surveys-compose default UI module

### LINKED ISSUE #102: PostHog Survey on Android
### Description

Relates to https://github.com/PostHog/posthog-js-lite/issues/110

leonhardprinz:
## Summary

Closes https://github.com/PostHog/posthog-android/issues/102, https://github.com/PostHog/posthog-android/issues/448

Adds `:posthog-android-surveys-compose`, an optional Jetpack Compose UI that renders PostHog surveys on Android. Until now the default surveys delegate only logged, so Android customers had to build their own survey UI. With this module on the classpath the core SDK auto-discovers it and renders surveys with no extra wiring — customers just add the dependency and set `surveys = true`. It's a separate module so apps that don't use surveys don't pay the APK-size cost. Ported from the iOS SwiftUI implementation.

## How it works

- **Opt-in via auto-discovery** — when no delegate is set, the core SDK reflectively loads `PostHogSurveysComposeDelegate` from the classpath. Explicit wiring (`config.surveysConfig.surveysDelegate = …`) is still supported.
- **Presentation** — a transparent, undimmed `ComponentDialog` (its own window) hosting a Material 3 `ModalBottomSheet`, shown above the foreground activity (found via an `ActivityLifecycleCallbacks`-backed `ActivityProvider`). Works over both XML and Compose hosts, on any `Activity` type, without interfering with host navigation.
- **Dismissal** — X-button only; swipe-down, touch-outside, and back are ignored.
- **Events** — the delegate invokes `onSurveyShown` / `onSurveyResponse` / `onSurveyClosed`; the host SDK emits `survey shown` / `survey sent` / `survey dismissed`. The delegate never calls `PostHog.capture` directly.

## Supported

- **Question / screen types:** open text, single choice, multiple choice, number rating (NPS 0–10, 1–5, 1–7), emoji rating (3- and 5-face plus thumbs up/down), link, and the thank-you / confirmation screen.
- **Multi-question surveys with server-driven branching** — the host SDK returns the next-question index after each answer; the sheet navigates to it (a `null` return ends the survey).
- **Configured popup delay** (`surveyPopupDelaySeconds`) before the sheet appears.
- **Theming from `PostHogDisplaySurveyAppearance`** — background, submit button, text, border, rating-button, placeholder, input, and thank-you colors/copy, with sensible defaults and a hex / CSS-color-name parser ported from iOS.

## Not yet supported

- HTML question / thank-you descriptions (rendered as plain text, matching iOS).
- Dark-mode polish (defaults tuned for light backgrounds).
- Compose UI tests + accessibility audit (verified via emulator + `@Preview` composables for now).

## Notes

- Pre-1.0 (`0.x`), versioned independently from the core SDK and excluded from binary-compatibility validation during the alpha.
- `USAGE.md` / `ARCHITECTURE.md` cover integration and the rationale for the separate module, the `ComponentDialog` presentation, appearance resolution, and the EmojiRating SVG-path approach.

## Companion documentation PRs

- posthog.com — SDK docs + surveys installation page: PostHog/posthog.com#17412
- posthog (monorepo) — surveys onboarding content + SDK feature-support matrix: PostHog/posthog#62288


posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/PostHogSurveyHost.kt
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    fun cleanup() {
+        runOnMain { dismissInternal(notifyClosed = false) }
+    }
+
+    private fun present(activity: Activity?) {
+        val survey = currentSurvey ?: return
+
+        if (activity == null || activity.isFinishing) {
```
turnipdabeets, line 159: heads up: if the app gets backgrounded during the popup delay, this fires when there's no foreground activity and we hit `dismissInternal(notifyClosed = true)` — so we send `survey dismissed` and mark it seen for a survey the user never actually saw. could we re-arm on the next resume instead of dismissing here?

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/theme/SurveyAppearance.kt
```diff
@@ -0,0 +1,142 @@
...
+        thankYouMessageCloseButtonText = thankYouMessageCloseButtonText,
+    )
+}
+
+private fun parseSurveyColorOrDefault(
+    input: String?,
+    default: Color,
+): Color {
+    val parsed = parseSurveyColor(input)
+    return if (parsed == Color.Transparent && input.isNullOrBlank()) default else parsed
```
turnipdabeets: small thing — we only fall back to the default when the input is blank. a bad/typo color like `"#GGG"` parses to transparent and we just use that, so you'd end up with e.g. an invisible submit button. probably want to fall back to the default whenever parsing fails too, not only when it's blank.

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/ui/SurveySheet.kt
```diff
@@ -0,0 +1,360 @@
...
+    onSubmit: (PostHogSurveyResponse) -> Unit,
+) {
+    val context = LocalContext.current
+    QuestionHeader(question)
+    BottomSection(
+        label = question.buttonText ?: localAppearance().submitButtonText,
+        enabled = true,
+        onClick = {
+            openLink(context, question.link)
+            onSubmit(PostHogSurveyResponse.Link(clicked = true))
```
turnipdabeets, line 282: we always send `clicked = true` here, but `openLink` quietly does nothing on a blank url and swallows `ActivityNotFoundException`, so we can report a click even when nothing actually opened. could `openLink` return whether it launched and we pass that through? e.g. `Link(clicked = openLink(context, question.link))`.

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/ui/EmojiRating.kt
```diff
@@ -0,0 +1,529 @@
...
+ */
+@Composable
+internal fun EmojiRating(
+    question: PostHogDisplayRatingQuestion,
+    selectedValue: Int?,
+    onSelect: (Int?) -> Unit,
+) {
+    val appearance = localAppearance()
+    val emojis = emojisForScale(question.scaleUpperBound - question.scaleLowerBound + 1)
+    val values = (1..emojis.size).toList()
```
turnipdabeets: question more than a blocker: emoji reports `1..size` while NumberRating uses the real scale bounds — works today but a bit fragile. also thumbs is `[ThumbsUp, ThumbsDown]`, so 👍 = 1 and 👎 = 2, which is backwards from the "1 is most negative" order the 3/5-face scales + rating buckets use. can you double check the value we record for a thumbs survey lines up with what the dashboard expects?

posthog-android-surveys-compose/gradle.lockfile
```diff
@@ -0,0 +1,116 @@
+# This is a Gradle generated file for dependency locking.
```
turnipdabeets, line 1: this lockfile only has the `debug*` configs — the other modules lock the release ones too. doesn't break the build (CI's green, default lock mode just resolves them unlocked), but it means the published artifact's release deps aren't actually pinned. probably got generated without the Android SDK around — can you re-run `make updateLocks` somewhere that builds the release variant? expect a big additive diff.

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/ui/SurveySheet.kt
```diff
@@ -0,0 +1,360 @@
...
+@Composable
+private fun UnsupportedQuestionPlaceholder(buttonLabel: String) {
+    val appearance = localAppearance()
+    Text(
+        text =
+            "This question type is not yet supported by the Compose UI module. " +
+                "Implement a custom PostHogSurveysDelegate or wait for a follow-up release.",
+        color = appearance.textColor,
+    )
+    BottomSection(label = buttonLabel, enabled = true, onClick = { /* host handles dismiss */ })
```
turnipdabeets: nit: this button doesn't do anything (`/* host handles dismiss */` but nothing does). we never hit it today since all 7 types are handled, but either wire it to `onClose` or drop the button.

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/ActivityProvider.kt
```diff
@@ -0,0 +1,67 @@
+package com.posthog.android.surveys.compose.internal
+
+import android.app.Activity
+import android.app.Application
+import android.os.Bundle
+import java.lang.ref.WeakReference
+
+/**
+ * Tracks the foreground activity so the surveys host can attach its
```
turnipdabeets: nit: doc still describes the old "attach a ViewGroup to the activity" approach — we actually present a separate-window `ComponentDialog` now. worth a quick update.

turnipdabeets (commented):
Overall this is really solid, a few inline notes below, mostly edge cases + nit. Nothing blocking that I can see.

Also the description looks like it might be a bit out of date vs what shipped — worth updating.

turnipdabeets:
It might also be good to consider: https://github.com/PostHog/posthog-android/issues/502

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/ui/EmojiRating.kt
ioannisj: Yeah thumb ordering seems to be matching what iOS and web are doing. I'll address the scale bounds though

web: https://github.com/PostHog/posthog-js/blob/65f9a67a264366a69bf24eaf8d3d283e2dbd23e0/packages/react-native/src/surveys/components/QuestionTypes.tsx#L410 and response is `idx + 1`

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/PostHogSurveyHost.kt
ioannisj: Good point. This is now addressed and I tested manually with a delay + background. When I foreground the app, the survey pops up

ioannisj: I actually need to check iOS side as well to see how this scenario is handled there

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/theme/SurveyAppearance.kt
ioannisj: Done

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/ui/SurveySheet.kt
ioannisj: Other SDKs just report `link clicked` on button press, so I don't think we should change the behavior here. Technically I will agree that it may suggest that the link was opened, but at the same time on the survey level we just track wether the link was clicked or not. I'll keep this as-is

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/ui/EmojiRating.kt
ioannisj: Addressed to match NumberRating now

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/ui/SurveySheet.kt
ioannisj: Fixed, even though we'll likely never hit this path

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/ActivityProvider.kt
ioannisj: Done

ioannisj:
> It might also be good to consider: #502

I think this is a substantial feature for both this implementation and iOS, since they are both implemented as bottom sheets, and would be a big scope creep. To be honest, I don't think anchoring makes a lot of sense on mobile either? 

I see mainly two display methods that would make sense to me: center modal and bottom-sheet. I don't feel top and side anchoring are ideal or make sense for mobile screens? Happy for a push back but I would keep the scope of this as iOS parity for first version, then maybe work on positioning for both iOS and Android together

.changeset/android-surveys-compose-ui.md
```diff
@@ -0,0 +1,20 @@
+---
+"posthog-android-surveys-compose": minor
+---
+
+Initial `0.x` release of the optional Compose survey UI (`com.posthog:posthog-android-surveys-compose`).
+This module is pre-1.0 — its public API may change between minor versions. Add the dependency
```
marandaneto, line 6: release as 0.1 so its clear its a preview

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
```diff
@@ -909,6 +955,11 @@ public class PostHogSurveysIntegration(
 
     private companion object {
         private const val NEXT_SURVEY_TRANSITION_DELAY_MS = 750L
+
+        // Fully-qualified name of the optional Compose UI delegate. Kept as a
+        // string so the core SDK has no compile-time dependency on the module.
+        private const val COMPOSE_DELEGATE_CLASS_NAME =
+            "com.posthog.android.surveys.compose.PostHogSurveysComposeDelegate"
```
marandaneto, line 966: if you dont want this to be minified, because it will fail the class name load, you need to add to the proguard rules

posthog-android-surveys-compose/USAGE.md
```diff
@@ -0,0 +1,80 @@
...
+[PostHog Android](https://github.com/PostHog/posthog-android) surveys. It renders surveys created
+in PostHog — with appearance, branching, and multi-question flows — in a Material 3 modal bottom
+sheet, in its own window on top of your foreground activity (so it never interferes with your app's
+navigation).
+
+> ⚠️ **Pre-1.0 (`0.x`).** The public surface may change between minor versions until `1.0.0`, and
+> the module is excluded from binary compatibility validation. Surveys are disabled by default and
+> must also be enabled in your PostHog project settings.
+
+## Setup
```
marandaneto: i am not a big fan of those markdowns here
do it under posthog.com/docs
so its a single source of truth, and no outdated docs
and then point to our docs here

posthog-android-surveys-compose/build.gradle.kts
```diff
@@ -0,0 +1,93 @@
...
+    // runtime — pulls in :posthog transitively
+    api(project(mapOf("path" to ":posthog-android")))
+    implementation(kotlin("stdlib-jdk8", PosthogBuildConfig.Kotlin.KOTLIN))
+
+    implementation("androidx.lifecycle:lifecycle-process:${PosthogBuildConfig.Dependencies.LIFECYCLE}")
+    // ComponentDialog — hosts the survey sheet in its own window above the host activity.
+    implementation("androidx.activity:activity:${PosthogBuildConfig.Dependencies.ANDROIDX_ACTIVITY}")
+
+    // Compose
+    implementation(platform("androidx.compose:compose-bom:2024.12.01"))
```
marandaneto: we need a sinsible version here otherwise its gonna conflict with the host app
not sure whats ideal but compose has had many breaking changes over time, so an older that works is likely better
and hopefully new versions wont break the APIs we use

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
ioannisj: I think this is handled in `build.gradle.kts` already with 

```
buildTypes {
        release {
            consumerProguardFiles("consumer-rules.pro")
        }
    }
```

which has `-keep class com.posthog.android.surveys.compose.PostHogSurveysComposeDelegate { *; }`

Not super familiar with proguard rules so let me know if this won't be enough

.changeset/android-surveys-compose-ui.md
ioannisj: yeah that's the plan. currently `surveysComposeVersion=0.0.0` and changeset bump type is set to minor, so we should get 0.1

posthog-android-surveys-compose/USAGE.md
ioannisj: yeah, this can go along with ARCHITECTURE.md. Will remove

posthog-android-surveys-compose/build.gradle.kts
ioannisj: Right, I'll see what's the minimum we can support here

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
marandaneto: yes should be added to consumer-rules.pro
you can check the android package that also has exception rules so pretty much the same

marandaneto: oh it was there already nvm, the file was collapsed, my bad

posthog-android-surveys-compose/build.gradle.kts
ioannisj: Updated to 1.3.0 which is the earliest on the current stable line. Seems that there are some breaking changes between 1.2.0 to 1.3.0 on `ModalBottomSheet` which we use, which means that we could go from "won't compile for a few older hosts" to "crashes on newer hots". 

I think I'll keep it as-is and if there's demand to go lower, then we may have to drop ModalBottomSheet and implement something custom from primitives? wdyt?

marandaneto: fine by me

> crashes on newer hots

sounds bad, so lets make sure we wrap stuff with try/catch and we swallow errors instead of crashing the host app, if not yet
better to not display anything in this case

ioannisj: Right, I'll do a pass on that

ioannisj: Added some guards at UI level. Resolving the delegate was properly guarded

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/ui/EmojiRating.kt
```diff
@@ -0,0 +1,532 @@
...
+                        Modifier
+                            .size(48.dp)
+                            .clip(CircleShape)
+                            .clickable {
+                                onSelect(if (isSelected) null else value)
+                            },
+                    contentAlignment = Alignment.Center,
+                ) {
+                    Canvas(modifier = Modifier.size(40.dp)) {
+                        drawPath(emojis[index].buildPath(size.width, size.height), color = tint)
```
turnipdabeets: `emojis[index]` will crash if it has more than 5 right? I think iOS might fall back to an empty view for anything unexpected so we could do that here.

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/theme/SurveyColor.kt
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * Parses a survey-appearance color string into a Compose [Color].
+ *
+ * Supports the web color forms PostHog appearance settings produce: named CSS
+ * colors (`"red"`, `"transparent"`), `#RGB`, `#RGBA`, `#RRGGBB`, `#RRGGBBAA`,
+ * and the same forms without the leading `#`. Returns [Color.Transparent] when
+ * the input is null, blank, or unrecognized.
+ */
+internal fun parseSurveyColor(input: String?): Color {
```
turnipdabeets, line 13: might be good to get a unit test if possible?

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
turnipdabeets, line 120: Should this also call cleanupSurveys()?

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/ui/EmojiRating.kt
```diff
@@ -0,0 +1,532 @@
...
+internal fun EmojiRating(
+    question: PostHogDisplayRatingQuestion,
+    selectedValue: Int?,
+    onSelect: (Int?) -> Unit,
+) {
+    val appearance = localAppearance()
+    val emojis = emojisForScale(question.scaleUpperBound - question.scaleLowerBound + 1)
+    // Record the rating against the question's real scale bounds (matching NumberRating and
+    // iOS), not a synthetic 1..size range. Emoji scales are 1-based on the server, so for
+    // thumbs this is 1 = 👍, 2 = 👎 — the same values iOS records and the dashboard expects.
```
turnipdabeets: My agent does this too but I feel like it's probably good to remove comments talking about another SDK since that can always change and then the doc says they match but they dont anymore.

turnipdabeets (approved):
Left some nits before merging

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/ui/EmojiRating.kt
ioannisj: yeah, agreed. need to finetune this! Simplified

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
ioannisj: yup, done

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/theme/SurveyColor.kt
ioannisj: Added

posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/ui/EmojiRating.kt
ioannisj: Correct if server ever responds with a value outside bounds. Iterating over emojis instead just to be safe

ioannisj:
Will get this merged first thing tomorrow so I can monitor the release. Thanx for the review all

PR #542, opened 2026-05-29
fix(compliance): add /get_feature_flag endpoint to the test adapter

turnipdabeets:
**TL;DR:** the compliance adapter was missing `/get_feature_flag` → harness 404s → **0/16** flag tests. This adds the endpoint (mirrors posthog-ios).

### DIFF
--- a/sdk_compliance_adapter/src/main/kotlin/com/posthog/compliance/ComplianceAdapter.kt
+++ b/sdk_compliance_adapter/src/main/kotlin/com/posthog/compliance/ComplianceAdapter.kt
@@ -18,9 +18,18 @@ import io.ktor.server.routing.post
 import io.ktor.server.routing.routing
 import okhttp3.Interceptor
 import okhttp3.Response
+import java.io.File
 import java.util.UUID
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
 import java.util.concurrent.locks.ReentrantLock
 import kotlin.concurrent.withLock
+
+// File-backed event queue; wiped between tests (see /init and /reset).
+private const val QUEUE_STORAGE_PREFIX = "/tmp/posthog-queue"
+
+// Settle window after flush() so events land before the next test (generous for Docker).
+private const val FLUSH_SETTLE_MS = 2000L
 
 /**
  * PostHog Android SDK Compliance Adapter
@@ -80,6 +89,8 @@ data class FlagRequest(
     val person_properties: Map<String, Any>? = null,
     val groups: Map<String, String>? = null,
     val group_properties: Map<String, Map<String, Any>>? = null,
+    // Decoded for harness parity but not applied: no public per-request geoip toggle and
+    // the SDK always loads remotely (known SDK gaps).
     val disable_geoip: Boolean? = null,
     val force_remote: Boolean? = null,
 )
@@ -233,6 +244,9 @@ fun main() {
                     // Close existing instance if any
                     AdapterContext.postHog?.close()
 
+                    // close() leaves queue files on disk; wipe so they don't replay into this test.
+                    File(QUEUE_STORAGE_PREFIX).deleteRecursively()
+
                     // Create new config
                     val flushIntervalMs = req.flush_interval_ms ?: 100
                     val flushIntervalSeconds = maxOf(1, flushIntervalMs / 1000) // Min 1 second
@@ -259,7 +273,7 @@ fun main() {
                     req.max_retries?.let { config.maxRetries = it }
 
                     // Set storage prefix for file-backed queue
-                    config.storagePrefix = "/tmp/posthog-queue"
+                    config.storagePrefix = QUEUE_STORAGE_PREFIX
 
                     // Set minimal context to provide $lib and $lib_version
                     config.context = TestPostHogContext("posthog-android", config.sdkVersion)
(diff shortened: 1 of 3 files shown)


sdk_compliance_adapter/src/main/kotlin/com/posthog/compliance/ComplianceAdapter.kt
```diff
@@ -305,6 +329,45 @@ fun main() {
...
+                // Apply person properties for flag evaluation without reloading yet, so exactly
+                // one /flags request is made below (the harness asserts request counts).
+                req.person_properties?.takeIf { it.isNotEmpty() }?.let {
+                    ph.setPersonPropertiesForFlags(it, reloadFeatureFlags = false)
+                }
+
+                // Register groups (with their properties) so they're included in the /flags body.
+                req.groups?.forEach { (type, key) ->
+                    ph.group(type, key, req.group_properties?.get(type))
+                }
```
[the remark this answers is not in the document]
turnipdabeets: Accurate, thanks. There's no way around this with the public android API: `group()` is the only way to put a group into the `/flags` body, and it unconditionally emits `$groupidentify` and reloads flags on a new group — there's no suppression parameter (unlike `setPersonPropertiesForFlags(reloadFeatureFlags = false)`, which the adapter does use for person properties). So the extra `/flags` + `$groupidentify` are inherent to using the public API for groups. It only affects the group-bearing flag tests (`groups_round_trip`), which are already red because `/init` fires a `/flags` regardless on the SDK side (every flag count is +1), so suppressing this wouldn't flip the test. Leaving the adapter on the public `group()` API.

sdk_compliance_adapter/src/main/kotlin/com/posthog/compliance/ComplianceAdapter.kt
```diff
@@ -305,6 +329,45 @@ fun main() {
...
+                val latch = java.util.concurrent.CountDownLatch(1)
+                ph.reloadFeatureFlags { latch.countDown() }
+                latch.await(5, java.util.concurrent.TimeUnit.SECONDS)
+
+                // Resolve the value, capturing the documented $feature_flag_called side effect.
+                val value = ph.getFeatureFlag(req.key, defaultValue = null, sendFeatureFlagEvent = true)
+
+                // Flush that event and wait so it can't spill into the next test's mock window.
+                ph.flush()
+                Thread.sleep(2000)
```
marandaneto: do we need this? Is there a better approach, like using a latch, waiting for the flush, or something?
Sleep slows down tests

marandaneto:
bsed on https://github.com/PostHog/posthog-android/pull/542#issuecomment-4577162982 that didnt work?

sdk_compliance_adapter/src/main/kotlin/com/posthog/compliance/ComplianceAdapter.kt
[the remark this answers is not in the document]
dustinbyrne: Is `setGroupPropertiesForFlags` not viable here?

sdk_compliance_adapter/src/main/kotlin/com/posthog/compliance/ComplianceAdapter.kt
```diff
@@ -227,6 +246,11 @@ fun main() {
                             flushIntervalSeconds = flushIntervalSeconds,
                             debug = true,
                             preloadFeatureFlags = false,
+                            // Don't load remote config on init either — it chains into a /flags
+                            // request, which would add a spurious request to every test's count
```
ioannisj: Not sure if this is the case cause even remoteConfig should only chain /flags call if `preloadFeatureFlags = true`. If it doesn't then I think this should be fixed?

sdk_compliance_adapter/src/main/kotlin/com/posthog/compliance/ComplianceAdapter.kt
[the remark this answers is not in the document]
turnipdabeets: addressed in 7adafd3

sdk_compliance_adapter/src/main/kotlin/com/posthog/compliance/ComplianceAdapter.kt
turnipdabeets: agreed, but tried this on iOS and it regressed 6 tests.

sdk_compliance_adapter/src/main/kotlin/com/posthog/compliance/ComplianceAdapter.kt
turnipdabeets: Fixed in 639d0cd by swapping to SDK.close() in the adapter's /reset, matching the iOS adapter.

turnipdabeets:
> bsed on [(earlier comment)](https://github.com/PostHog/posthog-android/pull/542#issuecomment-4577162982) that didnt work?

For some reason I can't see this comment

sdk_compliance_adapter/src/main/kotlin/com/posthog/compliance/ComplianceAdapter.kt
marandaneto: either tests or the mock server would have to change to make this possible
most if not all of this should be gone IMO https://github.com/search?q=repo%3APostHog%2Fposthog-android+Thread.sleep&type=code
can be a follow up ofc
once we start adding sleeps here and there, its hard to stop, take a step back, and make it in a way that we dont need it anymore

marandaneto:
> # posthog-android Compliance Report

its the 'posthog-android Compliance Report' where all tests were still not passing, so your changes didnt have any effect
the CI job updated the comment so now at least some are passing

sdk_compliance_adapter/src/main/kotlin/com/posthog/compliance/ComplianceAdapter.kt
turnipdabeets: updated, no more sleep

dustinbyrne (approved):
the rest of the failing flags tests are failing because this SDK is being tested as a server-side SDK, which I suppose is outside the scope of this PR

PR #546, opened 2026-06-02
fix: always invoke reloadFeatureFlags callback

turnipdabeets:

https://posthoghelp.zendesk.com/agent/tickets/58925 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/61593)

`reloadFeatureFlags(onFeatureFlags)` promises to invoke `onFeatureFlags` when the reload finishes, but several early-returns skipped it:

- `if (!isEnabled())` — SDK disabled or opted-out
- `if (distinctId.isBlank())` in `loadFeatureFlagsRequest`

When the callback is never invoked, any caller awaiting it stalls. This surfaced via the Flutter SDK, whose `reloadFeatureFlags()` returns a `Future` that completes from this callback — `await Posthog().reloadFeatureFlags()` would hang forever in those states. (Once execution reaches the network load, the callback already fires on every outcome via `executeFeatureFlags`, so only the front-door guards were affected.)

`./gradlew :posthog:compileKotlin` passes. Existing `reloadFeatureFlags` behavior is unchanged on the normal path; the change only adds the callback invocation to the two guard returns. Happy to add a unit test asserting the callback fires when `!isEnabled()` if preferred.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.changeset/reload-flags-callback-contract.md
+++ b/.changeset/reload-flags-callback-contract.md
@@ -1,0 +1,5 @@
+---
+"posthog": patch
+---
+
+`reloadFeatureFlags` now always invokes its completion callback, including when the SDK is disabled/opted-out or the distinct ID is blank. Previously these early-returns skipped the callback, which could leave callers that await it (e.g. the Flutter SDK's `reloadFeatureFlags`) hanging indefinitely.
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -1249,10 +1249,19 @@ public class PostHog private constructor(
         }
     }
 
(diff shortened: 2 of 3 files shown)


dustinbyrne (approved):
makes sense

turnipdabeets:
Re: @greptile-apps review on `302c9fd`:

- **Issue 2 (no tests for early-return callback)** — addressed in `acf400c`, which adds `reloadFeatureFlags invokes callback when not enabled` and routes the early-return callbacks through a `notifyFeatureFlagsCallback` try/catch helper. Note: only the `!isEnabled()` guard is covered; the blank-`distinctId` guard case is still untested.
- **Issue 1 (`internalOnFeatureFlags`/session replay skipped on blank `distinctId`)** — not addressed; the blank-`distinctId` path still invokes only the external callback. This is pre-existing behaviour rather than a regression from this PR.

PR #547, opened 2026-06-02
fix: keep session replay active across identify/reset

turnipdabeets:

https://posthoghelp.zendesk.com/agent/tickets/58925 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/61593)

After an in-session `identify()` / `reset()`, session replay stopped recording and only resumed after an app restart.

The recording config (linked flag, sample rate, endpoint) comes only from `/config`, fetched once at startup. A `/flags` reload (triggered by `identify()`/`reset()`) doesn't carry it, and the SDK was treating that absence as "recording disabled" — clearing the config and the recording flag until the next startup.

The same root cause affected two sibling project-level configs that also come only from `/config`: **error tracking** (`autocaptureExceptions`) and **network performance capture** (`network_timing`). Both were wiped on `reset()` and stayed off until an app restart. This PR fixes all three: their config is preserved across `reset()` and re-armed from the cache on the next `/flags` reload.

Verified on a real Android device: `isSessionReplayActive` goes `false → true` after `reset()` → `identify()` → reload, without an app restart. The configs are project-level, so keeping them across an identity change doesn't change behavior for disabled projects (cleared on an explicit `false` from `/config`) or linked-flag projects (re-evaluated against the new user's flags).

Added unit tests covering: re-arm across reset for all three configs; linked-flag re-evaluation turning replay OFF for a user outside the rollout; an explicit `false` from `/config` evicting each cached config; the Gson round-trip for a variant linked flag; and `STRINGIFIED_KEYS` pruning in the Android preferences `clear()`.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
@@ -405,12 +405,13 @@ public class PostHogRemoteConfig(
             config.logger.log("No cached session replay config to re-evaluate; replay stays disabled.")
             return
         }
-        // persist = false: just read from the cache, no need to write it back.
         processSessionRecordingConfig(recordingConfig, persist = false)
     }
 
     private fun processSessionRecordingConfig(
         sessionRecording: Any?,
+        // persist=true writes the config to the cache (the /config path); false only evaluates it
+        // in-memory (the /flags re-arm path, which just read it from that same cache).
         persist: Boolean = true,
     ) {
         when (sessionRecording) {
@@ -511,7 +512,6 @@ public class PostHogRemoteConfig(
             config.logger.log("No cached error tracking config to re-evaluate; autocapture stays disabled.")
             return
         }
-        // persist = false: just read from the cache, no need to write it back.
         processErrorTrackingConfig(errorTracking, persist = false)
     }
 
@@ -565,7 +565,6 @@ public class PostHogRemoteConfig(
             config.logger.log("No cached capture performance config to re-evaluate; network timing stays disabled.")
             return
         }
-        // persist = false: just read from the cache, no need to write it back.
         processCapturePerformanceConfig(capturePerformance, persist = false)
     }
 
@@ -647,6 +646,12 @@ public class PostHogRemoteConfig(
                             """Feature flags are quota limited, flags could not be updated.
                                     Learn more about billing limits at https://posthog.com/docs/billing/limits-alerts""",
                         )
+                        // Flags are quota limited, but session replay / error tracking / capture
+                        // performance config come from /config (not flags), so still re-arm them from
+                        // the cache against the flags we already have, instead of leaving them disabled.
+                        reevaluateSessionReplayFromCachedConfig()
+                        reevaluateCapturePerformanceFromCachedConfig()
+                        reevaluateErrorTrackingFromCachedConfig()
                         return@let
                     }
 
@@ -1227,8 +1232,9 @@ public class PostHogRemoteConfig(
 
         synchronized(remoteConfigLock) {
             clearSurveys()
-            // Zero the in-memory flags but keep the cached config so a reload can re-arm them
-            // (an explicit `false` from /config still evicts the cache).
+            // Zero the in-memory flags but keep the cached config so a reload can re-arm them.
+            // Deliberately NOT clearErrorTracking()/clearCapturePerformance(): those also evict the
+            // cache. An explicit `false` from /config is the only path that evicts.
             autoCaptureExceptions = false
             captureNetworkTiming = false
         }
--- a/posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
@@ -398,10 +398,14 @@ public class PostHogRemoteConfig(
         return milliseconds
     }
 
-    // Re-evaluates sessionReplayFlagActive from the persisted config (survives reset) + current flags.
-    private fun reevaluateSessionReplayFromStoredConfig() {
+    // Re-evaluates sessionReplayFlagActive from the cached config (survives reset) + current flags.
+    private fun reevaluateSessionReplayFromCachedConfig() {
         @Suppress("UNCHECKED_CAST")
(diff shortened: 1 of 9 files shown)


posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
```diff
@@ -637,8 +651,13 @@ public class PostHogRemoteConfig(
...
 
-                    // since flags might have changed, we need to check if session recording is active again
-                    processSessionRecordingConfig(it.sessionRecording)
+                    // /flags doesn't carry sessionRecording; re-evaluate from the cached config instead of clobbering it.
+                    val responseSessionRecording = it.sessionRecording
+                    if (responseSessionRecording is Map<*, *>) {
+                        processSessionRecordingConfig(responseSessionRecording)
+                    } else {
+                        reevaluateSessionReplayFromLastConfig()
+                    }
```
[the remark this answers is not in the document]
turnipdabeets: Good catch — fixed in c55bf8d. Switched to `!= null` so an explicit `sessionRecording: false` from /flags still disables recording; only a truly absent (null) value re-evaluates from the cached config.

posthog/src/test/java/com/posthog/internal/PostHogRemoteConfigTest.kt
```diff
@@ -74,6 +74,110 @@ internal class PostHogRemoteConfigTest {
...
+
+        // Must re-arm from the cached config, not stay clobbered to false.
+        assertTrue(sut.isSessionReplayFlagActive())
+
+        sut.clear()
+        http.shutdown()
+    }
+
+    @Test
+    fun `explicit sessionRecording false on a flags reload disables replay`() {
```
dustinbyrne: `sessionRecording` shouldn't ever be returned by `/flags`, this is an artifact of when we used to hit the endpoint via `/flags?config=true`, which appended remote config to the flags response.

We no longer do this, and this is part of the reason the rest of this change is necessary

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
dustinbyrne, line 1228: `consoleLogRecordingEnabled` is session replay configuration derived from remote config and it's not restored anywhere afaict

dustinbyrne (commented):
The `consoleLogRecordingEnabled` state looks like a bug worth fixing before merging

posthog/src/test/java/com/posthog/internal/PostHogRemoteConfigTest.kt
turnipdabeets: I think it might still be there for backwards compatibilities with old self hosted posthog server? But if you think it's ok I can try to remove that.

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
turnipdabeets: good catch, updated

posthog/src/test/java/com/posthog/internal/PostHogRemoteConfigTest.kt
dustinbyrne: I don't believe there's any code path that allows us to request remote config from the flags endpoint anymore, but definitely worth double checking

turnipdabeets: Don't see anything either, removed!

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
```diff
@@ -398,6 +398,17 @@ public class PostHogRemoteConfig(
         return milliseconds
     }
 
+    // Restores the full recording config from cache (survives reset), re-evaluated against current flags.
+    private fun reevaluateSessionReplayFromCachedConfig() {
+        @Suppress("UNCHECKED_CAST")
+        val recordingConfig = config.cachePreferences?.getValue(SESSION_REPLAY) as? Map<String, Any?>
```
ioannisj: nit: This is probably okay since we only cache in case of Map<String, Any> (can you please verify?). But since processSessionRecordingConfig() checks for Boolean, should we be casting to Any type here just to be future proof?

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
ioannisj: Missed, what changed here to handle it? Or was it already handled?

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
```diff
@@ -637,8 +648,9 @@ public class PostHogRemoteConfig(
                         this.featureFlagPayloads = normalizedPayloads
                     }
 
-                    // since flags might have changed, we need to check if session recording is active again
-                    processSessionRecordingConfig(it.sessionRecording)
+                    // /flags carries flag evaluations only; session recording config comes from
+                    // /config. Re-arm from the cached config, re-evaluated against the new flags.
+                    reevaluateSessionReplayFromCachedConfig()
```
ioannisj, line 695: We are reading from disk on every flags call (which is probably fine) but if we end up not clearing config mem (discussed [here](https://github.com/PostHog/posthog-ios/pull/630#discussion_r3351353903)) we could use that instead

ioannisj (approved):
LG, some discussion on reset() and local mem on iOS PR as well, but keeping as-is is probably fine as well

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
turnipdabeets: think you might have reviewed outdated file? Already addressed.

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
turnipdabeets: Android's a bit different from iOS here, I think the cache here is the single source of truth. I think Android SharedPreferences is already in-memory?

turnipdabeets:
@dustinbyrne included error tracking and network performance capture now

PR #553, opened 2026-06-08
feat: add public captureLog API with W3C trace correlation

turnipdabeets:

Android was the only PostHog SDK without a public `captureLog()` entry point that accepts W3C distributed-tracing fields. iOS, web, and React Native all let you attach `trace_id` / `span_id` / `trace_flags` to a log; on Android you could only set them by mutating the record inside a `beforeSend` hook — not a real API. The logs spec requires this entry point.

The record model (`PostHogLogRecord`) and OTLP serializer already carried the three fields and shipped them on the wire (as `traceId` / `spanId` / `flags`). The only gap was a public input path, so this is a thin wiring change rather than new plumbing.

- Added unit tests in `PostHogLogsCaptureTest`: trace fields land on the OTLP wire; omitted when absent (level defaults to `info`); explicit `traceFlags = 0` still emits `"flags":0` (the W3C sampled-false case, guarding the null-vs-zero trap).
- Full suite green: `PostHogLogsCaptureTest` 11/11, plus `PostHogLogRecordTest` / `PostHogLogsOTLPTest`.
- `apiCheck` green across modules after regenerating `posthog.api` (purely additive).

**Note for reviewers:** this adds a new `public abstract fun captureLog` to `PostHogInterface`. It's additive for callers and the regenerated `posthog.api` is additive-only; consistent with how `captureException` (#300) and `captureFeatureView`/`captureFeatureInteraction` (#416) were added. The `logger` facade signature/behavior is unchanged.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.changeset/log-capture-trace-correlation.md
+++ b/.changeset/log-capture-trace-correlation.md
@@ -1,0 +1,6 @@
+---
+"posthog": minor
+"posthog-android": minor
+---
+
+Add a public `captureLog()` API for capturing logs with optional W3C trace correlation (`traceId`/`spanId`/`traceFlags`), matching iOS, web, and React Native. The `logger` facade is unchanged.
--- a/posthog/api/posthog.api
+++ b/posthog/api/posthog.api
@@ -28,6 +28,7 @@ public final class com/posthog/PostHog : com/posthog/PostHogStateless, com/posth
 	public fun captureException (Ljava/lang/Throwable;Ljava/util/Map;)V
 	public fun captureFeatureInteraction (Ljava/lang/String;Ljava/lang/String;)V
 	public fun captureFeatureView (Ljava/lang/String;Ljava/lang/String;)V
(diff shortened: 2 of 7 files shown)


turnipdabeets:
Thanks @greptile-apps — both are fair, but I'm leaving them as-is:

1. **`assertFalse` vs `assertEquals(false, …)`** — agreed it's more idiomatic in isolation, but `PostHogLogsCaptureTest` already standardizes on `assertEquals(false, …)` (e.g. the existing `beforeSend` test). Switching only the new lines would make this file inconsistent with itself. (Sibling file `PostHogLogRecordTest` does use `assertFalse` — different local convention, so the new test I just added there matches *that* file.)

2. **Parameterize the two presence tests** — they read clearly as separate cases and there are only two. I'd rather keep them explicit than add a data-provider for marginal dedup. Happy to revisit if we add several more trace-field cases.

Separately: an independent edge-case pass flagged the null-vs-zero boundary for `traceFlags`, so I added a storage round-trip test pinning `traceFlags = 0` (a03a910).

dustinbyrne (approved):

PR #569, opened 2026-06-12
feat(errortracking): ignoredExceptionTypes config to skip RN-duplicate JS errors

### LINKED ISSUE #567: Native crash autocapture double-reports fatal React Native JS errors (JavascriptException)
### Problem

When the React Native SDK enables native crash autocapture (`@posthog/react-native-plugin`, `errorTracking.autocapture.nativeCrashes`, see PostHog/posthog-js#3677), a **fatal JS error can be captured twice**:

1. The JS layer captures it via `uncaughtExceptions` autocapture (with the JS stack trace), and
2. React Native rethrows fatal JS errors on the native side as a `com.facebook.react.common.JavascriptException` on the main thread. posthog-android's uncaught-exception handler then captures it as a separate native crash (with a native stack trace).

The result is two `$exception` events for one logical error.

### Suggested fix

posthog-android should expose a way to ignore exceptions that originate from React Native's fatal-JS path, and/or skip reporting throwables of type `com.facebook.react.common.JavascriptException`, so the JS-captured event is the single source of truth.

For reference, sentry-react-native does exactly this on the native side:
```java
// React native internally throws a JavascriptException.
// we want to ignore it on the native side to avoid sending it twice.
options.addIgnoredExceptionForType(JavascriptException.class);
```
(https://github.com/getsentry/sentry-react-native/blob/f170ec355cd622d3e935478a23660ad4cb97e896/packages/core/android/src/main/java/io/sentry/react/RNSentryStart.java#L355)

### Notes

- This pairs with the iOS equivalent (sibling issue in posthog-ios: PostHog/posthog-ios#653).
- Worth verifying empirically against current RN (0.79+) release builds, since the exact rethrow path can change between RN versions.
- Discovered while reviewing PostHog/posthog-js#3677.

tsushanth:
## Summary

Closes #567.

In React Native apps that enable native crash autocapture, a fatal JS error is captured twice:

1. The JS layer captures it via `uncaughtExceptions` autocapture (with the JS stack trace), and
2. React Native rethrows the same fatal JS error on the native side as `com.facebook.react.common.JavascriptException`. posthog-android's uncaught-exception handler then captures it as a second `$exception` event (with a native stack trace).

The result is two `$exception` events for one logical error.

## API

New optional field on `PostHogErrorTrackingConfig`:

```kotlin
public val ignoredExceptionTypes: MutableList<String> = mutableListOf()
```

Setup in an RN app:

```kotlin
PostHogAndroidConfig(apiKey).apply {
    errorTrackingConfig.ignoredExceptionTypes.add(
        "com.facebook.react.common.JavascriptException",
    )
}
```

Whenever the uncaught-exception handler fires, the throwable and every cause in its chain are checked against the list **by class name**:

- If any link matches → the SDK skips emitting `\$exception`, but **still chains to the next handler** (so the process termination / RN red-box / etc. behave the same way as before).
- If nothing matches (or the list is empty, the default) → behavior is unchanged.

Mirrors `sentry-android`'s [`addIgnoredExceptionForType(...)`](https://github.com/getsentry/sentry-react-native/blob/f170ec355cd622d3e935478a23660ad4cb97e896/packages/core/android/src/main/java/io/sentry/react/RNSentryStart.java#L355).

## Implementation notes

- **Class-name based, never `Class<*>::isInstance`.** This is deliberate: matching by name means apps that don't have React Native (or any other optional dependency) on their classpath are not affected at startup or call time. The SDK never has to load or even know about RN's types — the consumer just passes the FQCN string.
- **Cause-chain walk.** Real RN apps sometimes wrap the JS exception in a platform-level wrapper before it reaches the uncaught handler, so the check walks `cause` until either a match is found or a seen-set short-circuits a self-referential chain.
- **Always chains.** The `defaultExceptionHandler` is always invoked, ignored or not. The contract of `Thread.UncaughtExceptionHandler` (process termination, RN's red-box, etc.) shouldn't change based on whether we emit a `\$exception` event.
- **No public API removed.** Pure additive change. Apps that don't set `ignoredExceptionTypes` see zero behavior difference.

## Tests

Three new tests in `PostHogErrorTrackingAutoCaptureIntegrationTest`:

- `uncaughtException skips capture when throwable class is in ignoredExceptionTypes` — direct match.
- `uncaughtException skips capture when ignored class is anywhere in the cause chain` — `RuntimeException("outer", cause = JsException)`. The outermost type is *not* in the list, but the chain match still suppresses.
- `uncaughtException still captures when throwable class is not in ignoredExceptionTypes` — guards against the strip widening to all throwables when the list is non-empty.

The RN exception type is stubbed locally (`ReactNativeJavascriptExceptionStub`) so the test doesn't drag React Native into the SDK's test classpath — the name-based check exercises the same code path either way.

All 17 tests pass (14 existing + 3 new):

```
> Task :posthog:test
BUILD SUCCESSFUL in 22s
```

## Test plan

- [x] `./gradlew :posthog:test --tests "com.posthog.errortracking.*"` — 17/17 pass
- [x] `./gradlew :posthog:compileDebugKotlin` (via the test task) — clean
- [x] Manual verification: with `ignoredExceptionTypes = []`, all existing tests behave identically (no behavior change to the autocapture path).

## Sibling issue

A matching fix is needed on the iOS side per #653 (RN rethrows fatal JS as `RCTFatalException`). Happy to ship that next as a separate PR if this shape is what you want — the iOS surface would be `PostHogErrorTrackingConfig.ignoredExceptionNames: [String]` with the same cause-chain walk.

## Notes for review

@marandaneto — you suggested `beforeSend` as a "for now" workaround on the issue thread; this PR proposes the structural fix. Naming and scope are intentionally narrow: just the FQCN list, no Class-object overload, no auto-detection of RN. Happy to refactor if you'd rather keep the surface even smaller (e.g. hardcode the JavascriptException class internally and gate on an RN-detection bool) — let me know which shape you'd prefer.

### DIFF
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -687,6 +687,17 @@ public class PostHog private constructor(
         }
 
         try {
+            val ignored = config?.errorTrackingConfig?.ignoredExceptionTypes
+            if (!ignored.isNullOrEmpty()) {
+                val match = findIgnoredTypeInCauseChain(throwable, ignored)
+                if (match != null) {
+                    config?.logger?.log(
+                        "Skipping \$exception: ${match.name} (or a cause in its chain) matches ignoredExceptionTypes",
+                    )
+                    return
+                }
+            }
+
             val exceptionProperties =
                 throwableCoercer.fromThrowableToPostHogProperties(
                     throwable,
@@ -706,6 +717,21 @@ public class PostHog private constructor(
             // a captured exception to a PostHog exception event
             config?.logger?.log("captureException has thrown an exception: $e.")
         }
+    }
+
+    private fun findIgnoredTypeInCauseChain(
+        throwable: Throwable,
+        ignored: List<Class<out Throwable>>,
+    ): Class<out Throwable>? {
+        var current: Throwable? = throwable
+        var depth = 0
+        while (current != null && depth < MAX_CAUSE_DEPTH) {
+            val link = current
+            ignored.firstOrNull { it.isInstance(link) }?.let { return it }
+            current = if (link.cause === link) null else link.cause
+            depth++
+        }
+        return null
     }
 
     override fun addExceptionStep(
@@ -1803,6 +1829,10 @@ public class PostHog private constructor(
         private var defaultSharedInstance = shared
 
         private val apiKeys = mutableSetOf<String>()
+
+        // Cap on the cause-chain walk used by ignoredExceptionTypes.
+        // Guards against self-referential or cyclic causes.
+        private const val MAX_CAUSE_DEPTH: Int = 32
 
         /**
          * Captures application log records into PostHog's logs product
--- a/posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingConfig.kt
+++ b/posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingConfig.kt
@@ -37,4 +37,14 @@ public class PostHogErrorTrackingConfig
          * Record steps with [PostHog.addExceptionStep].
          */
         public val exceptionSteps: PostHogExceptionStepsConfig = PostHogExceptionStepsConfig(),
+        /**
+         * Throwable classes to skip during capture. For each captured throwable, the
+         * SDK walks the cause chain and drops the `$exception` event when any link is
+         * an instance of an entry in this list. Matching uses [Class.isInstance], so
+         * R8 / ProGuard renames don't break the filter. The downstream uncaught
+         * exception handler still runs.
+         *
+         * Defaults to empty.
+         */
+        public val ignoredExceptionTypes: MutableList<Class<out Throwable>> = mutableListOf(),
     )
--- a/posthog/src/test/java/com/posthog/PostHogTest.kt
+++ b/posthog/src/test/java/com/posthog/PostHogTest.kt
@@ -69,8 +69,6 @@ internal class PostHogTest {
         evaluationContexts: List<String>? = null,
         context: PostHogContext? = null,
         personProfiles: PersonProfiles = PersonProfiles.IDENTIFIED_ONLY,
-        exceptionStepsEnabled: Boolean = true,
-        exceptionStepsMaxBytes: Int = 32768,
     ): PostHogInterface {
         config =
             PostHogConfig(API_KEY, host).apply {
@@ -94,8 +92,6 @@ internal class PostHogTest {
                     addBeforeSend(beforeSend)
                 }
                 this.errorTrackingConfig.inAppIncludes.add("com.posthog")
-                this.errorTrackingConfig.exceptionSteps.enabled = exceptionStepsEnabled
-                this.errorTrackingConfig.exceptionSteps.maxBytes = exceptionStepsMaxBytes
                 this.context = context
                 this.personProfiles = personProfiles
             }
@@ -2541,6 +2537,69 @@ internal class PostHogTest {
         sut.close()
     }
 
+    /**
+     * Local stand-in for `com.facebook.react.common.JavascriptException` (the
+     * production target). Using a class defined here lets the test register
+     * the same `Class` reference at config time and at capture time without
+     * pulling React Native onto the classpath.
+     */
+    private class JavascriptExceptionStub(message: String) : RuntimeException(message)
+
+    @Test
+    fun `captureException drops events whose throwable matches ignoredExceptionTypes`() {
+        val http = mockHttp()
+        val url = http.url("/")
+
+        val sut = getSut(url.toString(), preloadFeatureFlags = false)
+        config.errorTrackingConfig.ignoredExceptionTypes.add(JavascriptExceptionStub::class.java)
+
+        sut.captureException(JavascriptExceptionStub("Unhandled JS Exception"))
+
+        queueExecutor.shutdownAndAwaitTermination()
+        assertEquals(0, http.requestCount)
+
+        sut.close()
+    }
+
+    @Test
+    fun `captureException drops events whose cause chain contains an ignored type`() {
+        val http = mockHttp()
+        val url = http.url("/")
+
+        val sut = getSut(url.toString(), preloadFeatureFlags = false)
+        config.errorTrackingConfig.ignoredExceptionTypes.add(JavascriptExceptionStub::class.java)
+
+        // PostHogThrowable / RuntimeException wrappers must not defeat the filter:
+        // the cause chain still surfaces the ignored type.
+        val wrapped =
+            RuntimeException(
+                "wrapper",
+                JavascriptExceptionStub("Unhandled JS Exception"),
+            )
+        sut.captureException(wrapped)
+
+        queueExecutor.shutdownAndAwaitTermination()
+        assertEquals(0, http.requestCount)
+
+        sut.close()
+    }
+
+    @Test
+    fun `captureException keeps events whose throwable is not in ignoredExceptionTypes`() {
+        val http = mockHttp()
+        val url = http.url("/")
+
+        val sut = getSut(url.toString(), preloadFeatureFlags = false)
+        config.errorTrackingConfig.ignoredExceptionTypes.add(JavascriptExceptionStub::class.java)
+
+        sut.captureException(IllegalStateException("normal failure"))
+
+        queueExecutor.shutdownAndAwaitTermination()
+        assertEquals(1, http.requestCount)
+
+        sut.close()
+    }
+
     @Test
     fun `sets default person properties on SDK setup when enabled`() {
         val http =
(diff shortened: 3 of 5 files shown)


tsushanth:
Thanks @greptile-apps — pushed [`7262292`](https://github.com/PostHog/posthog-android/pull/569/commits/7262292) addressing issues 1 and 3.

- **Issue 1 (suppress log accuracy):** reworded so the log surfaces "the throwable or a cause in its chain" matched — readers won't see a bare `RuntimeException` reported as suppressed when the ignore list only mentions `JavascriptException`.
- **Issue 3 (test-stub doc comment):** restated the actual design (the stub's own JVM name is what tests register as the ignored FQCN, exercising the same code path without dragging RN onto the classpath). The "named the same way" framing was wrong.

**Issue 2 (parameterising the new tests):** intentionally deferred. The individual test names — `skips capture when throwable class is in ignoredExceptionTypes`, `skips capture when ignored class is anywhere in the cause chain`, `still captures when throwable class is not in ignoredExceptionTypes` — read more clearly under a failure than a `@ParameterizedTest` table would, and the three cases aren't quite the structural pair Greptile reads them as (the cause-chain case threads `cause` through a wrapper, the negative case asserts the inverse contract). Happy to collapse to a parameterised form if maintainers prefer it.

All 17 tests still pass after the changes.

marandaneto:
what i'd do differently is to drop after capturing the error, where we do the beforeSend as well
so even if the event is manually captured, its also dropped
right now if you call captureException manually with a suprressed type, its still captured which is a bug

tsushanth:
Good call — pushed [d11c33d](https://github.com/PostHog/posthog-android/pull/569/commits/d11c33d) restructuring the filter to live at the `beforeSend` layer per your suggestion.

**What moved**

- New `PostHogIgnoredExceptionTypesBeforeSend` (internal) implementing `PostHogBeforeSend`. It scopes itself to `$exception` events only, reads `$exception_list[*].type` from properties, and drops if any entry's `type` is in `errorTrackingConfig.ignoredExceptionTypes`. The cause chain is already serialized in `$exception_list` outer-to-inner by `ThrowableCoercer`, so walking the list covers the wrapper case (`RuntimeException(cause=JavascriptException)`) without re-reaching for the original throwable.
- `PostHogErrorTrackingAutoCaptureIntegration.install` now calls `config.addBeforeSend(...)` for the new hook (and `removeBeforeSend(...)` on `uninstall`), so both autocapture and manual `captureException(throwable)` callers go through the same drop logic. The `isIgnored(throwable)` helper that was inside `uncaughtException` is gone; the handler now always hands off to `captureException` and the drop decision is the beforeSend's.

**Tests**

- `PostHogIgnoredExceptionTypesBeforeSendTest` — 6 cases covering: outer-type match, cause-chain match (wrapped), no-match passthrough, empty ignore list, non-`$exception` events ignored by the filter, and missing/empty `$exception_list` passthrough. The wrapped-cause case is the regression equivalent of the old `uncaughtException skips capture when ignored class is anywhere in the cause chain` test, now at the right layer.
- `PostHogErrorTrackingAutoCaptureIntegrationTest` — replaced the three old `uncaughtException skips capture …` tests with two new ones: one asserts `addBeforeSend` / `removeBeforeSend` are called with a `PostHogIgnoredExceptionTypesBeforeSend` instance on `install` / `uninstall`, the other asserts the handler always calls `captureException` regardless of the throwable's class. The drop semantics now live in the beforeSend's own unit tests.

Full posthog module suite passes locally: `./gradlew :posthog:test --tests "com.posthog.errortracking.*"` → 22 tests, 0 failures.

(The greptile "non-parameterised tests" P2 from the earlier review still applies to the `BeforeSend` tests by the same argument, but they're shaped distinctly enough — different inputs assert different invariants, not the same one with different values — that I'd default to leaving them. Easy to collapse if you'd prefer.)

posthog/src/main/java/com/posthog/errortracking/PostHogIgnoredExceptionTypesBeforeSend.kt
```diff
@@ -0,0 +1,45 @@
...
+) : PostHogBeforeSend {
+    override fun run(event: PostHogEvent): PostHogEvent? {
+        if (event.event != PostHogEventName.EXCEPTION.event) return event
+
+        val ignored = config.errorTrackingConfig.ignoredExceptionTypes
+        if (ignored.isEmpty()) return event
+
+        val list = event.properties?.get("\$exception_list") as? List<*> ?: return event
+        for (entry in list) {
+            val type = (entry as? Map<*, *>)?.get("type") as? String ?: continue
```
marandaneto: that wont work, types might be minified/not available, we have to match the type with the instance type

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingConfig.kt
```diff
@@ -30,4 +30,30 @@ public class PostHogErrorTrackingConfig
...
+         * suppresses the native duplicate so the JS-captured event is the single source
+         * of truth. Mirrors `sentry-android`'s `addIgnoredExceptionForType(...)`.
+         *
+         * The matching is purely class-name based; no `Class<*>` objects are loaded, so
+         * apps that don't have React Native (or any other optional dependency) on their
+         * classpath are not affected.
+         *
+         * Defaults to empty.
+         */
+        public val ignoredExceptionTypes: MutableList<String> = mutableListOf(),
```
ioannisj: I think I agree with @marandaneto on this, string matching will probably not work because of minification. So this should probably take a Class not a String and check via isInstance. This means that we can't use beforeSend to run these checks. 

I believe we can drop from https://github.com/PostHog/posthog-android/blob/3ca38b14944dda700564c90a2dabf2db386fa673/posthog/src/main/java/com/posthog/PostHog.kt#L646 which is called for both manual and automatic error captures 

we'll want to keep the cause-chain walk too cause we'll need that for PostHogThrowable

tsushanth:
Thanks @marandaneto and @ioannisj. Refactored to take `MutableList<Class<out Throwable>>` instead of `List<String>` and moved the check into `PostHog.captureException` so both autocapture and direct callers go through it, exactly as suggested. Matching is by `Class.isInstance` so R8 / ProGuard renaming can't silently break the filter, and the cause-chain walk is preserved (bounded by `MAX_CAUSE_DEPTH = 32` to guard against cyclic chains) so the `PostHogThrowable` wrapper from the uncaught handler still surfaces the underlying type.

`PostHogIgnoredExceptionTypesBeforeSend` is gone (and so is its register/unregister wiring on the autocapture integration); the two BeforeSend-shaped tests under `PostHogErrorTrackingAutoCaptureIntegrationTest` are replaced by three tests on `PostHogTest` covering the direct-match drop, the cause-chain drop, and the keep-when-not-matching path. 696 tests pass under `./gradlew :posthog:test`; spotless and `apiDump` are up to date.

Ready for another look whenever you have a moment.

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -651,6 +651,17 @@ public class PostHog private constructor(
...
+        val ignored = config?.errorTrackingConfig?.ignoredExceptionTypes
+        if (!ignored.isNullOrEmpty()) {
+            val match = findIgnoredTypeInCauseChain(throwable, ignored)
+            if (match != null) {
+                config?.logger?.log(
+                    "Skipping \$exception: ${match.name} (or a cause in its chain) matches ignoredExceptionTypes",
+                )
+                return
+            }
+        }
```
ioannisj: Let's move this inside try-catch below?

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingConfig.kt
```diff
@@ -30,4 +30,35 @@ public class PostHogErrorTrackingConfig
...
+         * fatal JS errors via `@posthog/react-native-plugin`'s JS-side
+         * autocapture: React Native rethrows the same fatal JS error on the
+         * native side as `com.facebook.react.common.JavascriptException`,
+         * and posthog-android would otherwise emit a second `$exception`
+         * event for the same logical error.
+         *
+         * Setting
+         * `ignoredExceptionTypes = mutableListOf(JavascriptException::class.java)`
+         * suppresses the native duplicate so the JS-captured event is the
+         * single source of truth. Mirrors `sentry-android`'s
```
ioannisj: Let's remove reference in public docs and move to header attribution if indeed necessary

posthog/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+
+- Error tracking: add `errorTrackingConfig.ignoredExceptionTypes` (a
+  `MutableList<Class<out Throwable>>`) so callers can suppress duplicate
+  captures by type. The list is consulted by `PostHog.captureException` —
+  both the autocapture handler and direct calls — and walks the cause
+  chain, so a wrapped throwable still matches. Matching is by `Class`
+  reference, not by name, so R8 / ProGuard renaming can't silently break
+  the filter. Primary use case is dropping the native rethrow of fatal JS
+  errors when the React Native plugin already captures them on the JS
+  side. Closes #567.
```
ioannisj: Lets not edit changelog directly but provide a changeset entry. Refer to [RELEASING.md](https://github.com/PostHog/posthog-android/blob/b35ee8b27fc5455c38d53c66cfd966f8cf75e897/RELEASING.md)

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingConfig.kt
ioannisj: Let's simplify this a little but, too chatty

ioannisj (commented):
Thanx @tsushanth, this is almost there. Left some last few comments + some merge conflicts to resolve

tsushanth:
Thanks @ioannisj. Squashed the branch onto current `main` (so the diff is now scoped to the 5 files this PR actually owns), addressed all four review threads:

- moved the `ignoredExceptionTypes` check inside `captureException`'s existing `try` block ([4abc99f](https://github.com/PostHog/posthog-android/pull/569/commits/4abc99f))
- removed the React Native reference from the public docstring on `errorTrackingConfig.ignoredExceptionTypes` and tightened the doc to the essentials
- dropped the direct `posthog/CHANGELOG.md` edit and added `.changeset/error-tracking-ignored-exception-types.md` per `RELEASING.md`
- `PostHogErrorTrackingAutoCaptureIntegration` now sits unchanged on top of `main` — no comment churn there either

706 tests pass under `./gradlew :posthog:test`; `make checkFormat` and `apiDump` are up to date. Ready for another pass whenever.

posthog/src/test/java/com/posthog/PostHogTest.kt
```diff
@@ -3381,199 +3440,4 @@ internal class PostHogTest {
 
         sut.close()
     }
-
-    private fun exceptionStepMessages(event: com.posthog.PostHogEvent): List<Any?> {
```
ioannisj, line 3385: These tests should not be removed here

posthog/src/test/java/com/posthog/PostHogTest.kt
```diff
@@ -69,8 +69,6 @@ internal class PostHogTest {
         evaluationContexts: List<String>? = null,
         context: PostHogContext? = null,
         personProfiles: PersonProfiles = PersonProfiles.IDENTIFIED_ONLY,
-        exceptionStepsEnabled: Boolean = true,
-        exceptionStepsMaxBytes: Int = 32768,
```
ioannisj, line 73: Same, probably stale commits from main

posthog/src/test/java/com/posthog/PostHogTest.kt
```diff
@@ -94,8 +92,6 @@ internal class PostHogTest {
                     addBeforeSend(beforeSend)
                 }
                 this.errorTrackingConfig.inAppIncludes.add("com.posthog")
-                this.errorTrackingConfig.exceptionSteps.enabled = exceptionStepsEnabled
-                this.errorTrackingConfig.exceptionSteps.maxBytes = exceptionStepsMaxBytes
```
ioannisj, line 98: https://github.com/PostHog/posthog-android/pull/569/changes#r3459559560

posthog/src/main/java/com/posthog/PostHog.kt
ioannisj, line 681: Now that we have the concrete types, I think it should be possible to also gate the generic capture("$exception") path as well

ioannisj (commented):
Need to rebase on main to remove some unrelated changes (e.g exceptionSteps + tests).   Left a comment, not a blocker since this does not sit directly on RN path/case but the config option would be more complete if we gated generic capture() method as well

turnipdabeets:
Superseded by #608, which carries this branch's commits forward onto current `main` (original authorship preserved), restores the unrelated test deletions flagged in review, and gates the generic `capture("$exception")` path per the remaining review comment. The head branch here lives on a fork without maintainer edits enabled, so the fix-ups couldn't land on this PR directly.

PR #574, opened 2026-06-16
feat(error-tracking): add exception steps

ioannisj:

Ports the [`exception-steps` capability](https://github.com/PostHog/sdk-specs) to the Android/JVM SDK (browser reference: posthog-js#3389; iOS port: posthog-ios#656).

Exception steps are breadcrumb-style context records the host app records over time. The SDK keeps a rolling, byte-bounded buffer of them and attaches a snapshot to **every** captured `$exception` event as `$exception_steps`, giving the error-tracking UI a timeline of activity leading up to each error.

New public API, alongside `captureException`:

```kotlin
PostHog.addExceptionStep("User tapped Checkout", mapOf("screen" to "cart"))
```

**What's included**

- `PostHogExceptionStepsConfig` — `enabled` (default `true`) and `maxBytes` (default `32768`), nested under `errorTrackingConfig.exceptionSteps`.
- `PostHogExceptionStepsBuffer` (internal) — thread-safe FIFO buffer; normalizes each step to its JSON wire form once via the existing event-property serializer, byte-counts in UTF-8, evicts oldest on overflow, rejects an oversized single step, strips reserved keys (`$message`/`$timestamp`), and never throws.
- `addExceptionStep(message, properties?)` on `PostHogInterface` + `PostHog` — `$timestamp` is captured on the caller thread and **recording is synchronous** (no background dispatch), so a step recorded immediately before a crash is already buffered when it is captured. Steps are attached in the shared `capture()` path for any `$exception` (so externally-built exceptions from RN/Flutter bridges get them too), only when the caller didn't already supply `$exception_steps`. The buffer persists across captures and identity changes (`reset()`), and is cleared on `close()`.
- Sample app — an `ErrorTrackingActivity` screen to record steps and capture manual / fatal exceptions.

**Design note — in-memory only.** Android captures fatal crashes in-process via `UncaughtExceptionHandler`, and steps attach to the event *before* it enters the persisted event queue. So, unlike iOS (where PLCrashReporter reconstructs the crash on the next launch), Android needs no crash-durable persistence — an instance-lifetime in-memory buffer is sufficient, which the spec explicitly permits for in-process JVM crash capture.

- **Unit tests** for the buffer: FIFO ordering, byte-budget eviction, oversized-step rejection, UTF-8 byte counting, reserved-key stripping, unrepresentable-value (`NaN`) dropping, `clear`, and concurrent add/snapshot consistency.
- **Integration tests** on `PostHog`: ordered attach, no-overwrite when the caller supplies `$exception_steps`, persistence across captures + `reset()`, disabled no-op, `maxBytes <= 0` no-op, and the generic `capture("$exception", …)` (hybrid-SDK) path.
- **Manual verification on an emulator:** recorded steps via the sample's Exception Steps screen (the counter increments), captured a manual exception, and confirmed the `$exception` event is produced and persisted to the on-disk event queue carrying `$exception_steps`. Fatal-crash attachment was verified manually on-device.
- `spotlessCheck` and `apiCheck` pass.

CleanShot 2026-06-17 at 15 39 03@2x

## :pencil: Checklist

### If releasing new changes

- [x] Added a changeset (`posthog` / `posthog-android` minor).

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — DRI: @ioannisj


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -663,6 +679,8 @@ public class PostHog private constructor(
                 exceptionProperties.putAll(it)
             }
 
+            exceptionStepsBuffer?.attachTo(exceptionProperties)
+
             capture(PostHogEventName.EXCEPTION.event, properties = exceptionProperties)
```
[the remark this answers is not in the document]
ioannisj: Raised here as well https://github.com/PostHog/posthog-ios/pull/656#discussion_r3415879951
Will address this properly soon

posthog/src/main/java/com/posthog/PostHogInterface.kt
```diff
@@ -96,6 +96,32 @@ public interface PostHogInterface : PostHogCoreInterface {
...
+     * Recording is synchronous: the `$timestamp` is captured at call time and the step
+     * is normalized, byte-budget enforced, and buffered before this method returns, so a
+     * step recorded immediately before an exception or crash is present when it is
+     * captured. The work is bounded and cheap, adding negligible latency to the caller.
+     *
+     * @param message a non-empty description of the step
+     * @param properties optional user-supplied properties (the reserved keys
+     * `$message` and `$timestamp` are stripped)
+     */
+    public fun addExceptionStep(
```
hpouillot, line 120: Robot says this will break posthog-server build as it should implement this abstract method ?

hpouillot (approved):

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -671,6 +707,25 @@ public class PostHog private constructor(
...
 
+    override fun addExceptionStep(
+        message: String,
+        properties: Map<String, Any>?,
+    ) {
+        try {
+            if (!isEnabled()) {
+                return
+            }
+            val buffer = exceptionStepsBuffer ?: return
```
turnipdabeets, line 721: Missing opt-out check. `capture()` has `if (config?.optOut == true) return` — `addExceptionStep` needs the same. Without it, steps keep buffering after `optOut()` and leak into post-`optIn` exception events.

Suggested fix:
```kotlin
val buffer = exceptionStepsBuffer ?: return
if (config?.optOut == true) return
```

Also worth clearing the buffer in `optOut()` itself:
```kotlin
config?.optOut = true
getPreferences().setValue(OPT_OUT, true)
exceptionStepsBuffer?.clear()
```

posthog/src/main/java/com/posthog/PostHogInterface.kt
ioannisj: I think robot is wrong since posthog-server implements its own [com.posthog.server.PostHogInterface](https://github.com/PostHog/posthog-android/blob/32e2f62c36f595bcb570818ac18b2bc7ae7f90db/posthog-server/src/main/java/com/posthog/server/PostHogInterface.kt) which is a separate sealed interface, not the core com.posthog.PostHogInterface we are modifying here. Confused because of identical naming probably

posthog/src/main/java/com/posthog/PostHog.kt
ioannisj: yeah, this should be guarded

ioannisj: Done with [32c4396](https://github.com/PostHog/posthog-android/pull/574/commits/32c43961b949a8dc12a29a714f5f208d15ceaf60)

PR #580, opened 2026-06-17
feat: expose getAnonymousId() on PostHogInterface

turnipdabeets:
## Summary

- Adds `getAnonymousId(): String` to `PostHogInterface`
- Implements it in `PostHog.kt` (reads the existing `anonymousId` field backed by SharedPreferences)
- Adds the delegation stub in the inner `NoOpPostHog` companion

The anonymous ID was already stored and managed internally — this just surfaces it through the public API.

## Motivation

iOS SDK already exposes `getAnonymousId()` publicly. Android was the only platform missing it, which blocked the [posthog-kmp](https://github.com/PostHog/posthog-kmp) KMP wrapper from implementing `getAnonymousId()` correctly. Without this, the KMP wrapper falls back to returning `distinctId()`, which is wrong for identified users.

Spotted in review: PostHog/posthog-kmp#1

## Test plan

- [ ] `getAnonymousId()` returns the anonymous ID before `identify()` is called
- [ ] `getAnonymousId()` still returns the original anonymous ID after `identify()` (does not change to the identified distinct ID)
- [ ] `getAnonymousId()` returns `""` when SDK is not enabled

### DIFF
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -383,6 +383,7 @@ public class PostHog private constructor(
         }
     }
 
+    @get:JvmName("getAnonymousIdInternal")
     private var anonymousId: String
         get() {
             var anonymousId: String?
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -1591,6 +1591,13 @@ public class PostHog private constructor(
         return distinctId
     }
 
+    override fun getAnonymousId(): String {
+        if (!isEnabled()) {
+            return ""
+        }
+        return anonymousId
+    }
+
     override fun getDeviceId(): String {
         if (!isEnabled()) {
             return ""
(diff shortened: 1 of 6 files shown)


turnipdabeets:
Addressed @greptile-apps[bot]'s feedback in 5c236a5: updated `@return` to "or an empty string if the SDK is not enabled" — the ID is always auto-generated on first access so "not yet initialized" was misleading.

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -383,6 +383,7 @@ public class PostHog private constructor(
         }
     }
 
+    @get:JvmName("getAnonymousIdInternal")
```
marandaneto, line 386: we dont need that i think

marandaneto:
you need to run the make api cmd

marandaneto (approved):

posthog/src/main/java/com/posthog/PostHog.kt
turnipdabeets: Tried removing it — compiler throws "Platform declaration clash: The following declarations have the same JVM signature (getAnonymousId()Ljava/lang/String;)" because Kotlin auto-generates a JVM getter named `getAnonymousId()` for the private `anonymousId` property. The `@get:JvmName` renames that private getter so it doesn't collide with the public override. Happy to rename the backing property instead if you prefer, but something here is needed.

PR #585, opened 2026-06-24
fix: Clear feature flag called cache on close

marandaneto:

`PostHogStateless.close()` shut down the queue and feature flag provider but left the in-memory feature flag called cache populated. If the same SDK instance is closed and set up again, stale cache entries can suppress future `$feature_flag_called` events for values that were only seen before shutdown.

This PR clears the feature flag called cache during close and adds regression coverage for that behavior.

- `make stop && ./gradlew :posthog:test --tests "com.posthog.PostHogStatelessTest.close clears feature flag called cache"`
- `make checkFormat`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

An OpenAI API coding agent prepared this PR using the existing dedicated `fix/ff-called-cache-shutdown` worktree. The branch was fast-forwarded to the latest `main` before committing because it was behind the base branch.

The implementation is intentionally narrow: clear the existing feature flag called cache during SDK close, add a regression test, and add a patch changeset for the affected release packages.

### DIFF
--- a/.changeset/clear-flag-cache-close.md
+++ b/.changeset/clear-flag-cache-close.md
@@ -1,8 +1,6 @@
 ---
 'posthog': patch
-'posthog-android': patch
 'posthog-server': patch
-'posthog-android-surveys-compose': patch
 ---
 
 Clear the feature flag called cache when closing the SDK.
--- a/.changeset/clear-flag-cache-close.md
+++ b/.changeset/clear-flag-cache-close.md
@@ -1,0 +1,8 @@
+---
+'posthog': patch
+'posthog-android': patch
+'posthog-server': patch
+'posthog-android-surveys-compose': patch
+---
+
+Clear the feature flag called cache when closing the SDK.
--- a/posthog/src/test/java/com/posthog/PostHogStatelessTest.kt
+++ b/posthog/src/test/java/com/posthog/PostHogStatelessTest.kt
@@ -1,5 +1,6 @@
 package com.posthog
 
+import com.posthog.internal.PostHogFeatureFlagCalledCache
 import com.posthog.internal.PostHogFeatureFlagsInterface
 import com.posthog.internal.PostHogMemoryPreferences
 import com.posthog.internal.PostHogPreferences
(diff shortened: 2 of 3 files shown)


.changeset/clear-flag-cache-close.md
```diff
@@ -0,0 +1,8 @@
+---
+'posthog': patch
+'posthog-android': patch
+'posthog-server': patch
+'posthog-android-surveys-compose': patch
```
marandaneto: we dont need `posthog-android-surveys-compose` here

.changeset/clear-flag-cache-close.md
```diff
@@ -0,0 +1,8 @@
+---
+'posthog': patch
+'posthog-android': patch
```
marandaneto: we dont need android here, we override this method

posthog/src/test/java/com/posthog/PostHogStatelessTest.kt
```diff
@@ -1019,6 +1026,27 @@ internal class PostHogStatelessTest {
...
+        sut.setMockQueue(mockQueue)
+        sut.setMockFeatureFlags(mockFeatureFlags)
+
+        sut.getFeatureFlagStateless("user123", "test_flag")
+        assertEquals(1, sut.featureFlagsCalledCacheSize())
+
+        sut.close()
+
+        assertEquals(0, sut.featureFlagsCalledCacheSize())
+    }
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in c4d785b — updated the stateless close test to exercise a close followed by setup again, then verify the same feature flag call emits a new `$feature_flag_called` event while still asserting the cache is cleared on close.

.changeset/clear-flag-cache-close.md
marandaneto: Fixed in 5984948 — removed `posthog-android` from the changeset; it now only bumps the packages affected by the stateless close change.

.changeset/clear-flag-cache-close.md
marandaneto: Fixed in 5984948 — removed `posthog-android-surveys-compose` from the changeset.

dustinbyrne (approved):

PR #587, opened 2026-06-25
fix(flags): retry flag requests on transient network errors

marandaneto:

`/flags/?v=2` evaluation should tolerate transient network failures, but should not retry HTTP/API responses such as 408, 429, or 5xx.

This PR adds bounded retry/backoff only for transient network failures (timeouts, connection resets/lost connections, and EOF-style failures) and keeps HTTP/API status errors terminal.

See the repo-specific command below.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented by pi after a maintainer-directed cross-SDK pass. The chosen policy is transient network-only retry/backoff for `/flags/?v=2`; HTTP/API statuses remain non-retryable by design. Connection-refused failures also fail fast where the platform exposes that distinction.

Tested with:

```bash
./gradlew :posthog:test --tests com.posthog.internal.PostHogApiTest --no-daemon
```

### DIFF
--- a/.changeset/quiet-flags-retry.md
+++ b/.changeset/quiet-flags-retry.md
@@ -2,4 +2,4 @@
 'posthog': patch
 ---
 
-Retry feature flag requests after network errors only.
+Retry feature flag requests after transient network errors only.
--- a/.changeset/quiet-flags-retry.md
+++ b/.changeset/quiet-flags-retry.md
@@ -1,0 +1,5 @@
+---
+'posthog': patch
+---
+
+Retry feature flag requests after network errors only.
--- a/.changeset/quiet-flags-retry.md
+++ b/.changeset/quiet-flags-retry.md
@@ -1,5 +1,7 @@
 ---
 'posthog': patch
+'posthog-android': patch
+'posthog-server': patch
 ---
 
 Retry feature flag requests after transient network errors only. The feature flag request retry count defaults to 1 and can be set to 0 to disable retries.
--- a/.changeset/quiet-flags-retry.md
+++ b/.changeset/quiet-flags-retry.md
@@ -2,4 +2,4 @@
 'posthog': patch
 ---
 
-Retry feature flag requests after transient network errors only.
+Retry feature flag requests after transient network errors only. The feature flag request retry count defaults to 1 and can be set to 0 to disable retries.
--- a/posthog/src/main/java/com/posthog/internal/PostHogApi.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogApi.kt
@@ -209,7 +209,7 @@ public class PostHogApi(
 
     @Throws(PostHogApiError::class, IOException::class)
     private fun executeFlagsWithRetry(request: Request): PostHogFlagsResponse? {
-        val maxRetries = config.maxRetries.coerceAtLeast(0)
+        val maxRetries = config.featureFlagRequestMaxRetries.coerceAtLeast(0)
         var retryAttempt = 0
 
         while (true) {
--- a/posthog/src/main/java/com/posthog/internal/PostHogApi.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogApi.kt
@@ -32,7 +32,7 @@ public class PostHogApi(
 ) {
     private companion object {
         private const val APP_JSON_UTF_8 = "application/json; charset=utf-8"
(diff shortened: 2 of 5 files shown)


posthog/src/test/java/com/posthog/internal/PostHogApiTest.kt
```diff
@@ -133,6 +144,58 @@ internal class PostHogApiTest {
...
+                        sut.flags("distinctId", anonymousId = "anonId", groups = emptyMap())
+                    }
+
+                assertEquals(statusCode, exc.statusCode)
+                assertEquals(1, http.requestCount)
+            } finally {
+                http.shutdown()
+            }
+        }
+    }
```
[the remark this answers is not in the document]
marandaneto: Addressed in f322de9: added `flags rethrows retryable IOException after max retry attempts`, which exercises a retryable SocketException through the configured retry budget and asserts the IOException is rethrown after 2 attempts.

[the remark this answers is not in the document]
marandaneto: Addressed in f322de9: replaced the `forEach` HTTP error coverage with `PostHogApiFlagsHttpErrorTest`, a JUnit parameterized test that runs each status code as its own case.

posthog/src/main/java/com/posthog/internal/PostHogApi.kt
```diff
@@ -193,9 +201,33 @@ public class PostHogApi(
                 config.serializer.serialize(flagsRequest, it.bufferedWriter())
             }
 
+        return executeFlagsWithRetry(request)
+    }
+
+    @Throws(PostHogApiError::class, IOException::class)
+    private fun executeFlagsWithRetry(request: Request): PostHogFlagsResponse? {
+        val maxRetries = config.maxRetries.coerceAtLeast(0)
```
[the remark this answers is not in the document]
marandaneto: Addressed in cb32305: flags retries now use separate `PostHogConfig.featureFlagRequestMaxRetries` KDoc/configuration instead of reusing event flush `maxRetries`.

.changeset/quiet-flags-retry.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog': patch
```
marandaneto, line 2: we should add posthog-android and posthog-server if affected as well

marandaneto: Addressed in ef71a36: added both affected downstream packages, `posthog-android` and `posthog-server`, to the changeset alongside `posthog`.

turnipdabeets (approved):

posthog/src/main/java/com/posthog/internal/PostHogApi.kt
```diff
@@ -46,6 +51,12 @@ public class PostHogApi(
             .addInterceptor(GzipRequestInterceptor(config))
             .build()
 
+    private val flagsClient: OkHttpClient by lazy {
```
marandaneto, line 54: we'll also need the config.proxy and GzipRequestInterceptor(config) as `client`

posthog/src/main/java/com/posthog/internal/PostHogApi.kt
```diff
@@ -29,6 +32,8 @@ public class PostHogApi(
 ) {
     private companion object {
         private const val APP_JSON_UTF_8 = "application/json; charset=utf-8"
+        private const val FLAGS_INITIAL_RETRY_DELAY_MS = 1_000L
```
marandaneto: all sdks start with 300ms delay then 600ms, always double down

posthog/src/main/java/com/posthog/internal/PostHogApi.kt
marandaneto: Already covered in the current implementation: `flagsClient` is built from `client.newBuilder()`, and `client` includes both `config.proxy` and `GzipRequestInterceptor(config)`, so the flags client inherits those settings while only disabling OkHttp connection retry.

posthog/src/main/java/com/posthog/internal/PostHogApi.kt
marandaneto: Addressed in 88101b2: changed the feature-flag retry backoff to start at 300ms; subsequent retries continue to double from there.

PR #590, opened 2026-06-26
fix: buffer session replay snapshots until first remote config resolves

ioannisj:

Session replay starts at cold start from the **disk-cached** `sessionRecording` config, set before any network call. Remote-config resolution is async, so when the cached decision disagrees with the freshly-fetched config (cached-on, fresh-off) there is a stale-cache window where the SDK records — and uploads — a session the fresh config disallows. Fresh installs are already safe, so this is a returning-user bug. There was also a latent gap where a mid-session fresh "off" only took effect at the next session rotation.

This is the **Android slice** of a multi-target session-replay fix (iOS counterpart: PostHog/posthog-ios#679). Tracking issue: PostHog/posthog-android#575.

Recording still starts optimistically off the cached value, but snapshots are **buffered (not persisted, not sent) until the first live remote config resolves**; on resolve the buffered window migrates to the replay queue only if the current session is recordable under the fresh config, and is dropped otherwise. No stale-cache `$snapshot` reaches the persisted queue or the network.

### What changed

**Buffer-and-decide gate** (`PostHogReplayIntegration`)
- An `awaitingFirstRemoteConfig` gate (guarded by `bufferingLock`) routes captured snapshots into the in-memory buffer until the first `/config` attempt resolves, then disarms exactly once. The gate is only armed when a remote-config fetch will actually happen — if neither `remoteConfig` nor `preloadFeatureFlags` is enabled, nothing would ever resolve it, so it stays disarmed and snapshots persist normally.
- The minimum-duration migrate trigger is suppressed while the gate is armed, so a cached `minimumDuration` can't drain stale-cache frames to the persisted queue before the fresh decision lands. A snapshot that reaches the persist boundary while replay is inactive is dropped rather than enqueued.

**Resolve uses the full recordability decision, not just the flag**
- On resolve, the buffered window migrates only when the session is recordable under the fresh config — recording flag on **and** sampled in for the current session id **and** not waiting on an event trigger. A sampled-out or trigger-gated session drops the window instead of uploading it.
- Migration is handed to the existing minimum-duration gate rather than force-flushed: a short opening window keeps buffering until it spans `minimumDuration`.
- A not-recordable resolve runs **stop → clear buffer → disarm** in that order, so an in-flight `add()` racing the disarm finds the buffer cleared and replay inactive and drops its snapshot.

**linkedFlag handling — single-callback on Android**
- Unlike iOS (which gets a separate `/flags` callback after `/config` and *defers* the resolve to it), Android's `loadRemoteConfig` loads `/flags` **nested and synchronously** (`notifyRemoteConfigLoaded = false`) and fires exactly **one** `onRemoteConfig`. The linked flag is already merged and fresh at that single callback, so Android resolves the buffer there directly. There is no second callback to defer to, and deferring would leave the gate armed for the whole session (silently buffering, never sending). This is the one place the Android realization intentionally diverges from the iOS deferral.

**Mid-session re-evaluation**
- After the gate disarms, `onRemoteConfig` re-evaluates recording: a subsequent config that turns recording off stops it promptly instead of waiting for session rotation; one that turns it back on resumes recording for the current session. `stopIfActive` flips the active flag synchronously before posting `stop()`, so a concurrent `add()` can't persist a snapshot after the off decision.

**Concurrency / reset**
- A `bufferingGeneration` counter lets an in-flight resolve detect a concurrent session rotation (`reset()` / identity change re-arming and clearing the buffer) and no-op instead of migrating or clearing the wrong session's window.
- The gate re-resolves from the cached config after `reset()` on the next `/flags` reload, so the post-reset opening window isn't dropped.

**Failure / quota paths** (`PostHogRemoteConfig`)
- A failed first `/config` (offline launch) falls back to the disk-cached decision and emits a single fallback log line.
- A quota-limited `/flags` response resolves via the not-recordable path (a live response, not a fetch failure) and re-arms the session-recording / error-tracking / capture-performance decisions from the cached config instead of losing them.
- Tracks whether the live config resolved for the current identity via `hasRemoteConfigFetched()`, reset on `clear()`.

The in-memory buffer is intentionally **not** size-capped — replay needs the complete opening snapshot sequence, so a frame-dropping cap would migrate an unplayable partial window. A slow or absent `/config` can grow it until the config resolves; it is bounded by session activity and self-heals on the next resolution.

Unit tests (`./gradlew :posthog:test :posthog-android:testDebugUnitTest :posthog-android:testReleaseUnitTest spotlessCheck apiCheck` — all green), covering the gate lifecycle in `PostHogReplayIntegrationTest` and the remote-config paths in `PostHogRemoteConfigTest` / `PostHogTest`:

- while awaiting the first config, `add()` routes snapshots to the buffer, not the persisted queue;
- a **single** `onRemoteConfig` delivery resolves the gate — flag-on migrates the buffered window, and a snapshot added afterwards persists directly (no second callback needed);
- flag-off clears the buffer and stops recording, leaving already-persisted events untouched (clearBuffer scoping);
- a minimum-duration-elapsed event while awaiting the first config does **not** migrate (suppression while armed);
- mid-session flag-off stops recording synchronously — a snapshot added before the posted `stop()` runs is not persisted;
- a subsequent config resumes recording once the config is already fetched;
- the gate does not arm when neither `remoteConfig` nor `preloadFeatureFlags` will fetch;
- `hasRemoteConfigFetched()` flips true on both the `/config` and `/flags` paths and resets on `clear()`; a quota-limited flags reload resolves and notifies.

Manual runs on the sample app against a local mock server that flips the configs. No manual on-device testing was performed.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -1870,8 +1897,76 @@ public class PostHogReplayIntegration(
...
+        if (config.remoteConfigHolder?.isSessionReplayFlagActive() == true) {
+            replayQueue?.let { migrateBufferToQueueOnBackgroundThread(it) }
+        } else {
+            // Self-gate the capturer first so no new snapshot re-enters the buffer, then drop it.
+            stop()
+            replayQueue?.clearBuffer()
+        }
+
+        synchronized(bufferingLock) { awaitingFirstRemoteConfig = false }
+    }
```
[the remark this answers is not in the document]
ioannisj: Valid but mitigated.  onRemoteConfig() only fires from onRemoteConfigLoaded.loaded(), which is invoked from loadRemoteConfig and executeFeatureFlags, both running on the single remote-config executor. So the calls are serialized

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -1870,8 +1908,78 @@ public class PostHogReplayIntegration(
...
+        if (isSessionReplayActive) {
+            config.logger.log("[Session Replay] $reason")
+            mainHandler.handler.post { if (isSessionReplayActive) stop() }
+        }
+    }
+
+    // Whether to buffer until the live remote config resolves: only when a remote config holder
+    // exists and hasn't resolved yet. Remote config is always fetched at setup, so this disarms on
+    // the first onRemoteConfig (and re-arms after clear() until the post-reset reload resolves it).
+    private fun shouldAwaitFirstRemoteConfig(): Boolean = config.remoteConfigHolder?.let { !it.hasRemoteConfigFetched() } ?: false
```
turnipdabeets: If neither `remoteConfig` nor `preloadFeatureFlags` is enabled, `onRemoteConfig` never fires and this stays armed — snapshots buffer indefinitely. Worth a guard, or is replay just unsupported in that config?

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -1870,8 +1908,78 @@ public class PostHogReplayIntegration(
...
+
+        val currentSessionId = postHog.getSessionId()?.toString() ?: return
+        if (!remoteConfig.makeSamplingDecision(currentSessionId)) {
+            stopIfActive("Remote config sampled this session out. Stopping.")
+            return
+        }
+
+        if (!isSessionReplayActive) {
+            config.logger.log("[Session Replay] Remote config enabled recording. Resuming.")
+            mainHandler.handler.post { if (!isSessionReplayActive) start(resumeCurrent = true) }
```
turnipdabeets: `resumeCurrent = true` skips `clearSnapshotStates`, so the resumed segment may not start with a full snapshot. Is the keyframe still emitted here?

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
ioannisj: Fixed in b202c07a shouldAwaitFirstRemoteConfig() no longer arms the gate when no fetch will happen)

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
ioannisj: Fixed. `resumeCurrent = true` now should clear snapshot state and force a redraw so the segment starts with a fresh meta + full snapshot, without rotating the session or re-arming the cold-start buffer

posthog/src/main/java/com/posthog/PostHogIntegration.kt
```diff
@@ -25,4 +25,13 @@ public interface PostHogIntegration {
...
     public fun onRemoteConfig() {
     }
+
+    /**
+     * Called when the first remote config attempt finished without a live response
+     * (e.g. the device is offline, or the request failed). No fresh config was applied, so
+     * integrations that buffered work while awaiting the live config should fall back to their
+     * cached state instead of waiting indefinitely.
+     */
+    public fun onRemoteConfigFailed() {
```
marandaneto: or do we add a param with a default value to `onRemoteConfig`?

marandaneto:
We have the same issue with error tracking, surveys, etc, everything that depends on remote config, right?
we should open new issues for those if not fixing right away

marandaneto (approved):

ioannisj:
> We have the same issue with error tracking, surveys, etc, everything that depends on remote config, right? we should open new issues for those if not fixing right away

Yeah I'll do a rundown later today on this and open some issues if needed

posthog/src/main/java/com/posthog/PostHogIntegration.kt
ioannisj: folded with onRemoteConfig(loaded: Boolean = true)

PR #592, opened 2026-06-29
feat: add requestHeaders config option

turnipdabeets:

Part of the cross-SDK work for PostHog/posthog-js#2132 (no way to include `Authorization` headers in reverse-proxy setups). React Native session replay and native crash uploads are sent directly by this native SDK, so they bypass any JS-layer header support. This adds first-class custom-header support to the Android SDK so those (and all other) requests can carry an `Authorization` header for an authenticated reverse proxy.

- Added tests in `PostHogApiTest` asserting the header reaches `/batch` and `/flags`, and is absent when unset (uses the existing MockWebServer harness).
- `./gradlew :posthog:test --tests "com.posthog.internal.PostHogApiTest"` passes.
- `./gradlew spotlessCheck` passes.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Directed by @ioannisj / @marandaneto while wiring `requestHeaders` through for React Native — the native SDK must honor the header for replay/crash uploads. Applied the headers via a small `CustomHeadersInterceptor` added to the OkHttp client (covers every endpoint, including the derived `flagsClient`, and no-ops when no headers are set), keeping the SDK's existing gzip interceptor and default client setup intact. Tool used: Claude Code.

---

Related: PostHog/posthog-js#2132

### DIFF
--- a/.changeset/request-headers.md
+++ b/.changeset/request-headers.md
@@ -1,6 +1,6 @@
 ---
-'posthog': patch
-'posthog-android': patch
+'posthog': minor
+'posthog-android': minor
 ---
 
 Add `requestHeaders` config option to send custom headers (e.g. `Authorization`) with every request to the PostHog API. Useful for reverse-proxy setups that require authentication.
--- a/.changeset/request-headers.md
+++ b/.changeset/request-headers.md
@@ -1,0 +1,6 @@
+---
+'posthog': patch
+'posthog-android': patch
+---
+
+Add `requestHeaders` config option to send custom headers (e.g. `Authorization`) with every request to the PostHog API. Useful for reverse-proxy setups that require authentication.
--- a/posthog/src/main/java/com/posthog/PostHogConfig.kt
+++ b/posthog/src/main/java/com/posthog/PostHogConfig.kt
@@ -355,6 +355,15 @@ public open class PostHogConfig(
     public var httpClient: OkHttpClient? = null
 
     /**
+     * Custom headers to send with every request to the PostHog API.
+     *
+     * Useful for reverse-proxy setups that require authentication, e.g. an `Authorization` header.
+     *
+     * Default: empty (no extra headers).
+     */
+    public var requestHeaders: Map<String, String> = emptyMap()
+
+    /**
      * The PostHog project API key, trimmed of leading and trailing whitespace.
      */
     public val apiKey: String = apiKey.trim()
--- a/posthog/src/main/java/com/posthog/PostHogConfig.kt
+++ b/posthog/src/main/java/com/posthog/PostHogConfig.kt
@@ -361,6 +361,7 @@ public open class PostHogConfig(
      *
      * Default: empty (no extra headers).
      */
+    @Volatile
     public var requestHeaders: Map<String, String> = emptyMap()
 
     /**
--- a/posthog/src/main/java/com/posthog/PostHogConfig.kt
+++ b/posthog/src/main/java/com/posthog/PostHogConfig.kt
@@ -359,6 +359,8 @@ public open class PostHogConfig(
      *
      * Useful for reverse-proxy setups that require authentication, e.g. an `Authorization` header.
      *
+     * Read once when the SDK is set up; changes after setup are ignored.
+     *
      * Default: empty (no extra headers).
      */
     @Volatile
--- a/posthog/src/main/java/com/posthog/internal/CustomHeadersInterceptor.kt
+++ b/posthog/src/main/java/com/posthog/internal/CustomHeadersInterceptor.kt
@@ -2,6 +2,7 @@ package com.posthog.internal
 
 import com.posthog.PostHogConfig
 import com.posthog.PostHogInternal
+import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import okhttp3.Interceptor
 import okhttp3.Response
 import java.io.IOException
@@ -13,20 +14,26 @@ import java.io.IOException
  */
 @PostHogInternal
 public class CustomHeadersInterceptor(private val config: PostHogConfig) : Interceptor {
+    private val configuredHost: String? = config.host.toHttpUrlOrNull()?.host
+
     @Throws(IOException::class)
     override fun intercept(chain: Interceptor.Chain): Response {
         val requestHeaders = config.requestHeaders
-        if (requestHeaders.isEmpty()) {
-            return chain.proceed(chain.request())
+        val request = chain.request()
+        // Only attach to the configured host so headers aren't sent to rewritten hosts (e.g. the CDN).
+        if (requestHeaders.isEmpty() || request.url.host != configuredHost) {
+            return chain.proceed(request)
         }
 
-        val request = chain.request()
         val builder = request.newBuilder()
         for ((key, value) in requestHeaders) {
-            // Keep headers the SDK already set on the request (e.g. localEvaluation's
-            // Authorization personal API key); those take precedence over config values.
-            if (request.header(key) == null) {
+            // SDK-set request headers (e.g. localEvaluation's Authorization) take precedence.
+            if (request.header(key) != null) continue
+            try {
                 builder.header(key, value)
+            } catch (e: IllegalArgumentException) {
+                // Drop headers okhttp rejects rather than failing the request.
+                config.logger.log("Dropping invalid request header '$key': ${e.message}")
             }
         }
         return chain.proceed(builder.build())
--- a/posthog/src/main/java/com/posthog/internal/CustomHeadersInterceptor.kt
+++ b/posthog/src/main/java/com/posthog/internal/CustomHeadersInterceptor.kt
@@ -4,6 +4,7 @@ import com.posthog.PostHogConfig
 import com.posthog.PostHogInternal
 import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import okhttp3.Interceptor
+import okhttp3.Request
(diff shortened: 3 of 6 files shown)


posthog/src/main/java/com/posthog/internal/CustomHeadersInterceptor.kt
```diff
@@ -0,0 +1,29 @@
...
+    @Throws(IOException::class)
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val requestHeaders = config.requestHeaders
+        if (requestHeaders.isEmpty()) {
+            return chain.proceed(chain.request())
+        }
+
+        val builder = chain.request().newBuilder()
+        for ((key, value) in requestHeaders) {
+            builder.header(key, value)
```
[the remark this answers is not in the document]
turnipdabeets: Good catch — real bug. Fixed in 1b980ac: the interceptor now skips any header key already present on the request, so SDK-set request headers (like `localEvaluation`'s `Authorization` personal API key) take precedence over `requestHeaders`. Added a regression test asserting `localEvaluation` still authenticates with the personal API key even when a custom `Authorization` is configured.

posthog/src/test/java/com/posthog/internal/PostHogApiTest.kt
```diff
@@ -88,6 +92,46 @@ internal class PostHogApiTest {
...
+        val http = mockHttp()
+        val url = http.url("/")
+
+        val sut = getSut(host = url.toString())
+
+        sut.batch(listOf(generateEvent()))
+
+        val request = http.takeRequest()
+        assertNull(request.headers["Authorization"])
+    }
```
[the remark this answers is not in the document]
turnipdabeets: This class predominantly uses plain `@Test` methods (the `Parameterized` import is only used for one specific case), so I kept these as three small, explicitly-named tests to match the file's dominant style and keep failures easy to read. Happy to parameterize if the team would prefer it as the convention here.

turnipdabeets:
Unblocks the RN plugin bridge: PostHog/posthog-js#4018 sets `config.requestHeaders` and can only merge once this lands + releases.

posthog/src/main/java/com/posthog/internal/CustomHeadersInterceptor.kt
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        val request = chain.request()
+        val builder = request.newBuilder()
+        for ((key, value) in requestHeaders) {
+            // Keep headers the SDK already set on the request (e.g. localEvaluation's
+            // Authorization personal API key); those take precedence over config values.
+            if (request.header(key) == null) {
+                builder.header(key, value)
+            }
```
marandaneto: `request.header(key) ?: builder.header(key, value)` style is nicer imo

posthog/src/main/java/com/posthog/internal/PostHogApi.kt
```diff
@@ -46,9 +46,15 @@ public class PostHogApi(
     }
 
     private val client: OkHttpClient =
-        config.httpClient ?: OkHttpClient.Builder()
-            .proxy(config.proxy)
-            .addInterceptor(GzipRequestInterceptor(config))
+        (
+            config.httpClient ?: OkHttpClient.Builder()
+                .proxy(config.proxy)
+                .addInterceptor(GzipRequestInterceptor(config))
```
marandaneto:
```suggestion
                .addInterceptor(GzipRequestInterceptor(config))
                .addInterceptor(CustomHeadersInterceptor(config))
```

you dont need `newBuilder` again

posthog/src/main/java/com/posthog/PostHogConfig.kt
```diff
@@ -354,6 +354,15 @@ public open class PostHogConfig(
...
     public var httpClient: OkHttpClient? = null
 
+    /**
+     * Custom headers to send with every request to the PostHog API.
+     *
+     * Useful for reverse-proxy setups that require authentication, e.g. an `Authorization` header.
+     *
+     * Default: empty (no extra headers).
+     */
+    public var requestHeaders: Map<String, String> = emptyMap()
```
marandaneto, line 367: is the goal to swap the instance or append to the list? see eg beforeSend
this would need to be `MutableList`, and also to make it thread safe

posthog/src/main/java/com/posthog/internal/CustomHeadersInterceptor.kt
```diff
@@ -0,0 +1,34 @@
...
+            return chain.proceed(chain.request())
+        }
+
+        val request = chain.request()
+        val builder = request.newBuilder()
+        for ((key, value) in requestHeaders) {
+            // Keep headers the SDK already set on the request (e.g. localEvaluation's
+            // Authorization personal API key); those take precedence over config values.
+            if (request.header(key) == null) {
+                builder.header(key, value)
```
ioannisj: could invalid values throw here?

.changeset/request-headers.md
```diff
@@ -0,0 +1,6 @@
+---
+'posthog': patch
+'posthog-android': patch
```
ioannisj: I would do a minor?

posthog/src/main/java/com/posthog/internal/PostHogApi.kt
```diff
@@ -46,9 +46,15 @@ public class PostHogApi(
...
+                .build()
+        )
+            // Applies config.requestHeaders to every request; no-ops when none are set.
```
ioannisj: Not sure but should SDK calls to static/cdn config strip these extra headers? Question applies to iOS implementation as well.

posthog/src/main/java/com/posthog/internal/PostHogApi.kt
turnipdabeets: Done

posthog/src/main/java/com/posthog/internal/CustomHeadersInterceptor.kt
turnipdabeets: yes, using try/catch now

posthog/src/main/java/com/posthog/internal/PostHogApi.kt
turnipdabeets: Good call — implemented on both. Custom headers now only attach when request.host == config.host (Android: in the interceptor; iOS: in applyCustomHeaders), so the rewritten static-config CDN host is skipped. c113dcb

posthog/src/main/java/com/posthog/PostHogConfig.kt
turnipdabeets: Good catch on thread-safety

Kept it a Map rather than a MutableList though: unlike beforeSend (a list of independent hooks you append/remove), requestHeaders is a single header set you replace — so swap-semantics fits better than add/remove. Happy to add add/remove helpers instead if you'd prefer.

posthog/src/main/java/com/posthog/internal/CustomHeadersInterceptor.kt
```diff
@@ -0,0 +1,50 @@
...
+    // Snapshot a copy at construction so config is treated as immutable after setup.
+    private val requestHeaders: Map<String, String> = config.requestHeaders.toMap()
+
+    @Throws(IOException::class)
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val request = chain.request()
+        // Only attach to the configured host so headers aren't sent to rewritten hosts (e.g. the CDN).
+        if (requestHeaders.isEmpty() || request.url.host != configuredHost) {
+            return chain.proceed(request)
+        }
```
marandaneto, line 34: will that ever happen? dont we add CustomHeadersInterceptor only for our own http client? which means all requests are gonna match request.url.host

posthog/src/test/java/com/posthog/internal/PostHogApiTest.kt
```diff
@@ -47,10 +48,14 @@ internal class PostHogApiTest {
         httpClient: OkHttpClient? = null,
         maxRetries: Int? = null,
         featureFlagRequestMaxRetries: Int? = null,
+        requestHeaders: Map<String, String>? = null,
     ): PostHogApi {
         val config = PostHogConfig(API_KEY, host)
         config.proxy = proxy
         config.debug = debug
+        if (requestHeaders != null) {
```
marandaneto: worth checking if its not empty either

marandaneto (approved):
left a few comments but unblocking

posthog/src/test/java/com/posthog/internal/PostHogApiTest.kt
turnipdabeets: done

posthog/src/main/java/com/posthog/internal/CustomHeadersInterceptor.kt
turnipdabeets: we only add it to our own default client (a user-supplied httpClient never gets it). But "our own client" doesn't mean "one host": remoteConfig() calls our client with the host rewritten to the assets CDN (us-assets/eu-assets) on the default hosts, so that request's host != config.host and the guard skips it — the CDN-stripping ioannisj wanted. And as a network interceptor it re-runs per redirect hop, so a cross-host 3xx is skipped too. For a custom proxy host everything stays on that host so it's usually a no-op, but that's what keeps a custom Authorization off the CDN / redirect targets.

PR #600, opened 2026-07-01
fix: stop duplicating distinct_id in flags person properties

marandaneto:
`/flags` requests already send `distinct_id` at the top level. Automatically copying the same value into `person_properties.distinct_id` sends duplicated identity data and differs from the desired payload shape.

This PR keeps the top-level `distinct_id` and stops automatically injecting it into `person_properties`. Caller-provided person properties are still forwarded.

- `./gradlew :posthog:test --tests com.posthog.internal.PostHogFlagsRequestTest --tests com.posthog.PostHogTest`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

This was implemented with pi in dedicated git worktrees. I changed the /flags request construction so SDKs keep the top-level distinct_id while no longer auto-copying it into person_properties; explicit caller-provided person_properties['distinct_id'] values are still preserved where covered.

### DIFF
--- a/.changeset/quiet-wolves-fix.md
+++ b/.changeset/quiet-wolves-fix.md
@@ -1,0 +1,5 @@
+---
+"posthog": patch
+---
+
+Stop duplicating `distinct_id` inside `/flags` person properties.
--- a/.changeset/quiet-wolves-fix.md
+++ b/.changeset/quiet-wolves-fix.md
@@ -1,5 +1,7 @@
 ---
 "posthog": patch
+"posthog-android": patch
+"posthog-server": patch
 ---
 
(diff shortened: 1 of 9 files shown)


.changeset/quiet-wolves-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog": patch
```
marandaneto, line 2: also add for android and server packages if affected

marandaneto: Fixed in cd515a6 — Added the Android and server packages to the changeset.

ioannisj (approved):

Issue #596, opened 2026-06-30
Some screens appear to be dropped in session replay recordings


ioannisj:
## Description

Some screens appear to be dropped in session replay recordings on Android. Certain screens that the user navigates to during a session do not appear in the resulting replay, leaving gaps in the recorded flow.

Ticket: https://posthoghelp.zendesk.com/agent/tickets/60852 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/63955)

PR #605, opened 2026-07-06
fix: session replay dropping screens inside Throttler window

tsushanth:
Helps with #596

## Problem

When a user navigates to a new screen during the throttle window, the screen is silently dropped from the session replay recording.

`Throttler.throttle()` schedules a `postDelayed` for the first draw that arrives after a snapshot and sets `isThrottling = true`. Any subsequent draws while `isThrottling` is true hit the `else` branch and are **silently discarded** — no new `postDelayed` is queued. If the user navigates A → B → C within a single throttle window:

1. B's first draw: `isThrottling` is `false` → `postDelayed` scheduled, `isThrottling = true`
2. C's draws: `isThrottling` is `true` → all silently dropped
3. `postDelayed` fires → snapshots whatever is on screen now (C) → produces an A→C diff

Screen B never appears in the recording.

## Fix

Add a `hasPendingDraw` flag. When a draw is discarded (`isThrottling == true`), set the flag. After the delayed snapshot fires, check the flag and call `throttle()` once more so the latest view tree is always captured.

```kotlin
@Volatile private var hasPendingDraw = false

// in postDelayed lambda:
val pendingAfter = hasPendingDraw
hasPendingDraw = false
executeAndReleaseThrottle(runnable)
if (pendingAfter) { throttle(runnable) }

// in else branch (was: silently dropped):
hasPendingDraw = true
```

1 file changed.

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/internal/Throttler.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/internal/Throttler.kt
@@ -17,7 +17,7 @@ internal class Throttler(
     // Set to true when a draw arrives while a postDelayed is already in flight.
     // The pending draw is re-captured once the delayed snapshot fires, ensuring
     // that a screen change landing inside the throttle window is not silently lost.
-    @Volatile private var hasPendingDraw = false
+    private val hasPendingDraw = AtomicBoolean(false)
 
     /**
      * Throttles the given [runnable] by delaying its execution until [delayNs] has passed since the last call.
@@ -37,8 +37,7 @@ internal class Throttler(
                 // Calculate remaining time needed to wait
                 val remainingDelayMs = TimeUnit.NANOSECONDS.toMillis(delayNs - timeSinceLastExecution)
                 mainHandler.handler.postDelayed({
-                    val pendingAfter = hasPendingDraw
-                    hasPendingDraw = false
+                    val pendingAfter = hasPendingDraw.getAndSet(false)
                     executeAndReleaseThrottle(runnable)
                     // A draw arrived while we were waiting; take one more snapshot so
                     // no screen change that occurred inside the throttle window is lost.
@@ -49,7 +48,7 @@ internal class Throttler(
             } else {
                 // A draw arrived while a postDelayed is already scheduled.
                 // Mark it so the delayed lambda re-captures the current view tree after it fires.
-                hasPendingDraw = true
+                hasPendingDraw.set(true)
             }
         }
     }
--- a/posthog-android/src/main/java/com/posthog/android/replay/internal/Throttler.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/internal/Throttler.kt
@@ -14,6 +14,11 @@ internal class Throttler(
     private val delayNs = TimeUnit.MILLISECONDS.toNanos(throttleDelayMs)
     private val isThrottling = AtomicBoolean(false)
 
+    // Set to true when a draw arrives while a postDelayed is already in flight.
+    // The pending draw is re-captured once the delayed snapshot fires, ensuring
+    // that a screen change landing inside the throttle window is not silently lost.
+    @Volatile private var hasPendingDraw = false
+
     /**
      * Throttles the given [runnable] by delaying its execution until [delayNs] has passed since the last call.
      */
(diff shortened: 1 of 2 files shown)


ioannisj (commented):
Let's add a changeset for this. Greptile comments look reasonable to me. Also, I'm not sure if this fixes #596 since completely but maybe partially addresses it.

posthog-android/src/main/java/com/posthog/android/replay/internal/Throttler.kt
```diff
@@ -28,13 +33,23 @@ internal class Throttler(
...
+                    // A draw arrived while we were waiting; take one more snapshot so
+                    // no screen change that occurred inside the throttle window is lost.
+                    if (pendingAfter) {
+                        throttle(runnable)
+                    }
                 }, remainingDelayMs)
+            } else {
+                // A draw arrived while a postDelayed is already scheduled.
+                // Mark it so the delayed lambda re-captures the current view tree after it fires.
+                hasPendingDraw = true
```
dustinbyrne: Hi @tsushanth, thanks for the pull request. An agent found the following finding, it's worth verifying:

---

**retry runs after the skipped screen is gone**

**Test sketch (not executed):** at t=1000 draw A; t=1100 draw B; t=1200 draw C; run delayed work at t=2000 and t=3000.

Expected a sequence containing B; current code snapshots current state C at both deadlines, producing A,C,C.

Trigger: ordinary replay with two state changes inside the default 1s throttle window.

Entry: `PostHogAndroid.setup` → replay `onDraw` → `NextDrawListener` → `Throttler` → `generateSnapshot`.

**Fix:** capture/retain immutable pending state before it is replaced, or narrow the claim and issue closure. Add state-based A→B→C, single-pending-draw, continuous-draw, and no-pending-draw tests.

tsushanth:
Hey @dustinbyrne, thanks for digging into this — the concern is valid and I want to be transparent about what this PR does and doesn't fix.

**Why the retry still captures C, not B**

`NextDrawListener.onDraw()` passes `{ onDrawThrottlerCallback() }` to `throttle()` on every draw. The lambda closes over `onDrawThrottlerCallback` — a single reference set at construction time — so it reads the current view tree at the moment it *runs*, not the moment it was enqueued. This means:

- Draw A fires → snapshot taken at t=1000 (A)
- Draw B fires → `postDelayed` scheduled (B's lambda stored in `isThrottling` path), isThrottling=true
- Draw C fires → `hasPendingDraw = true`, C's lambda discarded
- `postDelayed` fires at ~t=2000 → runs B's lambda → snapshots **current screen (C)** → `hasPendingDraw=true` so calls `throttle()` again
- Second snapshot at t=3000 → snapshots **C** again

Result: A, C, C — B is still missing. You're right.

**What this PR actually fixes**

The `hasPendingDraw` retry does eliminate one class of loss: if the user navigates A → B and B arrives right at the throttle boundary, the retry ensures B is captured after the window (since B is still on screen when the lambda runs). But it doesn't help for A → B → C within the same window.

**A proper fix**

Fixing B→C within a window would require `onDrawThrottlerCallback` to capture a snapshot of the view tree eagerly (at `onDraw()` time), not lazily (at `runnable.run()` time). That's an architectural change upstream of `Throttler` — the replay code that creates `NextDrawListener` would need to snapshot the tree and pass the serialised data, rather than passing a function that reads the tree later.

I'm happy to either:
1. Narrow this PR to what it actually delivers: eliminating the "pending draw silently dropped after window" case where the skipped screen is still visible when the retry fires, and update the issue/PR description accordingly.
2. Tackle the eager-snapshot approach in a follow-up if that's the direction you'd prefer.

Let me know which way you'd like to go.

tsushanth:
Hi @ioannisj — the changeset is already included in the branch at `.changeset/fix-throttler-dropped-screens.md` (patch bump for posthog-android). Let me know if anything else is needed to move this forward.

ioannisj (approved):
Hey @tsushanth, agreed that the eager snapshot needs a bigger change than this PR, so it can be addressed as a follow-up. I went ahead and applied the bot's suggestions and removed "Fixes" from the PR body since this does not genuinely close the whole gap.

ioannisj:
Had to resign comment to get this merged, hence the force-push here

PR #608, opened 2026-07-09
feat(errortracking): ignoredExceptionTypes config to skip duplicate exception captures

### LINKED ISSUE #567: Native crash autocapture double-reports fatal React Native JS errors (JavascriptException)

turnipdabeets:

Closes #567. Supersedes #569 (stale-closed; its head branch lives on a fork we can't push to, so this PR carries its commits forward instead — original authorship preserved).

When React Native apps enable native crash autocapture, a fatal JS error is captured twice: once by the JS layer, and once natively when RN rethrows it as `com.facebook.react.common.JavascriptException`. This adds `errorTrackingConfig.ignoredExceptionTypes: MutableList<Class<out Throwable>>` so the native duplicate can be suppressed (mirrors sentry-android's `addIgnoredExceptionForType`; the iOS equivalent already merged in PostHog/posthog-ios#666).

`PostHog.captureException` (autocapture + manual callers) checks the throwable and its cause chain with `Class.isInstance` — R8-safe, and the cause-chain walk covers the `PostHogThrowable` wrapper. The generic `capture("$exception", ...)` path is gated too, by matching the class names serialized in `$exception_list` (per review on #569). The uncaught-exception handler still chains either way — ignoring an exception never changes crash behavior.

On top of #569's final state, this PR:
- restores the ~199 lines of `exceptionSteps` tests that #569 accidentally deleted from `PostHogTest.kt` (the blocker that stalled it) — the test file is now purely additive
- gates the generic capture path (the remaining review ask)
- adds subclass-match and `PostHogThrowable` end-to-end tests

`./gradlew :posthog:test` — full suite green; `PostHogTest` runs 130 tests (7 covering `ignoredExceptionTypes`: direct match, cause-chain match, subclass match, `PostHogThrowable` wrapper, non-match passthrough, and the two generic-capture cases). `spotlessCheck` and `apiCheck` pass.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

- Claude Code session directed by @turnipdabeets: reconstructed #569 on current `main`, keeping tsushanth's feature commit and the review-settled design (Class-based `isInstance` matching, check inside `captureException`) and dropping only the unrelated test deletions.
- All review threads on #569 (marandaneto, ioannisj, greptile) were re-checked against this diff; the two open items (deleted tests, generic-capture gating) are addressed here.
- An independent reviewer agent audited the diff (cause-chain edge cases, `$exception_list` name reconstruction, API compatibility); its two findings — missing subclass and `PostHogThrowable` tests — are included.
- Unverified commits were recreated signed with original authorship per repo signature rules.

### DIFF
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -687,6 +687,17 @@ public class PostHog private constructor(
         }
 
         try {
+            val ignored = config?.errorTrackingConfig?.ignoredExceptionTypes
+            if (!ignored.isNullOrEmpty()) {
+                val match = findIgnoredTypeInCauseChain(throwable, ignored)
+                if (match != null) {
+                    config?.logger?.log(
+                        "Skipping \$exception: ${match.name} (or a cause in its chain) matches ignoredExceptionTypes",
+                    )
+                    return
+                }
+            }
+
             val exceptionProperties =
                 throwableCoercer.fromThrowableToPostHogProperties(
                     throwable,
@@ -706,6 +717,21 @@ public class PostHog private constructor(
             // a captured exception to a PostHog exception event
             config?.logger?.log("captureException has thrown an exception: $e.")
         }
+    }
+
+    private fun findIgnoredTypeInCauseChain(
+        throwable: Throwable,
+        ignored: List<Class<out Throwable>>,
+    ): Class<out Throwable>? {
+        var current: Throwable? = throwable
+        var depth = 0
+        while (current != null && depth < MAX_CAUSE_DEPTH) {
+            val link = current
+            ignored.firstOrNull { it.isInstance(link) }?.let { return it }
+            current = if (link.cause === link) null else link.cause
+            depth++
+        }
+        return null
     }
 
     override fun addExceptionStep(
@@ -1803,6 +1829,10 @@ public class PostHog private constructor(
         private var defaultSharedInstance = shared
 
         private val apiKeys = mutableSetOf<String>()
+
+        // Cap on the cause-chain walk used by ignoredExceptionTypes.
+        // Guards against self-referential or cyclic causes.
+        private const val MAX_CAUSE_DEPTH: Int = 32
 
         /**
          * Captures application log records into PostHog's logs product
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -710,15 +710,8 @@ public class PostHog private constructor(
         }
 
         try {
-            val ignored = config?.errorTrackingConfig?.ignoredExceptionTypes
-            if (!ignored.isNullOrEmpty()) {
-                val match = findIgnoredTypeInCauseChain(throwable, ignored)
-                if (match != null) {
-                    config?.logger?.log(
-                        "Skipping \$exception: ${match.name} (or a cause in its chain) matches ignoredExceptionTypes",
-                    )
-                    return
-                }
+            if (isIgnoredThrowable(throwable)) {
+                return
             }
 
             val exceptionProperties =
@@ -756,21 +749,6 @@ public class PostHog private constructor(
             val className = if (module.isNullOrEmpty()) type else "$module.$type"
             className in ignoredNames
         }
-    }
-
-    private fun findIgnoredTypeInCauseChain(
-        throwable: Throwable,
-        ignored: List<Class<out Throwable>>,
-    ): Class<out Throwable>? {
-        // same cycle detection as ThrowableCoercer, so both walks cover the same chain
-        val seen = hashSetOf<Throwable>()
-        var current: Throwable? = throwable
-        while (current != null && seen.add(current)) {
-            val link = current
-            ignored.firstOrNull { it.isInstance(link) }?.let { return it }
-            current = link.cause
-        }
-        return null
     }
 
     override fun addExceptionStep(
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -762,13 +762,13 @@ public class PostHog private constructor(
         throwable: Throwable,
         ignored: List<Class<out Throwable>>,
     ): Class<out Throwable>? {
+        // same cycle detection as ThrowableCoercer, so both walks cover the same chain
+        val seen = hashSetOf<Throwable>()
         var current: Throwable? = throwable
-        var depth = 0
-        while (current != null && depth < MAX_CAUSE_DEPTH) {
+        while (current != null && seen.add(current)) {
             val link = current
             ignored.firstOrNull { it.isInstance(link) }?.let { return it }
-            current = if (link.cause === link) null else link.cause
-            depth++
+            current = link.cause
         }
         return null
     }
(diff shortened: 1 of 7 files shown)


turnipdabeets:
@ioannisj #569 was closed due to being stale but it was really close to ready so I took those commits, resolved the last requests you had on there and reopened it here (couldn’t push to other brnach) - mind taking a look?

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -718,6 +742,37 @@ public class PostHog private constructor(
...
+    ): Boolean {
+        val exceptionList = properties?.get("\$exception_list") as? List<*> ?: return false
+        val ignoredNames = ignored.map { it.name }
+        return exceptionList.any { entry ->
+            val exception = entry as? Map<*, *> ?: return@any false
+            val type = exception["type"] as? String ?: return@any false
+            val module = exception["module"] as? String
+            // module + type rejoins the runtime class name ThrowableCoercer split apart
+            val className = if (module.isNullOrEmpty()) type else "$module.$type"
+            className in ignoredNames
```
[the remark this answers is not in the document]
turnipdabeets: This asymmetry is inherent to the serialized path, so we're keeping exact-name matching here. For bridge events we only have class-name strings — resolving them with `Class.forName` at capture time would trigger class initialization for arbitrary event-supplied names, and still couldn't work for minified or JS-originated types. Every native throwable (autocapture and manual `captureException`) goes through the `isInstance` path and gets full subclass matching; this `$exception_list` gate is a best-effort backstop for pre-serialized events, matching the merged iOS behavior (PostHog/posthog-ios#666, exact-name). The KDoc on `ignoredExceptionTypes` documents the name-based matching for this path.

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -718,6 +742,37 @@ public class PostHog private constructor(
...
+        }
+    }
+
+    private fun findIgnoredTypeInCauseChain(
+        throwable: Throwable,
+        ignored: List<Class<out Throwable>>,
+    ): Class<out Throwable>? {
+        var current: Throwable? = throwable
+        var depth = 0
+        while (current != null && depth < MAX_CAUSE_DEPTH) {
```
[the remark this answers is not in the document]
turnipdabeets: Good catch — fixed in cefe9f4: the walk now uses the same seen-set cycle detection as `ThrowableCoercer` instead of a depth cap, so the two walks always cover the same chain and an ignored type can't be serialized into `$exception_list` while escaping the `isInstance` check.

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -620,6 +620,19 @@ public class PostHog private constructor(
                 groupIdentify = true
             }
 
+            // Externally-built $exception events (e.g. the Flutter/RN bridge) carry only
```
ioannisj, line 623: Should we run this in captureExceptionStateless as well?

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingConfig.kt
```diff
@@ -37,4 +37,12 @@ public class PostHogErrorTrackingConfig
...
          */
         public val exceptionSteps: PostHogExceptionStepsConfig = PostHogExceptionStepsConfig(),
+        /**
+         * Throwable classes to drop during capture. The throwable and every cause in
+         * its chain are matched via [Class.isInstance]. Also applies to `$exception`
+         * events sent through the generic capture path (matched by class name).
+         *
+         * Defaults to empty.
+         */
+        public val ignoredExceptionTypes: MutableList<Class<out Throwable>> = mutableListOf(),
```
ioannisj, line 47: iOS ships with a default RCTFatalException here, do we need something similar?

posthog/src/main/java/com/posthog/PostHog.kt
ioannisj: Then again looking at posthog-ios we only capture through captureException() so there's a gap there?

ioannisj (approved):
LG but left a couple of comments

posthog/src/main/java/com/posthog/PostHog.kt
turnipdabeets: should I update iOS as well?

ioannisj: yeah I think that we should be ignoring through capture() as well. Probably missed it in the iOS PR

ioannisj: For Android, there's also captureExceptionStateless exposed as a public API though

turnipdabeets: Both covered now:

- `captureExceptionStateless` runs the same cause-chain gate as of 6994035 — the check moved into `PostHogStateless` so both entry points share one implementation (+2 stateless tests).
- For the iOS gap: opened PostHog/posthog-ios#713. All the generic `capture(...)` overloads funnel into `captureInternal`, so the `$exception_list` check now lives there; the `captureException` copy became redundant and was removed, and the crash-report path keeps its early check to short-circuit report processing before the drop.

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingConfig.kt
turnipdabeets: We don't think Android can ship a safe default here: the list is `Class`-based (per the R8 discussion on #569), so a built-in default would mean referencing `JavascriptException::class` — an RN dependency in the core SDK — and a name-string default instead would break under R8 renaming, which is why name matching was rejected in the first place. Our plan is for the RN plugin to register `com.facebook.react.common.JavascriptException::class.java` when it enables native crash autocapture — same split sentry uses (core exposes the API, the RN layer supplies the default). Happy to go another way if you'd prefer.

PR #613, opened 2026-07-13
feat(replay): internal session-replay snapshot hook for out-of-engine capture

turnipdabeets:

First-party wrapper SDKs (starting with posthog-flutter) need to record native screens that cover their engine-rendered UI — paywalls and other activities presented on top of a Flutter surface. The wrapper runs its own out-of-engine occlusion detection and needs to drive a one-shot capture of the top-most native window on its own cadence.

This adds an opt-in internal hook for that, marked `@PostHogInternalReplayApi` (`@RequiresOptIn`). It is **not public API**: it shares snapshot state with the normal timer-driven capture, so only a coordinating first-party SDK should call it.

Sibling of PostHog/posthog-ios#714; consumed by PostHog/posthog-flutter#473. PostHog/posthog-flutter#151 is the end-user issue this enables fixing.

## :hammer: Changes

- `PostHogReplayIntegration.captureSessionReplaySnapshot(excludeView, forceFullSnapshot, isStillValid, onResult)` — schedules a one-shot capture of the top-most application window on the capture executor. Returns whether work was *scheduled*; `onResult` reports whether a frame was actually *delivered* (the caller's retry signal). `excludeView` (the caller's own decor view) is never captured; `isStillValid` lets queued work self-drop once the caller's episode ends; `forceFullSnapshot` resets the decor view's snapshot state (meta + full snapshot + last frame) so a reused activity opens each episode with a meta carrying the activity's name and a full frame.
- `PostHogInternalReplayApi` — the opt-in marker annotation.
- `generateSnapshot(...forceScreenshot)` now returns `Boolean` (whether a frame was produced) so the bridge caller can tell a real capture from a silent early-bail.

- Exercised end-to-end from posthog-flutter's example app on the Pixel 6 emulator: native paywall activity over Flutter; the episode held for the full duration and captured the native screen, with graceful fallback to the wrapper's placeholder on repeated capture failure.
- `compileDebugKotlin` and ktlint clean.
- No new unit tests here: the capture path needs a hosted window; the driving state machine is tested on the Flutter side.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

- Implemented with Claude Code (Opus 4.8 and Fable 5 sessions) driven by @turnipdabeets, alongside the matching posthog-ios SPI hook (PostHog/posthog-ios#714) and the posthog-flutter feature consuming both.
- Design decisions: `@RequiresOptIn` marker over public API; top-most **activity** window only (a Dialog decor would collapse the replay viewport, a PopupWindow has no phoneWindow — both are treated as overlays of the activity beneath); scheduled-vs-delivered split in the return/callback contract so the wrapper's strike/fallback logic keys on actual frame delivery.
- Builds on the previously extracted fixes #609 (empty-wireframe skip) and #610 (decor-view map synchronization), which this branch no longer carries copies of after rebase.

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogInternalReplayApi.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogInternalReplayApi.kt
@@ -1,0 +1,17 @@
+package com.posthog.android.replay
+
+/**
+ * Marks a session-replay API that exists only for first-party PostHog wrapper
+ * SDKs (e.g. posthog-flutter) to drive out-of-engine capture. It is not a
+ * public API: it has no source/binary stability guarantees and can change or
+ * be removed without a major-version bump. App code should never opt in.
+ */
+@RequiresOptIn(
+    message =
+        "Internal PostHog session-replay API for first-party SDK integrations only. " +
+            "Not covered by semantic versioning; do not use from app code.",
+    level = RequiresOptIn.Level.ERROR,
+)
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.FUNCTION, AnnotationTarget.CLASS, AnnotationTarget.PROPERTY)
+public annotation class PostHogInternalReplayApi
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -555,6 +555,15 @@ public class PostHogReplayIntegration(
                 return false
             }
             executor.submit {
+                // A throwing onResult would land in the catch below and fire a
+                // second time — report exactly once per scheduled capture.
+                var resultReported = false
+
+                fun report(delivered: Boolean) {
+                    if (resultReported) return
+                    resultReported = true
+                    onResult(delivered)
+                }
                 try {
                     // Validity and the first-of-episode reset both happen on
                     // the capture thread: the reset mutates snapshot status
@@ -564,7 +573,7 @@ public class PostHogReplayIntegration(
                         // The contract promises onResult for every scheduled
                         // capture; a silent self-drop would leave the caller's
                         // in-flight tracking latched forever.
-                        onResult(false)
+                        report(false)
                         return@submit
                     }
                     if (forceFullSnapshot) {
@@ -578,10 +587,10 @@ public class PostHogReplayIntegration(
                     if (!delivered) {
                         config.logger.log("Session Replay bridge capture produced no frame (will retry next tick).")
                     }
-                    onResult(delivered)
+                    report(delivered)
                 } catch (e: Throwable) {
                     config.logger.log("Session Replay bridge capture failed: $e.")
-                    onResult(false)
+                    report(false)
                 }
             }
             return true
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -31,6 +31,7 @@ import android.view.PixelCopy
 import android.view.View
 import android.view.ViewGroup
 import android.view.Window
+import android.view.WindowManager
 import android.webkit.WebView
 import android.widget.Button
 import android.widget.CheckBox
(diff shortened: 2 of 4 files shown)


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -502,6 +503,90 @@ public class PostHogReplayIntegration(
...
+            }
+            executor.submit {
+                try {
+                    // Validity and the first-of-episode reset both happen on
+                    // the capture thread: the reset mutates snapshot status
+                    // fields that are otherwise only touched here, and a
+                    // stale queued capture must not emit after the episode.
+                    if (!isStillValid()) {
+                        return@submit
+                    }
```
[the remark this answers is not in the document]
turnipdabeets: Good catch — fixed in 0ea843f: the stale-capture bail now reports onResult(false) so the delivered-frame contract holds for every scheduled capture. (The current caller self-healed via its episode-transition reset, but the contract should not rely on that.)

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -502,6 +503,94 @@ public class PostHogReplayIntegration(
...
+                            status.sentFullSnapshot = false
+                            status.sentMetaEvent = false
+                            status.lastSnapshot = null
+                        }
+                    }
+                    val delivered = generateSnapshot(WeakReference(decorView), WeakReference(window), forceScreenshot = true)
+                    if (!delivered) {
+                        config.logger.log("Session Replay bridge capture produced no frame (will retry next tick).")
+                    }
+                    onResult(delivered)
```
ioannisj: i think onResult can fire twice when the first onResult throws (since we call it inside a catch)

posthog-android/src/main/java/com/posthog/android/replay/PostHogInternalReplayApi.kt
```diff
@@ -0,0 +1,17 @@
...
+ */
+@RequiresOptIn(
+    message =
+        "Internal PostHog session-replay API for first-party SDK integrations only. " +
+            "Not covered by semantic versioning; do not use from app code.",
+    level = RequiresOptIn.Level.ERROR,
+)
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.FUNCTION, AnnotationTarget.CLASS, AnnotationTarget.PROPERTY)
+public annotation class PostHogInternalReplayApi
```
ioannisj, line 17: Wonder if we should rework the existing `@PostHogInternal` annotation here? Even though it's going to be a bigger change, so we can test how this works with reaply

ioannisj (approved):

posthog-android/src/main/java/com/posthog/android/replay/PostHogInternalReplayApi.kt
turnipdabeets: I have this PR open https://github.com/PostHog/posthog-flutter/pull/473 and i was testing with these locally. Or is there another way you mean to test?

ioannisj: Sorry I mean like we have an SPI `PostHogInternal` which is a bit more generic, I was wndering if we should reuse (and rewrite) the existing `@PostHogInternal` annotation here to expose internal bridge methods to flutter, rather than rolling out a new `PostHogInternalReplayApi`

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
turnipdabeets: Confirmed — a throwing callback would land in the catch and fire a second time (the stale-bail path had the same exposure). Added a once-latch so every scheduled capture reports exactly once. Fixed in aa378e1.

posthog-android/src/main/java/com/posthog/android/replay/PostHogInternalReplayApi.kt
turnipdabeets: I looked at reusing it — the existing `@PostHogInternal` is a plain doc annotation (not `@RequiresOptIn`), so making it opt-in would ripple through every existing cross-module usage. They also mean different things: `@PostHogInternal` is "public only because of the module layout", while this one is "a real contract for first-party wrappers, opt in deliberately". I'd keep them separate here unless we want one taxonomy — wdyt?

PR #618, opened 2026-07-14
feat: add bootstrap support for identity and feature flag seeding

ioannisj:
## Problem

On a cold start, before the first `/flags` response comes back, the SDK has no caller-controlled identity or flag state. Events captured synchronously during `setup()` carry the SDK-generated anonymous UUID as `$distinct_id`, and feature flag reads fall back to not-loaded defaults. An app that already knows who the user is (for example from a backend session token), or that wants specific flags available instantly on first launch, has no way to seed either. posthog-js has a `bootstrap` option for exactly this; the mobile SDKs didn't.

## Changes

Adds `PostHogConfig.bootstrap`, a `PostHogBootstrapConfig` you set before `setup()`, that pre-seeds identity and feature flag state before any network request. Identity is seeded only on the first launch when nothing is persisted; the feature-flag base layer applies each launch until `/flags` overlays it or `reset()` drops it. It mirrors the `bootstrap` option in posthog-js, and a matching change is going into posthog-ios.

On identity, it seeds the distinct id on a fresh install, either as the anonymous id or, when `isIdentifiedId` is true, as an already-identified user's id. If a local anonymous user already exists it merges into the identified bootstrap via `identify()`, and if a different user is already identified it keeps the existing identity and logs a warning. It never overwrites a persisted identity, and it keeps an identified bootstrap out of the anonymous and device id so a real user id can't leak onto later users after `reset()`.

On feature flags, bootstrapped flags and payloads form a base layer that's served immediately, before the first `/flags` response. Loaded values overlay the base on overlapping keys while bootstrapped-only keys survive a load. `$feature_flag_called` is enriched with `$feature_flag_bootstrapped_response`, `$feature_flag_bootstrapped_payload`, and `$used_bootstrap_value`. The base layer is dropped on `reset()`, and bootstrapped keys are kept out of the persisted flag cache, so a later user or a cold restart is never served the previous session's bootstrapped values. Host-provided values are sanitized in `PostHogRemoteConfig` at construction, so a non-serializable entry (a `NaN`, an arbitrary object) can't take down the whole `FEATURE_FLAGS` cache write.

## Testing

Unit tests in `PostHogBootstrapTest` and `PostHogRemoteConfigTest` cover identity seeding (anonymous, identified, fresh install, existing-anon merge, preserving a different identified user, blank id, and the opted-out merge that matches posthog-js), the flag base-layer overlay (loaded values winning, bootstrapped-only keys surviving, cache load, the `hasFeatureFlags=false` re-seed, `reset()` clearing the layer, and bootstrapped keys never reaching the durable flag cache), the `$feature_flag_called` enrichment, and value sanitization. `./gradlew :posthog:apiCheck :posthog:test` is green and the public API dump is updated.

**DRI:** @ioannisj

### DIFF
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -186,15 +186,22 @@ public class PostHog private constructor(
                     )
                 val onRemoteConfigLoaded =
                     PostHogOnRemoteConfigLoaded {
-                        try {
-                            val surveys = remoteConfig?.getSurveys() ?: emptyList()
-                            surveysHandler?.onSurveysLoaded(surveys)
-                        } catch (e: Throwable) {
-                            config.logger.log("Failed to notify surveys loaded: $e.")
-                        }
+                        // The callback fires whether the attempt succeeded or terminally failed;
+                        // hasRemoteConfigFetched() tells them apart. On a failure no fresh values
+                        // were applied, so integrations fall back to their cached state instead.
+                        if (remoteConfig?.hasRemoteConfigFetched() == true) {
+                            try {
+                                val surveys = remoteConfig?.getSurveys() ?: emptyList()
+                                surveysHandler?.onSurveysLoaded(surveys)
+                            } catch (e: Throwable) {
+                                config.logger.log("Failed to notify surveys loaded: $e.")
+                            }
 
-                        // Notify all integrations about remote config changes
-                        notifyIntegrationsRemoteConfig(config)
+                            // Notify all integrations about remote config changes
+                            notifyIntegrationsRemoteConfig(config, loaded = true)
+                        } else {
+                            notifyIntegrationsRemoteConfig(config, loaded = false)
+                        }
                     }
 
                 val featureFlags =
@@ -328,10 +335,13 @@ public class PostHog private constructor(
         }
     }
 
-    private fun notifyIntegrationsRemoteConfig(config: PostHogConfig) {
+    private fun notifyIntegrationsRemoteConfig(
+        config: PostHogConfig,
+        loaded: Boolean,
+    ) {
         config.integrations.forEach { integration ->
             try {
-                integration.onRemoteConfig()
+                integration.onRemoteConfig(loaded)
             } catch (e: Throwable) {
                 config.logger.log("Integration ${integration.javaClass.name} onRemoteConfig failed: $e.")
             }
@@ -694,6 +704,19 @@ public class PostHog private constructor(
                 groupIdentify = true
             }
 
+            // Externally-built $exception events (e.g. the Flutter/RN bridge) carry only
+            // serialized class names, so they are matched by name instead of isInstance.
+            val ignored = config?.errorTrackingConfig?.ignoredExceptionTypes
+            if (!ignored.isNullOrEmpty() &&
+                event == PostHogEventName.EXCEPTION.event &&
+                hasIgnoredTypeInExceptionList(properties, ignored)
+            ) {
+                config?.logger?.log(
+                    "Skipping \$exception: an entry in \$exception_list matches ignoredExceptionTypes",
+                )
+                return
+            }
+
             // Attach the buffered exception steps to any $exception event (unless the caller
             // already supplied them), so externally-built exceptions (e.g. from the Flutter/RN
             // bridge) carry steps too, not only those captured via captureException(throwable).
@@ -771,6 +794,10 @@ public class PostHog private constructor(
         }
 
         try {
+            if (isIgnoredThrowable(throwable)) {
+                return
+            }
+
             val exceptionProperties =
                 throwableCoercer.fromThrowableToPostHogProperties(
                     throwable,
@@ -789,6 +816,22 @@ public class PostHog private constructor(
             // we swallow all exceptions that the SDK has thrown by trying to convert
             // a captured exception to a PostHog exception event
             config?.logger?.log("captureException has thrown an exception: $e.")
+        }
+    }
+
+    private fun hasIgnoredTypeInExceptionList(
+        properties: Map<String, Any>?,
+        ignored: List<Class<out Throwable>>,
+    ): Boolean {
+        val exceptionList = properties?.get("\$exception_list") as? List<*> ?: return false
+        val ignoredNames = ignored.map { it.name }
+        return exceptionList.any { entry ->
+            val exception = entry as? Map<*, *> ?: return@any false
+            val type = exception["type"] as? String ?: return@any false
+            val module = exception["module"] as? String
+            // module + type rejoins the runtime class name ThrowableCoercer split apart
+            val className = if (module.isNullOrEmpty()) type else "$module.$type"
+            className in ignoredNames
         }
     }
 
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -314,9 +314,8 @@ public class PostHog private constructor(
                 // only because of testing in isolation, this flag is always enabled
                 @Suppress("DEPRECATION")
                 if (reloadFeatureFlags) {
-                    // Bootstrap flags were applied in RemoteConfig's init; fire the flags-loaded
-                    // callbacks immediately (matches posthog-js firing onFeatureFlags when bootstrap is
-                    // applied) so listeners aren't blocked on the first network response.
+                    // Bootstrap already seeded the flags in RemoteConfig's init; fire the flags-loaded
+                    // callbacks now so listeners aren't blocked on the first network response.
                     if (remoteConfig?.hasBootstrapFlags() == true) {
                         notifyFeatureFlagsCallback(internalOnFeatureFlagsLoaded)
                         notifyFeatureFlagsCallback(config.onFeatureFlags)
--- a/posthog/src/main/java/com/posthog/PostHog.kt
+++ b/posthog/src/main/java/com/posthog/PostHog.kt
@@ -317,14 +317,15 @@ public class PostHog private constructor(
                         config.preloadFeatureFlags -> reloadFeatureFlags(config.onFeatureFlags)
                     }
                 }
+
+                // Reconcile a differing identified bootstrap now that the SDK is enabled. Runs under
+                // setupLock — safe on Android because identify() never takes it (unlike iOS) — which
+                // keeps it mutually exclusive with close() and guarded by the catch below.
+                reconcileBootstrapIdentityIfNeeded(config)
             } catch (e: Throwable) {
                 config.logger.log("Setup failed: $e.")
             }
         }
-
-        // After the setup lock is released: reconcile a differing identified bootstrap against the
-        // existing local identity. This may call identify(), which captures an event and reloads flags.
-        reconcileBootstrapIdentityIfNeeded()
     }
 
     private fun notifyIntegrationsRemoteConfig(config: PostHogConfig) {
@@ -407,15 +408,13 @@ public class PostHog private constructor(
     }
 
     /**
-     * Reconciles a differing identified bootstrap against the existing local identity. A fresh
-     * install is already seeded by [applyBootstrapIfNeeded]; this only runs when the bootstrapped
-     * id differs from the current distinct id: an existing anonymous user is merged into the
-     * identified id via [identify], while a different already-identified user is preserved with a
-     * warning. Must run after setup completes — [identify] requires the SDK enabled and captures an
-     * `$identify` event.
+     * Reconciles a differing identified bootstrap against the local identity — fresh installs are
+     * already seeded by [applyBootstrapIfNeeded]. Merges an existing anonymous user via [identify];
+     * preserves a different identified user with a warning. Must run after setup — [identify] needs
+     * the SDK enabled and captures an event.
      */
-    private fun reconcileBootstrapIdentityIfNeeded() {
-        val bootstrap = config?.bootstrap ?: return
+    private fun reconcileBootstrapIdentityIfNeeded(config: PostHogConfig) {
+        val bootstrap = config.bootstrap ?: return
         if (!bootstrap.isIdentifiedId) {
             return
         }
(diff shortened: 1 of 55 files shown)


posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
```diff
@@ -701,24 +724,27 @@ public class PostHogRemoteConfig(
...
 
                             val normalizedPayloads = normalizePayloads(normalizedResponse.featureFlagPayloads)
-                            this.featureFlagPayloads = (this.featureFlagPayloads ?: mapOf()) + normalizedPayloads
+                            this.featureFlagPayloads =
+                                withBootstrapPayloadBase((this.featureFlagPayloads ?: mapOf()) + normalizedPayloads)
                         }
                     } else {
                         this.flags = normalizedResponse.flags
-                        this.featureFlags = normalizedResponse.featureFlags
+                        this.featureFlags = withBootstrapBase(normalizedResponse.featureFlags)
```
dustinbyrne: [P1] A complete response should not independently overlay remote flags and payloads on the permanent bootstrap base.

For example:

```
bootstrap flags:    beta-ui = "variant-a", legacy = true
bootstrap payload:  beta-ui = { "color": "blue" }
complete response:  beta-ui = "variant-b", payloads = {}
```

The current result is `beta-ui="variant-b"` with the stale `variant-a` payload, while `legacy=true` also survives despite being absent remotely.

posthog-js replaces both maps on a complete response (`packages/browser/src/posthog-featureflags.ts:128-175`) and explicitly tests that existing/bootstrap-only flags disappear (`featureflags.test.ts:3264-3271`).

Please replace the maps on complete responses and retain merge behavior only for partial/error responses. If preserving bootstrap-only keys is intentional, payloads for every remotely evaluated flag must still be cleared when the response omits their payload.

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -322,6 +324,11 @@ public class PostHog private constructor(
                         config.preloadFeatureFlags -> reloadFeatureFlags(config.onFeatureFlags)
                     }
                 }
+
+                // Reconcile a differing identified bootstrap now that the SDK is enabled. Runs under
+                // setupLock — safe on Android because identify() never takes it (unlike iOS) — which
+                // keeps it mutually exclusive with close() and guarded by the catch below.
+                reconcileBootstrapIdentityIfNeeded(config)
```
dustinbyrne, line 339: [P1] This reconciliation happens after setup-time work that already consumes identity.

The current order is `startSession()`, integration installation, initial remote/flag loading, and only then reconciliation. `PostHogAppInstallIntegration.install()` synchronously captures `Application Installed` or `Application Updated` (`posthog-android/.../PostHogAppInstallIntegration.kt:24-65`).

For example, with persisted `anonymousId="anon-123"` and an identified bootstrap for `user-456`, the app-install event and initial `/flags` request can use `anon-123`; only afterward does `identify("user-456")` run. A second flag reload cannot undo the incorrectly attributed lifecycle event or prevent callers from observing the first anonymous flag result.

Please reconcile after the SDK is enabled but before session startup, integration installation, and initial remote loading. A test integration that captures from `install()` should observe `user-456`, as should the first flags request.

dustinbyrne (commented):
Compared the bootstrap implementation against posthog-js browser semantics (and the iOS PR). Leaving inline comments for the material behavioral differences and one mobile-specific precedence decision that should be documented if intentional.

posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -367,6 +374,83 @@ public class PostHog private constructor(
...
+        if (bootstrapId.isNullOrBlank() || !isEnabled()) {
+            return
+        }
+
+        // Self-guard like applyBootstrapIfNeeded: the identity reads and identify() (which captures
+        // an event) must not abort setup.
+        try {
+            if (distinctId == bootstrapId) {
+                // Already this identity (fresh-install seed, or unchanged); reconciling would only warn spuriously.
+                return
```
dustinbyrne, line 452: [P1] Matching IDs do not necessarily mean reconciliation is a no-op.

Example persisted state:

```
anonymousId = "user-123"
distinctId resolves to "user-123"
isIdentified = false
```

Then setup receives `bootstrap.distinctId="user-123"` with `isIdentifiedId=true`. This branch returns and leaves `isIdentified=false`, so `PersonProfiles.IDENTIFIED_ONLY` continues treating the user as anonymous.

posthog-js keeps the same distinct ID but changes `USER_STATE` to identified (`packages/browser/src/posthog-core.ts:784-794`). It does not emit `$identify` because the ID did not change.

Please mark the user identified when the IDs match but local state is still anonymous.

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
```diff
@@ -903,8 +930,8 @@ public class PostHogRemoteConfig(
 
             synchronized(featureFlagsLock) {
                 this.flags = flags
-                this.featureFlags = featureFlags
-                this.featureFlagPayloads = payloads
+                this.featureFlags = withBootstrapBase(featureFlags)
```
dustinbyrne: [P2/design] Is it intentional that persisted cached values win over caller-provided bootstrap values?

For example, if the persisted cache contains `checkout-redesign=false` and a new bootstrap contains `checkout-redesign=true`, Android serves `false`. posthog-js applies bootstrap as its initial complete snapshot and serves `true`.

If cache-first precedence is intentional for mobile, this should be documented as a platform-specific difference rather than described as mirroring posthog-js. Otherwise bootstrap should replace overlapping cached values during initialization.

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
ioannisj: Good catch, and it was the spec drifting from posthog-js that I'd inherited. parseFlagsResponse replaces on a complete response and only merges for partial/errorsWhileComputingFlags, so bootstrapped-only keys and the stale payload shouldn't survive a complete load. Fixed it to replace on complete and merge only on partial/errored, and added a test for your beta-ui payload case. We corrected the shared spec too (a fix-bootstrap-js-parity change) since it had drifted here.

posthog/src/main/java/com/posthog/PostHog.kt
ioannisj: Yeah, reconcile was putting an existing-anon + identified bootstrap's Application Installed on the old anon id. I looked at moving reconcile before integrations like posthog-ios did, but on Android's single-threaded remote-config executor that makes identify()'s flags reload resolve remote config ahead of setup's own /config load, so session-replay gating would decide on cached config first. Since the $identify merge already links the anon install event to the identified user server-side, I left reconcile at setup-end rather than take that ordering hazard. Happy to do the reload-suppression version if you'd want the capture-time id too, wdyt?

posthog/src/main/java/com/posthog/PostHog.kt
ioannisj: Right, the matching-id early return left isIdentified false, so IDENTIFIED_ONLY kept treating them as anonymous. posthog-js flips USER_STATE to identified even when the id matches, so I added that to the spec and here: matching id with an identified bootstrap now marks identified, no redundant $identify.

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
ioannisj: This flipped from "document it" to a real fix once we re-audited posthog-js. initialize() applies bootstrap as a complete receivedFeatureFlags, so bootstrap wins over the persisted cache, not the reverse. The spec now says bootstrap-wins and Android matches, the cache load applies the bootstrap snapshot last.

ioannisj:
Did a wider js-parity sweep while on this and found two more small differences, both already handled here the same way as iOS. The writeup's on the iOS PR: https://github.com/PostHog/posthog-ios/pull/715#issuecomment-4977899605

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
```diff
@@ -738,13 +768,31 @@ public class PostHogRemoteConfig(
...
+                    // keys into the durable cache and a later launch (bootstrap dropped, no reset())
+                    // would serve them as genuine remote flags. A key the server resolved to a
+                    // different value is kept; with no bootstrap this session the filter is a no-op.
                     val featureFlags = this.featureFlags ?: mapOf()
-                    preferences.setValue(FEATURE_FLAGS, featureFlags)
+                    val serverFeatureFlags =
+                        if (bootstrappedFlags.isEmpty()) {
+                            featureFlags
+                        } else {
+                            featureFlags.filterKeys { bootstrappedFlags[it] != featureFlags[it] }
```
dustinbyrne: [P2] Filtering cache entries by value equality can discard legitimate server-confirmed flags.

Equality with bootstrap does not prove that the value came only from bootstrap. For example:

```
bootstrap:
    beta-ui = true
    payload = {"color":"blue"}

complete remote response:
    beta-ui = true
    payload = {"color":"blue"}
```

The server has explicitly confirmed both values, but these filters remove them because they equal the bootstrap values. The in-memory result is correct for this run, while the durable served-value caches omit `beta-ui` (a V4 response may still retain separate flag-detail metadata).

On the next launch without bootstrap, or while offline, the public flag and payload getters cannot recover `beta-ui` from the served-value cache even though the previous `/flags` response confirmed it. The payload filter can independently remove a valid server payload as well.

Consider tracking which keys were successfully returned by `/flags` and persisting those keys, rather than inferring provenance from value inequality. For a partial response, this could persist returned successful keys while excluding only unresolved bootstrap-only keys.

A regression test could load a response equal to bootstrap, construct a second `PostHogRemoteConfig` with the same preferences but no bootstrap, and assert that the confirmed flag and payload remain available.

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
```diff
@@ -1180,6 +1230,111 @@ public class PostHogRemoteConfig(
...
+        this.evaluatedAt = null
+    }
+
+    /**
+     * Whether a bootstrap snapshot was applied at init. Lets setup fire the flags-loaded callbacks
+     * immediately (matching posthog-js firing onFeatureFlags when bootstrap is applied) so listeners
+     * aren't blocked on the first network response. Stays true after a load overlays the snapshot;
+     * only read once at setup, before any load completes.
+     */
+    public fun hasBootstrapFlags(): Boolean = bootstrappedFlags.isNotEmpty()
```
dustinbyrne: [P2] These bootstrap helpers unnecessarily become supported public API.

The API snapshot now exposes:

- `hasBootstrapFlags()`
- `getBootstrappedFeatureFlag()`
- `getBootstrappedFeatureFlagPayload()`
- `hasLoadedFeatureFlagsFromRemote()`

`hasBootstrapFlags()` is consumed only by `PostHog`, while the other three are consumed only by tests. Because `PostHogRemoteConfig` is already present in the public API dump, callers can compile against these methods and we would need to preserve them.

Consider making these helpers `internal`, removing the test-only getters where practical, and testing through the existing internal atomic metadata API. `apiCheck` should then show only the intended `PostHogBootstrapConfig` and `PostHogConfig.bootstrap` additions.

posthog/src/main/java/com/posthog/PostHogBootstrapConfig.kt
```diff
@@ -0,0 +1,57 @@
...
+ * caller-controlled `$distinct_id` rather than the SDK-generated UUID, and feature-flag
+ * reads return caller-provided values before the first `/flags` response. Mirrors the
+ * [bootstrap option in posthog-js](https://posthog.com/docs/feature-flags/bootstrapping).
+ *
+ * Bootstrap identity applies to the very first session. An anonymous bootstrap
+ * ([isIdentifiedId] false) seeds the anonymous id only when none is persisted, and is
+ * otherwise ignored. An identified bootstrap ([isIdentifiedId] true) seeds the distinct id on
+ * a fresh install; if a local anonymous user already exists it is merged into the identified
+ * id via `identify()`, while a different already-identified user is preserved and a warning is
+ * logged. Bootstrapped feature flags form a base layer only: values from `/flags` overlay them
```
dustinbyrne: [P2/docs] This KDoc still describes the old permanent-base behavior.

It says bootstrapped-only keys remain available after `/flags`, but the current implementation replaces the complete
snapshot at `PostHogRemoteConfig.kt:745-752`.

For example:

```
bootstrap: beta-ui=true, legacy=true
complete response: beta-ui=false
```

The implementation returns `beta-ui=false` and removes `legacy`, while this documentation says `legacy` remains available.

Consider describing bootstrap as an initial snapshot: it wins over persisted cache at setup, a complete response replaces it entirely, and only partial or errored responses preserve unrecomputed keys. The matching text in `PostHogConfig.kt:317-322` could be updated as well.

dustinbyrne (commented):
Follow-up review after the latest bootstrap changes.

dustinbyrne (approved):
comments above are decent considerations, but this is looking pretty good now

i'd at least consider reviewing the public API scope before merge

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
ioannisj: Good catch. The value-inequality filter conflated "equals bootstrap" with "came from bootstrap", so a server-confirmed flag that matched its bootstrap value got dropped from the durable cache.

Switched it to provenance: I track the keys `/flags` actually evaluated this session (accumulated across partials, reset on `clear()`) and persist by membership instead, so a confirmed key survives even when it equals bootstrap while bootstrap-only keys still never reach disk.

Added your exact regression too: a complete response equal to bootstrap, then a second `PostHogRemoteConfig` with the same prefs and no bootstrap still resolves the flag and its payload. 17614652

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
ioannisj: Agreed, made all four `internal`. `hasBootstrapFlags()` has a single internal caller and the other three were test-only, and tests reach internals in-module fine, so the dump now shows only `PostHogBootstrapConfig` and `PostHogConfig.bootstrap`. a20ddf71

posthog/src/main/java/com/posthog/PostHogBootstrapConfig.kt
ioannisj: Fixed. Rewrote it as an initial snapshot rather than a permanent base: it wins over the persisted cache at setup, a complete `/flags` response replaces it entirely (bootstrap-only keys dropped), and only a partial or errored response keeps the keys it didn't recompute. Updated the matching text in `PostHogConfig.kt` too. d7d5f9c0

PR #627, opened 2026-07-16
feat(flags): minimize $feature_flag_called events for non-experiment flags

haacked:

`$feature_flag_called` events carry the full enriched properties (context envelope, super properties, `$active_feature_flags`, the `$feature/<key>` enumeration) on every flag call, even for flags not linked to an experiment. This PR trims those events to a strict allowlist **iff** the server-controlled gate is on (top-level `minimalFlagCalledEvents` in the v2 `/flags` response) **AND** the flag's `has_experiment` is exactly `false`. Any missing signal (field absent, legacy response shape, cached state without it, `has_experiment` unknown) sends the full legacy shape unchanged, and experiment-linked flags always send the full envelope. Server-gated because rolling this out unconditionally could break existing insights; announcement/comms come before enablement.

Implementation notes:

- All changes live in the pure-JVM `posthog` core module.
- The gate is persisted in preferences alongside the cached flags so it survives app restarts; the key is in `ALL_INTERNAL_KEYS`, so it cannot leak into events via registered super properties.
- The gate is parsed overwrite-not-merge: a response without the field flips it off, so full events resume as soon as the server stops sending it. It is also cleared on `reset()`/identity change.
- The allowlist is applied at the single capture choke point (`captureStateless`) after property assembly, building the minimal property set from the allowlist rather than deleting from the full set.
- On a real Android cold restart, cached flag details deserialize without `has_experiment`, so events stay full until the first fresh `/flags` response of the session (fail-safe, bounded staleness).

Part of a cross-SDK rollout; reference implementation and fuller context: PostHog/posthog-python#748

Ran locally:

- `./gradlew :posthog:test` — 805 tests including the new ones: gated + no experiment → exact allowlist key set (super props, context envelope, `$feature/<key>`, `$active_feature_flags` stripped); gated + experiment → full; ungated / field absent → full; gate + minimal shape persist across a simulated SDK restart; gate parse/persist/flip-off/clear at the `PostHogRemoteConfig` level.
- `./gradlew :posthog-server:test`
- `./gradlew spotlessCheck`
- `./gradlew :posthog:apiCheck :posthog-server:apiCheck` (after `apiDump`)

Could **not** run `:posthog-android:testDebugUnitTest` or Android lint in the authoring environment (no Android SDK available) — no Android-module sources changed, but CI must confirm.

## :pencil: Checklist

### If releasing new changes

---


posthog/src/main/java/com/posthog/PostHog.kt
```diff
@@ -1692,6 +1692,10 @@ public class PostHog private constructor(
         }
     }
 
+    override fun isMinimalFlagCalledEventsEnabled(): Boolean {
+        return remoteConfig?.isMinimalFlagCalledEventsEnabled() == true
+    }
```
dustinbyrne, line 1697: this definition would enable minimal flag called events for Android - should `posthog-android` be listed in the changeset?

or is there something else missing?

haacked: Yep, `posthog-android` belongs in there. Added it.

Nothing is missing on the Android side. `PostHogAndroid.setup()` builds the core `PostHog`, so this override is the Android implementation, and `PostHogSharedPreferences` filters on the shared `ALL_INTERNAL_KEYS`, which already includes the new preference key.

The gap is in releasing. `posthog-android` re-exports core via `api(project(":posthog"))` and pins the core version in its published POM, so shipping core alone would leave Android consumers on the old core. That's what forced the re-pin releases in #611 and #625. Bumped it as a minor rather than a patch since the event shape changes for Android users.

Left `posthog-server` out: its `PostHogFeatureFlags` doesn't override `isMinimalFlagCalledEventsEnabled()`, so it keeps the `false` default and nothing changes for it.

dustinbyrne (approved):

PR #642, opened 2026-07-20
feat: add push notification support for PostHog Workflows

ioannisj:

Adds push notification support for Android (iOS is #729).

What's added:

- `registerPushNotificationToken(deviceToken, appId)` on `PostHogInterface`. When `firebase-messaging` is on the classpath and `capturePushNotificationSubscriptions` is enabled, tokens are picked up from Firebase and registered at startup, so most apps don't wire anything.
- `unregisterPushNotificationToken()` on `PostHogInterface`. On `reset()` (logout), the token is unregistered for the old identity and re-registered under the new anonymous id, so notifications stop following a logged-out user.
- `capturePushNotificationOpened(title, body, payload, action)`. Cold-start taps on a tray notification are auto-detected from the launch intent when `capturePushNotificationOpened` is enabled; the manual call covers the paths auto-detection can't see (foreground data messages, warm-start `onNewIntent`). Each key of `payload["posthog"]` is attached as a `$notification_<key>` property.
- `PostHogPushSubscriptionManager` persists the latest token, re-sends it against the current `distinctId`, retries transient failures with backoff (honoring `Retry-After`), re-registers on `identify`, and stops after `optOut`.
- `firebase-messaging` is a `compileOnly` dependency, so the SDK degrades gracefully when Firebase isn't on the classpath (log line, no request, no crash).

Unregister sends `DELETE /api/push_subscriptions`, reusing the register `POST`'s path and five-field body so the backend can `$unset` the `$device_push_subscription_<app_id>` property. The backend `DELETE` route lands in PostHog/posthog#72490, which needs to merge first; until then the client ships ahead of it. An unregister is best-effort (single attempt, no retry or persistence), and any non-2xx is logged and ignored, so nothing breaks before it lands.

Unit tests for the new surface: `PostHogPushSubscriptionManagerTest`, `PostHogApiTest`, `PostHogPushSubscriptionIntegrationTest`, and `PostHogActivityLifecycleCallbackIntegrationTest`, covering payload shape, backoff and recovery, `Retry-After`, latest-wins, re-register on `identify`, opt-out, cold-start open dedup, the `DELETE` request shape, and the `reset()` move (unregister the old identity then re-register under the new anonymous id, and no unregister when `reuseAnonymousId` keeps the same id).

Mock-server E2E: the sample app on an emulator against a local mock PostHog server (no real backend, no Firebase project), covering registration backoff cadence, `Retry-After`, 400-vs-500 retry behavior, latest-wins, opt-out, the cold-start tray-tap open path with `message_id` dedup, the manual open API, flag-off, and Firebase-absent graceful skip.

Registration, real FCM delivery, and open capture were also validated end to end on an emulator against a live PostHog cloud project and Workflow. The unregister path is covered by unit and mock-server tests only, since the backend `DELETE` isn't implemented yet.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj


posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
```diff
@@ -0,0 +1,409 @@
...
+            isSending.set(false)
+        }
+    }
+
+    private fun handleFailure(
+        e: Throwable,
+        record: PendingRecord,
+    ) {
+        if (!isRetryable(e)) {
+            // 400/401 etc.: stop retrying this session but keep the record for one retry next launch.
```
turnipdabeets, line 302: [suggestion] Nothing sets a suppression flag here, so every `flush()` → `retryPending()` re-POSTs the known-bad request for the rest of the session — the comment promises we stop. iOS sets a `halted` flag on this path (no in-session retries, one retry next launch); could we do the same?

posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
```diff
@@ -55,6 +63,44 @@ internal class PostHogActivityLifecycleCallbackIntegration(
...
+     * warm-start `onNewIntent` and foreground data messages need the manual API. Dedupes on message id
+     * so an Activity recreation doesn't double-capture.
+     */
+    private fun capturePushNotificationOpenedIfNeeded(activity: Activity) {
+        val intent = activity.intent ?: return
+        // Reading extras unmarshals the whole Bundle; a launch intent carrying a Serializable/Parcelable
+        // extra whose class isn't on this app's classloader throws BadParcelableException here. This runs
+        // inside the framework onActivityCreated callback, so an uncaught throw crashes the host app.
+        try {
+            val messageId = intent.getStringExtra(GOOGLE_MESSAGE_ID) ?: return
```
turnipdabeets, line 80: [question] Keying on `google.message_id` alone means taps on any FCM notification (Braze, OneSignal, an app's own backend) emit `$push_notification_opened`, not just Workflows sends. Is capture-all deliberate? If not, requiring the `posthog` extra would scope it — either way worth a doc note, since it shifts Workflows open metrics for multi-provider apps.

posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
```diff
@@ -55,6 +63,44 @@ internal class PostHogActivityLifecycleCallbackIntegration(
...
+
+            if (messageId == lastHandledPushMessageId) {
+                return
+            }
+            lastHandledPushMessageId = messageId
```
turnipdabeets, line 85: [nit] The dedup is in-memory, so relaunching from recents after process death redelivers the launch intent with the same `google.message_id` and re-captures the open. Persisting the last handled id would close it.

posthog/src/main/java/com/posthog/PostHogInterface.kt
```diff
@@ -412,6 +412,57 @@ public interface PostHogInterface : PostHogCoreInterface {
...
+        appId: String,
+    )
+
+    /**
+     * Unregisters this device's push token from PostHog so Workflows stop targeting it — for example
+     * from your logout flow.
+     *
+     * Sends a best-effort `DELETE /api/push_subscriptions` for the current [distinctId] (the backend
+     * unsets the subscription property) and forgets the locally stored token. Unlike registration
+     * this is not retried. It is called for you on [reset] when `capturePushNotificationSubscriptions`
```
turnipdabeets: [nit] `handleReset` runs whenever a stored record exists — the core module can't see this Android-layer flag, so a manually-registered token with the flag off still moves (which seems right). Could we reword to "when a token was registered"? Same sentence exists in the iOS doc comment.

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
ioannisj: Done in 98984fdc — went with the `halted` flag like iOS. After a non-retryable failure (or once retries are exhausted) it stops retrying for the session and keeps the record for one retry on next launch. Guarded at the single `attempt()` choke point, so `flush()`-driven `retryPending()` is a no-op now. Added tests for the 400 path and the exhausted-retries one. Thanks for the iOS pointer, made this easy.

posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
ioannisj: Yes symmetric with iOS. Android's google.message_id gate is the exact analogue of iOS's UNPushNotificationTrigger gate and both mean "remote push only, any provider." Apps could use these events for analytics regardless of using workflow or not? I think this is a product decision @PostHog/team-workflows ?

ioannisj:
@turnipdabeets should be ready for another review run. This https://github.com/PostHog/posthog-android/pull/642#discussion_r3629925861 still not addressed, waiting for team reply but since this is now base on a feature branch should be safe to postpone before release.

turnipdabeets (approved):

PR #650, opened 2026-07-22
fix(error-tracking): install autocapture handler by default on first launch

### LINKED ISSUE #648: Bug: error tracking autocapture misses crashes on first launch before remote config resolves
### Version

3.55.2 (also reproduces on current `main`)

### Steps to Reproduce

1. Integrate posthog-android with error tracking autocapture enabled locally (`config.errorTrackingConfig.autoCapture = true`) and enabled in the project settings (remote `autocaptureExceptions`).
2. Install the app for the **first time** (no prior disk-cached remote config).
3. Trigger an uncaught exception very early on first launch, before the remote config (`/flags`) response arrives.
4. Relaunch and observe the crash is **not** reported in PostHog Error Tracking.
5. Trigger the same crash on a **later launch** (cache now present) and observe it **is** reported.

### Expected Result

On first launch, the `UncaughtExceptionHandler` should install by default when local `errorTrackingConfig.autoCapture` is on and nothing in the cached or loaded remote config explicitly disables it. That way a crash early on the first launch, before the remote config fetch finishes, still gets captured. If the config then comes back with `autocaptureExceptions: false`, uninstall the handler. Same shape as the iOS fix in PostHog/posthog-ios#731 (issue PostHog/posthog-ios#551).

### Actual Result

The integration installs synchronously during `setup()`, before the async remote config fetch finishes. `install()` bails unless the remote config already says autocapture is enabled. On a first launch there is no disk cache, so `isAutocaptureExceptionsEnabled()` is `false` and `install()` returns without registering the handler. Any uncaught exception between app start and the remote config response is lost.

This is not a permanent miss like the iOS bug (PostHog/posthog-ios#551). Once the config resolves in the same session, `onRemoteConfig(loaded = true)` retries the install and the handler comes up, so later crashes in that session are caught. The gap is only the first-launch window before the first config response.

A couple of related cases:

- if the first-launch config fetch fails, `onRemoteConfig(loaded = false)` returns early, so autocapture never installs for that whole session (offline first launch).
- the same window applies the first time autocapture gets turned on server-side, until the config lands and is cached.

<details>
<summary><h3>Code References</h3></summary>

Pinned to `92fa733`.

1. **Integration registered, then `install()` called synchronously during setup.** [PostHog.kt#L256](https://github.com/PostHog/posthog-android/blob/92fa7331a97a6d45391b33d05d7096e6a622f650/posthog/src/main/java/com/posthog/PostHog.kt#L256) registers it; [PostHog.kt#L282](https://github.com/PostHog/posthog-android/blob/92fa7331a97a6d45391b33d05d7096e6a622f650/posthog/src/main/java/com/posthog/PostHog.kt#L282) runs the install loop.

2. **`install()` bails unless remote config already reports enabled.** [PostHogErrorTrackingAutoCaptureIntegration.kt#L38-L40](https://github.com/PostHog/posthog-android/blob/92fa7331a97a6d45391b33d05d7096e6a622f650/posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt#L38-L40):
   ```kotlin
   val autocaptureExceptionsEnabled = config.remoteConfigHolder?.isAutocaptureExceptionsEnabled() ?: false
   if (!autocaptureExceptionsEnabled) {
       return
   }
   ```

3. **Only the disk cache is available at that point.** `preloadErrorTrackingConfig()` reads the cached value in the `PostHogRemoteConfig` `init` block ([PostHogRemoteConfig.kt#L552](https://github.com/PostHog/posthog-android/blob/92fa7331a97a6d45391b33d05d7096e6a622f650/posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt#L552)); `isAutocaptureExceptionsEnabled()` is `autoCaptureExceptions && errorTrackingConfig.autoCapture` ([PostHogRemoteConfig.kt#L635](https://github.com/PostHog/posthog-android/blob/92fa7331a97a6d45391b33d05d7096e6a622f650/posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt#L635)). On first launch there is no cache, so it is `false`.

4. **The remote config fetch is async, fired after the install loop.** [PostHog.kt#L321](https://github.com/PostHog/posthog-android/blob/92fa7331a97a6d45391b33d05d7096e6a622f650/posthog/src/main/java/com/posthog/PostHog.kt#L321) `loadRemoteConfigRequest(...)` runs on `remoteConfigExecutor` after `install()` has already returned.

5. **Late install works, but only after the fetch.** `onRemoteConfigLoaded` calls `notifyIntegrationsRemoteConfig(config, loaded = true/false)` ([PostHog.kt#L201-L203](https://github.com/PostHog/posthog-android/blob/92fa7331a97a6d45391b33d05d7096e6a622f650/posthog/src/main/java/com/posthog/PostHog.kt#L201-L203)), which iterates integrations calling `onRemoteConfig(loaded)` ([PostHog.kt#L342-L346](https://github.com/PostHog/posthog-android/blob/92fa7331a97a6d45391b33d05d7096e6a622f650/posthog/src/main/java/com/posthog/PostHog.kt#L342-L346)); the integration then installs iff enabled ([PostHogErrorTrackingAutoCaptureIntegration.kt#L71](https://github.com/PostHog/posthog-android/blob/92fa7331a97a6d45391b33d05d7096e6a622f650/posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt#L71)). A `loaded = false` (failed fetch) is a no-op.

</details>

### Suggested Fix

Mirror the iOS approach in PostHog/posthog-ios#731: in `install()`, install the handler by default when local `errorTrackingConfig.autoCapture` is on and there is no cached-or-loaded remote config that explicitly disables it, instead of waiting for an affirmative enable. Keep `onRemoteConfig(loaded = true)` to `uninstall()` if the resolved config comes back with `autocaptureExceptions: false`. That catches the first-launch pre-config crashes while still honoring the remote kill-switch, with local config as the primary gate.

This reaches Flutter too: the Flutter Android plugin maps `captureNativeExceptions` onto `errorTrackingConfig.autoCapture`, so a Flutter-on-Android app has the same first-launch gap.

ioannisj:

Fixes #648.

On the very first app launch there's no disk-cached remote config yet, so the native crash handler (`PostHogErrorTrackingAutoCaptureIntegration`, a `Thread.UncaughtExceptionHandler`) didn't install until the async `/flags` remote-config response arrived. Any uncaught exception in that window, from app start to the first config response, was lost. Offline made it worse: a failed first-launch fetch left autocapture off for the whole session.

This flips the gate to match the iOS fix. The handler now installs by default at setup when local `errorTrackingConfig.autoCapture` is on, and only skips when a remote config that actually exists (fetched this session or disk-cached from a prior launch) explicitly reports `autocaptureExceptions: false`. Remote config becomes a kill-switch rather than a prerequisite: if the freshly-loaded config disables autocapture, `onRemoteConfig` uninstalls the handler, and a failed fetch leaves the default install in place. Local `autoCapture` stays the primary gate, so a host that turned it off still gets no handler.

To tell "first launch, no config yet" apart from "config exists and says disabled", `PostHogRemoteConfig` gains `hasCachedErrorTrackingConfig()`, set at startup when the error-tracking config was already on disk. It's the Android analogue of iOS's `hasCachedRemoteConfig`.

This is the Android counterpart of iOS #551, ported from iOS #731. It also covers Flutter-on-Android, which maps `captureNativeExceptions` onto `errorTrackingConfig.autoCapture` and inherited the same gap.

Unit tests mirroring the iOS installation tests, in `PostHogErrorTrackingAutoCaptureIntegrationTest`:

- installs by default on first launch, before remote config arrives
- does not install when local `autoCapture` is off, even with no remote config
- does not install when a cached or fetched remote config disables autocapture
- installs when a cached remote config enables autocapture
- `onRemoteConfig` uninstalls the default install when a freshly-loaded config disables autocapture, and keeps it when the config enables autocapture
- a failed fetch (`loaded = false`) keeps the default install

Ran `./gradlew spotlessApply`, `:posthog:apiCheck`, and the full `:posthog:test` suite, all green.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj


.changeset/error-tracking-first-launch-crashes.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-android": patch
```
turnipdabeets, line 3: The code is all under `posthog/`, which changesets knows as `posthog` — and `posthog-android` needs a bump too, or its POM keeps pinning the old core and consumers can't reach the fix without an explicit override. The workspace `package.json` files declare no deps, so `updateInternalDependencies` can't cascade it for us — that's what #611 had to clean up after the last core-only error-tracking changeset.

```suggestion
"posthog": patch
"posthog-android": patch
```

.changeset/error-tracking-first-launch-crashes.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Error tracking autocapture now installs the uncaught-exception handler on the very first app launch, before the remote config (`/flags`) response arrives, so an uncaught exception in that startup window is no longer missed. Local `errorTrackingConfig.autoCapture` stays the primary gate; remote config acts only as a kill-switch, uninstalling the handler if the resolved config reports `autocaptureExceptions: false`. Ports the iOS fix (#731 / #551).
```
turnipdabeets: [nit] `#731 / #551` are posthog-ios numbers but read as Android PRs in our changelog — could we drop them? Same for `mirrors iOS #551/#731` in the new test section header.

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
```diff
@@ -69,7 +80,9 @@ public class PostHogErrorTrackingAutoCaptureIntegration : PostHogIntegration, Th
     }
 
     override fun onRemoteConfig(loaded: Boolean) {
-        // Only react to a live config; a failed attempt applies no fresh values.
+        // Only react to a live config; a failed attempt applies no fresh values, so leave the
```
turnipdabeets, line 142: When this lands disabled it calls `uninstall()`, which restores `defaultExceptionHandler` without checking the current global default is still us. Default-on makes install→uninstall a routine first-launch path now, so a handler installed between `setup()` and this callback gets replaced by the pre-PostHog one — confirmed with a fake adapter: after `onRemoteConfig` the app's handler is gone. We need to guard the restore:

```kotlin
if (adapterExceptionHandler.getDefaultUncaughtExceptionHandler() === this) {
    adapterExceptionHandler.setDefaultUncaughtExceptionHandler(defaultExceptionHandler)
}
```

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
```diff
@@ -519,19 +524,19 @@ public class PostHogRemoteConfig(
...
             is Boolean -> {
-                // if errorTracking is a Boolean, it's always false (disabled)
-                clearErrorTracking()
+                // errorTracking as a Boolean is always false (disabled). Persist the disabled
+                // stance instead of evicting the key, so the next launch skips the default
+                // first-launch install before its live /config lands (mirrors iOS).
+                autoCaptureExceptions = false
+                if (persist) {
+                    config.cachePreferences?.setValue(ERROR_TRACKING, mapOf("autocaptureExceptions" to false))
+                }
```
turnipdabeets: [suggestion] Persisting on any Boolean means an `errorTracking: true` response would cache "disabled" permanently, so every later launch skips the default install — #648 back for good. `processSessionRecordingConfig` just above already distinguishes the two, so guarding the write keeps them consistent.

```suggestion
                // errorTracking as a Boolean means disabled. Persist the disabled stance
                // instead of evicting the key, so the next launch skips the default
                // first-launch install before its live /config lands.
                autoCaptureExceptions = false
                if (persist && !errorTracking) {
                    config.cachePreferences?.setValue(ERROR_TRACKING, mapOf("autocaptureExceptions" to false))
                }
```

posthog/src/test/java/com/posthog/internal/PostHogRemoteConfigTest.kt
```diff
@@ -690,8 +690,10 @@ internal class PostHogRemoteConfigTest {
         executor.shutdownAndAwaitTermination()
 
         assertFalse(sut.isAutocaptureExceptionsEnabled())
-        // Must evict the cache, else a later reset()+reload could re-arm a disabled project.
-        assertNull(preferences.getValue(ERROR_TRACKING))
+        // Persist the disabled stance (not evict), so the next launch knows the project
+        // disabled autocapture before /config lands and skips the default first-launch install.
+        // Overwrites the stale enabled cache, so a later reset()+reload can't re-arm it.
+        assertEquals(mapOf("autocaptureExceptions" to false), preferences.getValue(ERROR_TRACKING))
```
turnipdabeets, line 696: [suggestion] This asserts the write but not the read-back the fix exists for — `hasCachedErrorTrackingConfig()` is only ever a Mockito stub today, so a break in either half would still go green.

```kotlin
@Test
fun `a returning launch sees the cached disabled stance before its own config lands`() {
    val disabled = File("src/test/resources/json/basic-remote-config-features-disabled.json").readText()
    val http = mockHttp(response = MockResponse().setBody(disabled))
    val sut = getSut(host = http.url("/").toString())
    config!!.errorTrackingConfig.autoCapture = true
    sut.loadRemoteConfig("my_identify", anonymousId = "anonId", emptyMap())
    executor.shutdownAndAwaitTermination()
    sut.clear()

    // Fresh process over the same on-disk preferences.
    val next = getSut(host = http.url("/").toString())
    config!!.errorTrackingConfig.autoCapture = true
    assertTrue(next.hasCachedErrorTrackingConfig())
    assertFalse(next.isAutocaptureExceptionsEnabled())

    http.shutdown()
}
```

turnipdabeets (commented):
LGTM, left some comments. Do you think we need to address RN as well?

.changeset/error-tracking-first-launch-crashes.md
ioannisj: Added `posthog: patch`, kept `posthog-android` for the POM-pin reason you flagged. 03ffbdf8

.changeset/error-tracking-first-launch-crashes.md
ioannisj: Dropped the iOS PR numbers from the changelog line and the test section header. 03ffbdf8

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
ioannisj: Added the ownership guard, only restore the handler when we're still the active one. 03ffbdf8

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
ioannisj: Guarded the cache write with `!errorTracking` so a truthy value can't persist the disabled stance. 03ffbdf8

posthog/src/test/java/com/posthog/internal/PostHogRemoteConfigTest.kt
ioannisj: Added the read-back test, fresh process over the same on-disk prefs, asserts cached + disabled. 03ffbdf8

ioannisj:
> LGTM, left some comments. Do you think we need to address RN as well?

Yeah and we'd probably have to gate flutter (dart side) exceptions as well I think

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
```diff
@@ -63,13 +74,19 @@ public class PostHogErrorTrackingAutoCaptureIntegration : PostHogIntegration, Th
         if (!integrationInstalled) {
             return
         }
-        adapterExceptionHandler.setDefaultUncaughtExceptionHandler(defaultExceptionHandler)
+        // Only restore if we are still the active handler. Something installed after us (or another
+        // PostHog instance that owns the process-wide flag) must not be replaced by our saved handler.
+        if (adapterExceptionHandler.getDefaultUncaughtExceptionHandler() === this) {
+            adapterExceptionHandler.setDefaultUncaughtExceptionHandler(defaultExceptionHandler)
+        }
         integrationInstalled = false
```
turnipdabeets: The guard does fix the clobbering — confirmed with a fake adapter, the app's handler now survives. But when the restore is skipped we still set `integrationInstalled = false` while we're actually still linked as that handler's delegate, and two things follow from it.

`uncaughtException` has no disabled check, so we keep capturing `$exception` for a project that just turned autocapture off — the kill-switch stops killing. And because the flag says uninstalled, a later `onRemoteConfig` re-enable runs `install()` again and sets `defaultExceptionHandler` to the app handler that delegates back to us; one crash then cycles (my probe re-entered the chain 21 times before I capped it, so in production that's a StackOverflowError inside the crash handler).

I think we need to separate "are we linked" from "are we capturing":

```kotlin
captureEnabled = false
if (adapterExceptionHandler.getDefaultUncaughtExceptionHandler() === this) {
    adapterExceptionHandler.setDefaultUncaughtExceptionHandler(defaultExceptionHandler)
    integrationInstalled = false // only uninstalled once we're actually out of the chain
}
```

with `uncaughtException` bailing when `captureEnabled` is false, and `install()` flipping it back on for the already-installed case.

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
ioannisj: Yeah you're right on both, the flag was saying uninstalled while we were still linked as the delegate.

Went with your split: added a `captureEnabled` separate from `integrationInstalled`.

- `uncaughtException` skips capture when `captureEnabled` is false but still delegates down-chain
- `uninstall` always sets `captureEnabled` false, and only clears `integrationInstalled` when it actually unlinks (`=== this`)
- a mid-chain re-enable just flips `captureEnabled` back on instead of re-linking, so the loop can't form

Added tests for the dormant-but-delegating case and the re-enable-without-relink one. Fixed in 8348351e.

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
ioannisj: Yeah you're right on both, the flag was saying uninstalled while we were still linked as the delegate.

Went with your split: added a `captureEnabled` separate from `integrationInstalled`.

- `uncaughtException` skips capture when `captureEnabled` is false but still delegates down-chain
- `uninstall` always sets `captureEnabled` false, and only clears `integrationInstalled` when it actually unlinks (`=== this`)
- a mid-chain re-enable just flips `captureEnabled` back on instead of re-linking, so the loop can't form

Added tests for the dormant-but-delegating case and the re-enable-without-relink one. Fixed in 8348351e.

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
```diff
@@ -95,11 +139,14 @@ public class PostHogErrorTrackingAutoCaptureIntegration : PostHogIntegration, Th
...
-            postHog.flush()
+        if (captureEnabled) {
+            postHog?.let { postHog ->
+                postHog.captureException(PostHogThrowable(throwable, thread))
+                postHog.flush()
+            }
         }
 
+        // Always delegate: we may still be mid-chain even while dormant.
         defaultExceptionHandler?.uncaughtException(thread, throwable)
```
turnipdabeets, line 168: [nit] `captureEnabled` is `@Volatile`, but `defaultExceptionHandler` isn't — and this reads it from whatever thread is crashing, while `install()`/`uninstall()` write it under the instance monitor. A thread that already existed when `setup()` ran has no happens-before edge with those writes, so it can read `null` here and skip delegating, meaning the app's handler and the system default never run for that crash. The new `defaultExceptionHandler = null` on the unlink path widens the window a little. I think this field wants `@Volatile` too.

Worth saying I couldn't reproduce this — JMM visibility races aren't reliably testable — so it's from reading the code, not a failing test.

turnipdabeets (approved):
looks good, reviewhog has some good comments as well and then I think it’s ready!

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
ioannisj: Good catch, made it `@Volatile`. Same crash-thread read as `captureEnabled` so it should match. Pushed in 1520c715.

PR #652, opened 2026-07-23
fix(replay): handle null bitmap drawables

marandaneto:

Session replay could crash while traversing an `ImageView` backed by a `BitmapDrawable` whose bitmap is null. Android allows `BitmapDrawable.getBitmap()` to return null, but the masking check called `Bitmap.isValid()` without guarding the platform type.

This change treats a null bitmap as non-maskable, preventing the uncaught `NullPointerException` while preserving existing behavior for valid and recycled bitmaps. It also removes the repository guidance that instructed agents to add a `release` label to release-bound PRs.

- `./gradlew :posthog-android:testDebugUnitTest --tests 'com.posthog.android.replay.PostHogReplayIntegrationTest'`
- `make checkFormat`
- Added a Robolectric regression test using a null-backed `BitmapDrawable`.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi coding agent investigated the crash, implemented the null guard and regression test, updated the repository guidance as directed, and prepared the PR. A read-only reviewer subagent independently checked correctness, test quality, API implications, and scope and reported no actionable findings. The masking helper was changed from private to Kotlin `internal` so the regression test can call the production logic directly without reflection; this does not add a published API.

### DIFF
--- a/.changeset/calm-bitmaps-rest.md
+++ b/.changeset/calm-bitmaps-rest.md
@@ -1,0 +1,5 @@
+---
+"posthog-android": patch
+---
+
+Prevent session replay from crashing when an image uses a `BitmapDrawable` with a null bitmap.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -59,8 +59,6 @@ pnpm changeset
 
 This will prompt for the affected packages, semver bump type, and a summary of changes. Commit the generated `.changeset/*.md` file with the PR.
 
-When the PR is merged with a `release` label, changesets will automatically bump versions, publish packages, and create GitHub releases.
-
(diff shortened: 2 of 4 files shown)


marandaneto:
https://play.google.com/sdk-console/u/1/accounts/3843563740275146028/sdks/7065738917375733979/crashes/6123fbce/details?timeRange=LAST_28_DAYS&affectedApps=AFFECTED_APPS_UNSPECIFIED&crashStatuses=SDK_CRASH_STATUS_UNSPECIFIED&errorTypes=CRASH

ioannisj (approved):

PR #655, opened 2026-07-27
perf(android): avoid screen and device-type IPC candidates

marandaneto:

Screen autocapture queried `ActivityInfo` through `PackageManager` on every `Activity.onStart`, and device-type detection queried `UiModeManager` while building static context. Both paths can run on the main thread and can be served from state already loaded in the process. Screen autocapture now uses the activity's current title, so apps that set titles dynamically may emit a different `$screen_name` for `$screen` events and change report grouping; this is called out in the changeset.

The network-time candidate was also reviewed. Android already caches the synchronized network-time sample, and reading it does not perform an NTP request. This PR therefore keeps the simpler existing once-per-minute synchronous cache and documents that its occasional cached-value IPC is accepted.

- `make test`
- `make checkFormat`
- Focused unit tests for screen-name compatibility without `getActivityInfo` and API 28 device-type compatibility

An API 28 Perfetto Binder trace was considered, but no Android device or emulator was connected locally, so this PR does not claim runtime trace verification.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent using repository file, shell, Git, Gradle, and GitHub CLI tools. No shareable agent session link was generated. Activity and UI-mode queries use already-loaded local state instead of adding callback/cache complexity. The proposed asynchronous network-time refresh was rolled back after reviewing AOSP: the OS already caches network time, and the existing once-per-minute synchronous read is the simpler accepted tradeoff. `TelephonyManager.networkOperatorName` remains unchanged because the callback alternative would require location permission.

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidUtils.kt
+++ b/posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidUtils.kt
@@ -133,21 +133,20 @@ internal fun Context.telephonyManager(): TelephonyManager? {
     return getSystemService(Context.TELEPHONY_SERVICE) as? TelephonyManager
 }
 
-@Suppress("DEPRECATION")
 internal fun Activity.activityLabelOrName(config: PostHogAndroidConfig): String? {
     return try {
-        val activityInfo = packageManager.getActivityInfo(componentName, GET_META_DATA)
-        val activityLabel = activityInfo.loadLabel(packageManager).toString()
+        val activityLabel = title?.toString().orEmpty()
+        val activityName = componentName.className
         val applicationLabel = applicationInfo.loadLabel(packageManager).toString()
 
         if (activityLabel.isNotEmpty() && activityLabel != applicationLabel) {
-            if (activityLabel == activityInfo.name) {
+            if (activityLabel == activityName) {
                 activityLabel.substringAfterLast('.')
             } else {
                 activityLabel
             }
         } else {
-            activityInfo.name.substringAfterLast('.')
+            localClassName.substringAfterLast('.')
         }
     } catch (e: Throwable) {
         config.logger.log("Error getting the Activity's label or name: $e.")
--- a/.changeset/calm-clocks-rest.md
+++ b/.changeset/calm-clocks-rest.md
@@ -2,4 +2,4 @@
 "posthog-android": patch
 ---
 
-Avoid remaining main-thread IPC candidates in network time, screen autocapture, and device type detection.
+Avoid main-thread IPC candidates in screen autocapture and device type detection.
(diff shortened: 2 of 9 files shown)


posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidUtils.kt
```diff
@@ -133,21 +133,20 @@ internal fun Context.telephonyManager(): TelephonyManager? {
     return getSystemService(Context.TELEPHONY_SERVICE) as? TelephonyManager
 }
 
-@Suppress("DEPRECATION")
 internal fun Activity.activityLabelOrName(config: PostHogAndroidConfig): String? {
     return try {
-        val activityInfo = packageManager.getActivityInfo(componentName, GET_META_DATA)
-        val activityLabel = activityInfo.loadLabel(packageManager).toString()
+        val activityLabel = title?.toString().orEmpty()
```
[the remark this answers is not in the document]
marandaneto: Thanks for flagging this. The behavior is intentional: screen autocapture should reflect the Activity’s current title, including a title set before `onActivityStarted`. We reviewed the compatibility difference and prefer the runtime title for this use case, so we’re keeping the current implementation.

ioannisj, line 138: On this, what about apps that set a dynamic title (e.g Inbox (5)). Wouldn't they see a behavior change on their analytics?

ioannisj (approved):
LG Left a discussion point on greptile thread

posthog-android/src/main/java/com/posthog/android/internal/PostHogAndroidUtils.kt
marandaneto: see https://github.com/PostHog/posthog-android/pull/655#discussion_r3655185599
yes its gonna be a behavior change, i'd say this would be even a bug since thats how i thought it used to be, but if people rely on this, indeed it can break, not sure then, its a trade off, perf vs correctness, opinions?

ioannisj: I think I would keep it since it's a more "correct" capture and dynamic titles do carry more context as well. Maybe drop a warning on changelog that this could potentially affect some reports?

marandaneto: Good call. Added the warning in d3dce1ea: the changeset now notes that dynamic activity titles may produce a different `$screen_name` for `$screen` events and affect existing reports. I also updated the PR description.

PR #656, opened 2026-07-27
feat: attach optional identity token to push subscription requests

ioannisj:

PostHog/posthog#72488 (draft) adds opt-in identity verification to `/api/push_subscriptions`: your backend mints a short-lived HS256 JWT (claims `sub` = distinct_id, `app_id`, `aud` = `posthog:push_identity`, `exp`) signed with the project's secret API key, and the endpoint verifies it on both POST and DELETE. Modes per integration: `disabled` (default), `optional`, `required`. Without SDK support there's no way to attach the token, so `required` mode can't be used. The provider shape was agreed with Dan in #team-messaging.

This adds the SDK half:

- New `pushIdentityProvider: (distinctId, appId, completion) -> Unit` constructor param on `PostHogConfig` (core module, no Firebase dependency). The app passes a backend-minted token (or null) to `completion`, from any thread, exactly once.
- The token goes out as `identity_token` on the register POST and unregister DELETE. When absent the key is omitted entirely, so the body stays byte-identical to today's 5-field contract and nothing changes for apps that don't opt in.
- Minting is on demand, not per attempt: the token is cached in memory per exact `(distinctId, appId)`, reused across in-session backoff retries, and re-minted on identity change, on the next launch, and once after a 401.
- A 401 with a provider drops the cache, re-mints, and retries once; a second 401 halts for the session (record kept, next launch retries). Without a provider it's terminal, with a debug log pointing at the config hook.
- Register and unregister share one verification state: the `reset()` DELETE resolves a token for the old identity, the re-register POST for the new anonymous id.
- Opt-out clears the cached token, and the send guards are re-checked after the async mint.

Mirrors PostHog/posthog-ios#735. Field names, provider invocation counts, and 401 semantics are kept identical between the two SDKs.

Encoded the shared cross-SDK vectors 9-14 in the suite (token on both legs, key omission when absent, reset-leg tokens and invocation counts, 500-retry cache reuse, 401 refresh then terminal, 401 without a provider), plus throwing-provider, exactly-once, and foreign-thread completion tests. `make testJava` green (857 tests), `make api` dumps regenerated, ktlint clean.

## :pencil: Checklist

### If releasing new changes

- [ ] Ran `pnpm changeset` to generate a changeset file (deliberately omitted: feature branch, not releasing yet)

## 🤖 Agent context

**DRI:** @ioannisj

### DIFF
--- a/posthog/src/main/java/com/posthog/PostHogConfig.kt
+++ b/posthog/src/main/java/com/posthog/PostHogConfig.kt
@@ -328,6 +328,22 @@ public open class PostHogConfig(
      * Defaults to null (no bootstrap).
      */
     public var bootstrap: PostHogBootstrapConfig? = null,
+    /**
+     * Hook that supplies a signed identity token for push subscription requests, enabling the
+     * optional identity verification of the push subscriptions API.
+     *
+     * Invoked with the distinctId and appId the request is about to carry, whenever a token is
+     * needed and none is cached for that exact pair — and once more after a 401 rejection. Your
+     * backend mints the token — an HS256 JWT signed with the project's secret API key (never
+     * embedded in the app), with claims `sub` = distinct id, `app_id`, `aud` =
+     * "posthog:push_identity", and an `exp` — and you pass it to `completion`, from any thread. Pass `completion(null)`
+     * when no token can be minted (e.g. the user is anonymous); the request is then sent without
+     * one. `completion` must be called exactly once: further calls are ignored, and never calling
+     * it strands the current registration until the next app launch.
+     *
+     * Defaults to null (requests are sent without an identity token).
+     */
+    public var pushIdentityProvider: ((distinctId: String, appId: String, completion: (String?) -> Unit) -> Unit)? = null,
 ) {
     @Volatile
     private var tracingHeadersList: List<String>? = null
--- a/posthog/src/main/java/com/posthog/PostHogConfig.kt
+++ b/posthog/src/main/java/com/posthog/PostHogConfig.kt
@@ -341,6 +341,10 @@ public open class PostHogConfig(
      * one. `completion` must be called exactly once: further calls are ignored, and never calling
      * it strands the current registration until the next app launch.
      *
+     * The hook itself runs on the SDK's push executor thread (only `completion` may be called from
+     * any thread), so return quickly: do the token minting asynchronously and call `completion` when
+     * it finishes. Blocking here stalls the push retry/offline-resume loop.
+     *
      * Defaults to null (requests are sent without an identity token).
      */
     public var pushIdentityProvider: ((distinctId: String, appId: String, completion: (String?) -> Unit) -> Unit)? = null,
--- a/posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
@@ -450,7 +450,12 @@ internal class PostHogPushSubscriptionManager(
                 if (completed.compareAndSet(false, true)) {
                     executor.executeSafely {
                         if (token != null) {
-                            cachedIdentityToken = CachedIdentityToken(token, distinctId, appId)
+                            // A mint can complete after opt-out cleared the cache; caching it would
+                            // resurrect a stale credential on a later opt-in. The 401 refresh covers
+                            // the residual race window.
+                            if (!closed && !config.optOut) {
+                                cachedIdentityToken = CachedIdentityToken(token, distinctId, appId)
+                            }
                             config.logger.log("Push subscription request sent with freshly minted identity token.")
                         } else {
                             config.logger.log("Push subscription request sent without identity token (provider completed null).")
--- a/posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
@@ -88,7 +88,7 @@ internal class PostHogPushSubscriptionManager(
 
     // Watchdog window for pushIdentityProvider: if the host never calls completion within this, fall
     // back to a token-less send so a misbehaving provider can't wedge sending for the whole process.
-    // ponytail: fixed 10s heuristic; a slow legitimate mint on a bad network is cut off and retried
+    // Fixed 10s heuristic; a slow legitimate mint on a bad network is cut off and retried
     // token-less (401 re-mints). Tune here (and keep parity with iOS) if that proves too tight.
     internal var identityTokenMintTimeoutMillis: Long = 10_000L
 
@@ -282,7 +282,15 @@ internal class PostHogPushSubscriptionManager(
     /** Opt-out: stop the retry/offline-poll timer now. The guard in [attempt] blocks any actual send. */
     fun onOptOut() {
         cancelTimer()
-        cachedIdentityToken = null
+        // Order both clears on the executor with the mint-completion cache write so an opt-out
+        // mid-mint can't leave a stale token cached (the residual race resolveIdentityToken notes).
+        // didAuthRetry is cleared too: a 401 before opt-out would otherwise strand the flag, and the
+        // retryPending() resume path after opt-in doesn't clear it, so the next 401 goes terminal
+        // with no refresh.
+        executor.executeSafely {
+            cachedIdentityToken = null
+            didAuthRetry = false
+        }
     }
 
     private fun isWithinBackoffWindow(): Boolean = System.currentTimeMillis() < nextAttemptAtMs
--- a/posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
@@ -61,6 +61,15 @@ internal class PostHogPushSubscriptionManager(
 
     @Volatile private var closed = false
 
+    // In-memory only: a short-lived credential must never land on disk; a fresh process re-mints
+    // via config.pushIdentityProvider. Reused across in-session backoff retries for the same
+    // (distinctId, appId); any other pair is a miss and re-mints.
+    @Volatile private var cachedIdentityToken: CachedIdentityToken? = null
+
+    // One fresh-token retry per send-cycle after a 401 (single refresh, no loop). Cleared on
+    // success, a new register, and an identity-change resend; a fresh process starts clear.
+    @Volatile private var didAuthRetry = false
+
     private val pendingFile: File? by lazy {
         val prefix = config.storagePrefix ?: return@lazy null
         // Must stay out of <storagePrefix>/<apiKey>: PostHogQueue scans that whole directory as
@@ -107,6 +116,7 @@ internal class PostHogPushSubscriptionManager(
         retryCount = 0
         nextAttemptAtMs = 0L
         halted = false
+        didAuthRetry = false
         cancelTimer()
         attempt()
     }
@@ -141,6 +151,7 @@ internal class PostHogPushSubscriptionManager(
             }
             retryCount = 0
             halted = false
+            didAuthRetry = false
             cancelTimer()
             attempt()
         }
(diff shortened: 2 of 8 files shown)


posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
```diff
@@ -276,16 +297,36 @@ internal class PostHogPushSubscriptionManager(
...
                 appId = record.appId,
+                identityToken = identityToken,
             )
             config.logger.log("Push notification token registered successfully.")
             retryCount = 0
             nextAttemptAtMs = 0L
+            didAuthRetry = false
             // Keep the record with the delivered marker so a later identify() can re-register.
             val delivered = record.copy(deliveredForDistinctId = distinctId)
             pendingRecord = delivered
```
turnipdabeets, line 372: With an async provider the captured `record` can be stale by the time the mint lands, but we write it back unconditionally. Confirmed with a failing test: `unregisterCurrent()` during an in-flight mint sends the DELETE, then the late mint re-POSTs and rewrites `push_subscription.pending` — the user stays subscribed, and permanently, since the resurrected record is stamped delivered. I think we need to re-read `currentRecord()` after the mint and bail if it's gone or no longer matches this record, the way iOS `markDelivered` does.

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
```diff
@@ -276,16 +297,36 @@ internal class PostHogPushSubscriptionManager(
             return
         }
 
+        // isSending stays claimed across an async token mint so resume paths can't double-send;
+        // [performSend] releases it on every path.
+        resolveIdentityToken(distinctId, record.appId) { identityToken ->
```
turnipdabeets, line 335: `isSending` was never contended before this — `attempt()` held the executor thread for the whole blocking send. Now the thread is free during the mint, so a `register()` or `resendIfDistinctIdChanged()` landing mid-mint hits the CAS above and returns silently. Confirmed with a failing test: an FCM token rotation during a mint never reaches the wire and never recovers in-session, because `retryPending()` early-returns on the delivered record. I think we need an iOS-style `pendingResend` that replays the dropped request when this send completes.

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
```diff
@@ -176,16 +187,24 @@ internal class PostHogPushSubscriptionManager(
...
-                deviceToken = deviceToken,
-                platform = platform,
-                appId = appId,
-            )
-            config.logger.log("Push notification token unregistered successfully.")
-        } catch (e: Throwable) {
-            config.logger.log("Push unregister failed: $e. Ignoring (best-effort).")
+        // Best-effort stays single-shot: the token is resolved once (old id on the reset path) and
+        // a 401 is not refreshed — the durable path is the re-register POST.
+        resolveIdentityToken(distinctId, appId) { identityToken ->
```
turnipdabeets, line 207: This makes the DELETE async, so `handleReset`'s "keeping the DELETE ordered before the re-register POST" no longer holds — `performUnregister` returns as soon as the mint is dispatched and `performRegister` runs immediately. Confirmed with a failing test: on a cold cache (the ordinary second-launch path — `retryPending()` early-returns, so nothing mints before logout) the POST for `anon-2` goes out before the DELETE for `user-B`, and by the reasoning at L220-223 that trailing DELETE unsets the subscription we just created. I think we need to keep the legs ordered — resolve both tokens up front, or chain the re-register on the DELETE's completion.

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
```diff
@@ -360,6 +424,54 @@ internal class PostHogPushSubscriptionManager(
...
+    // Resolves the identity token for [distinctId]/[appId], preferring a cached exact match. The
+    // provider's completion may arrive from any thread and only the first call counts; no provider,
+    // a null completion, or a throw all fall back to token-less — the pre-identity behavior.
+    private fun resolveIdentityToken(
+        distinctId: String,
+        appId: String,
+        onResolved: (String?) -> Unit,
+    ) {
+        val cached = cachedIdentityToken
+        if (cached != null && cached.distinctId == distinctId && cached.appId == appId) {
```
turnipdabeets: [question] iOS checks `pushIdentityProvider` before the cache (`PostHogPushSubscriptionHandler.swift:304-318`), so an app that sets the provider back to `null` mid-session stops attaching tokens there but keeps sending the cached one here. Deliberate, or worth flipping the order to match?

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
```diff
@@ -360,6 +424,54 @@ internal class PostHogPushSubscriptionManager(
...
+            config.logger.log("Push subscription request sent with cached identity token.")
```
turnipdabeets: [nit] This logs "request sent" before anything is sent, and it also fires on the DELETE path — so an unregister logs "Push subscription request sent with cached identity token". (iOS has the same wording, so no divergence, just confusing in support logs.)

posthog/src/main/java/com/posthog/PostHogConfig.kt
```diff
@@ -328,6 +328,22 @@ public open class PostHogConfig(
...
     public var bootstrap: PostHogBootstrapConfig? = null,
+    /**
+     * Hook that supplies a signed identity token for push subscription requests, enabling the
+     * optional identity verification of the push subscriptions API.
+     *
+     * Invoked with the distinctId and appId the request is about to carry, whenever a token is
+     * needed and none is cached for that exact pair — and once more after a 401 rejection. Your
+     * backend mints the token — an HS256 JWT signed with the project's secret API key (never
+     * embedded in the app), with claims `sub` = distinct id, `app_id`, `aud` =
+     * "posthog:push_identity", and an `exp` — and you pass it to `completion`, from any thread. Pass `completion(null)`
```
turnipdabeets, line 339: [nit] Worth also noting that the provider itself is invoked on the SDK's push executor thread — a blocking mint stalls the retry/offline-poll loop. "from any thread" only describes `completion`, so that isn't obvious.

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
ioannisj: Fixed in abe5ebec. `performSend` re-reads `currentRecord()` after the mint now and bails if the record's gone or the token/appId no longer matches, so a late mint can't re-POST a subscription that `unregisterCurrent()` already DELETEd. Added a regression test for the register-mid-mint case.

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
ioannisj: Fixed in abe5ebec with an iOS-style `pendingResend`. An attempt that lands on the held `isSending` sets the flag instead of dropping now, and `performSend` replays one fresh attempt on release. Same test covers the FCM-rotation-mid-mint path.

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
ioannisj: Fixed in abe5ebec. `performUnregister` takes an `onComplete` now and `handleReset` chains the re-register on the DELETE's completion, so the DELETE hits the wire before the POST even with async minting. Fixed the KDoc that claimed the single executor task guaranteed ordering too.

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
ioannisj: Good call, flipped it to match iOS in abe5ebec. The provider is checked before the cache now, so clearing `pushIdentityProvider` mid-session stops attaching tokens right away instead of riding the cached one. posthog-js has no push feature so iOS is the reference here.

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
ioannisj: Reworded in abe5ebec. The four lines read `Attaching … identity token to push request` / `No identity token attached …` now instead of the premature, DELETE-wrong `request sent`. Mirrored the same strings on iOS in posthog-ios#735 so they stay in parity.

posthog/src/main/java/com/posthog/PostHogConfig.kt
ioannisj: Added in abe5ebec. The KDoc notes the hook runs on the push executor thread now (only `completion` is any-thread) and to mint asynchronously. Same note on iOS.

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
```diff
@@ -235,11 +276,13 @@ internal class PostHogPushSubscriptionManager(
         closed = true
         cancelTimer()
         retryCount = 0
+        cachedIdentityToken = null
     }
 
     /** Opt-out: stop the retry/offline-poll timer now. The guard in [attempt] blocks any actual send. */
     fun onOptOut() {
         cancelTimer()
+        cachedIdentityToken = null
```
turnipdabeets: [suggestion] iOS just added `didAuthRetry = false` here too (posthog-ios 1e9522a4). Same gap applies: a 401 sets `didAuthRetry`, the queued retry hits the opt-out guard in `attempt()` and returns without clearing it, so after opt-in the next 401 goes terminal with no refresh. `performRegister` clears it, but the `retryPending()` path doesn't.
```suggestion
        cachedIdentityToken = null
        didAuthRetry = false
```

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
```diff
@@ -360,6 +476,71 @@ internal class PostHogPushSubscriptionManager(
...
+        try {
+            provider(distinctId, appId) { token ->
+                if (completed.compareAndSet(false, true)) {
+                    watchdog.cancel()
+                    executor.executeSafely {
+                        if (token != null) {
+                            // A mint can complete after opt-out cleared the cache; caching it would
+                            // resurrect a stale credential on a later opt-in. The 401 refresh covers
+                            // the residual race window.
+                            if (!closed && !config.optOut) {
```
turnipdabeets, line 532: [suggestion] `onOptOut()` runs on the caller's thread (`PostHog.optOut()` calls it directly), while this write is on the push executor — so an opt-out landing between this check and the write still leaves a stale token cached, which is the case the comment above calls the "residual race window". iOS closed it by sampling opt-out under the same lock that clears the cache (1e9522a4). I think the cheap equivalent here is routing the clear through the executor so the two are ordered on one thread — `onOptOut()` keeps `cancelTimer()` inline and does `executor.executeSafely { cachedIdentityToken = null }`.

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
```diff
@@ -72,6 +86,12 @@ internal class PostHogPushSubscriptionManager(
     // millis. Production keeps the real 1000; tests shrink it so retries fire near-instantly.
     internal var retryDelayMillisPerSecond: Long = 1_000L
 
+    // Watchdog window for pushIdentityProvider: if the host never calls completion within this, fall
+    // back to a token-less send so a misbehaving provider can't wedge sending for the whole process.
+    // ponytail: fixed 10s heuristic; a slow legitimate mint on a bad network is cut off and retried
```
turnipdabeets: [nit] `ponytail:` looks like a leftover authoring marker — it's the only occurrence in the repo. Worth dropping the prefix and keeping the rest of the note, which is genuinely useful.

turnipdabeets (approved):

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
ioannisj: Fixed in 4408f611. `onOptOut` clears `didAuthRetry` alongside the cache now, so the `retryPending()` resume path after opt-in still gets the 401 refresh instead of going terminal.

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
ioannisj: Fixed in 4408f611. `onOptOut` routes the clear through the executor now so it's ordered with the mint-completion cache write, which closes the stale-token race. Folded `didAuthRetry` into the same block.

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
ioannisj: Dropped the prefix in 4408f611, kept the note.

PR #659, opened 2026-07-28
feat(errortracking): capture native ndk crashes from tombstones

cat-ph:

Native (NDK) crashes kill the process before any JVM handler runs, so today they are invisible to error tracking. This adds capture for them without shipping any native code in the SDK: on startup, the SDK reads the crash records the OS kept via `ApplicationExitInfo` (`REASON_CRASH_NATIVE`, Android 12+), parses the attached tombstone protobuf, and captures one `$exception` event per crash using the native stack frame contract PostHog already resolves for the Rust and Go SDKs.

Per crash, the event carries:

- raw native frames (`platform: "native"`, `instruction_addr`, `image_addr`, optional client-resolved `function`/`symbol_addr`) in canonical bottom-up order
- `$debug_images` entries derived from the tombstone's per-frame GNU build ids, so the server matches frames to `.so` symbols uploaded with `posthog-cli symbol-sets upload`
- signal metadata (`SIGSEGV` / `SEGV_MAPERR at 0x...`, abort message when present), `$exception_level: fatal`, and the original crash timestamp

Design notes:

- Tombstone parsing is a minimal hand-rolled protobuf wire reader (field numbers are frozen in AOSP), avoiding a protobuf runtime dependency. Unknown fields are skipped.
- `instruction_addr` is biased by +1: tombstone pcs are already the correct lookup address (the leaf is the faulting instruction and libunwindstack rewinds caller pcs to the call instruction), so the bias cancels the server's uniform -1 return-address adjustment. This is pinned by a cymbal fixture test on the server side.
- Opt-in via `errorTrackingConfig.captureNativeCrashes`, additionally gated on the project's exception autocapture remote toggle. A dedicated, synchronous timestamp watermark prevents duplicate capture across launches and survives identity resets, advancing per record so dying mid-scan cannot re-capture.
- Frames matching the app's native library, base/split APK, or data paths are in-app; OS and unknown mappings are not. The package-name process scans the package-wide exit history, preventing duplicate capture in typical multi-process apps. Apps that override the application-level `android:process` still need handling before merge.

Known limitations (draft): events are associated with the identity at next launch, not at crash time; `$exception_steps` recorded in the new process may attach to the previous run's crash; API 31+ only (tombstone protos attach from Android 12).

- Unit tests for the tombstone parser (against independently hand-encoded proto wire bytes) and the event coercer (frame shape, address math, debug-id derivation pinned to the same vocabulary as the server fixture and CLI).
- Full gradle test suite on a remote linux box; the new tests pass. On the latest head, 17 focused parser/coercer/scanner tests cover tied timestamps, I/O retry versus deterministic parse skip, accurate captured/skipped counts, fresh-executor re-enable, the process guard and main-process gate, omitted ARM32 architecture, one debug image per mapped base, and app-path `in_app` classification. The tied-timestamp regression test fails when transplanted onto the previous head and passes on this head. Pre-existing `PostHogAndroidTest` conscrypt failures on that box reproduce on `main` (arm64 environment issue, no conscrypt aarch64 linux artifact) and are unrelated.
- Latest-head device/devbox E2E (`c93f163` with #660 `ecc097a`): installed a debug build on an API 35 ARM64 emulator, triggered a real SIGSEGV (`ApplicationExitInfo` reason 5/status 11), then cold-relaunched. The SDK queued exactly one `$exception`; another relaunch and an identity reset produced no duplicate.
- The event persisted once in ClickHouse with the exact uploaded ELF debug reference. Cymbal resolved the JNI, C++, and inline frames at lines 19, 12, and 8 with bundled `latest-v5` source context. Only an actual Play Console upload/distribution remains untested.

## :pencil: Checklist

- [ ] I updated the docs if needed. (docs PR is staged separately, gated on release)

### If releasing new changes


posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
```diff
@@ -646,6 +646,13 @@ public class PostHogRemoteConfig(
      */
     public fun isAutocaptureExceptionsEnabled(): Boolean = autoCaptureExceptions && config.errorTrackingConfig.autoCapture
 
+    /**
+     * Returns whether native (NDK) crash capture is enabled.
+     * Both remote config (errorTracking.autocaptureExceptions) AND local config
```
marandaneto, line 651: does this exist on remote config already?

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,136 @@
...
+        } catch (e: Throwable) {
+            config.logger.log("Native crash scan failed: $e.")
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.S)
+    private fun scan(postHog: PostHogInterface) {
+        val activityManager =
+            context.getSystemService(Context.ACTIVITY_SERVICE) as? ActivityManager ?: return
+        val preferences = config.cachePreferences ?: return
```
marandaneto: we can move this to PostHogAndroidUtils.kt i think

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,136 @@
...
+            return
+        }
+        // While the store is unreadable (Direct Boot) the watermark reads as
+        // absent, which would re-capture already-reported crashes.
+        if (config.cachePreferences?.isAvailable() == false) {
+            return
+        }
+        integrationInstalled = true
+
+        Thread({ scanSafely(postHog) }, "PostHogNativeCrashScanner")
```
marandaneto: we can probably use a single thread executor instead, and we'd like to be able to pass the executor as ctor param so we can unit test

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,136 @@
...
+                } catch (e: Throwable) {
+                    config.logger.log("Tombstone parsing failed: $e.")
+                    null
+                }
+
+            properties?.let {
+                postHog.capture(
+                    PostHogEventName.EXCEPTION.event,
+                    properties = it,
+                    timestamp = Date(exitInfo.timestamp),
```
marandaneto, line 243: is this also UTC?

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,136 @@
...
+                    exitInfo.traceInputStream?.use { stream ->
+                        coercer.toPostHogProperties(parser.parse(stream))
+                    }
+                } catch (e: Throwable) {
+                    config.logger.log("Tombstone parsing failed: $e.")
+                    null
+                }
+
+            properties?.let {
+                postHog.capture(
```
marandaneto, line 240: we should attach the cached properties as we do for error tracking hard crashes (iOS)
not a blocker for this PR, but an improvement we'd need to make at some point to avoid issues with error investigation
eg sdk version 1.0.0, app version 2.0.0, but the app got upgraded after sending this error, customers would be investigating the wrong version eg 2.0.1 instead
this should be written down in the docs/config since its an important caveat that can waste lots of hours of investigation

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingConfig.kt
```diff
@@ -45,4 +45,18 @@ public class PostHogErrorTrackingConfig
...
+         * native crash records the OS kept for the app and captures an `$exception`
+         * event per crash, with raw stack frames for server-side symbolication
+         * against uploaded `.so` debug symbols.
+         *
+         * Exception autocapture must also be enabled in the project's error tracking
+         * settings (the same remote toggle that gates [autoCapture]).
+         *
+         * Disabled by default
+         */
+        public var captureNativeCrashes: Boolean = false,
```
marandaneto: blocking: Adding this defaulted parameter to the public Kotlin primary constructor changes/removes the synthetic default-argument constructor. `@JvmOverloads` preserves the ordinary overloads, but not that synthetic descriptor, so existing Kotlin callers that omit any constructor argument can fail at runtime with `NoSuchMethodError` after upgrading. I reproduced this by compiling `PostHogErrorTrackingConfig(inAppIncludes = mutableListOf("com.example"))` against `main` and running it against this branch. Could we keep the primary constructor unchanged and declare `captureNativeCrashes` as a property in the class body instead?

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,136 @@
...
+ * autocapture must also be enabled in the project settings.
+ */
+public class PostHogNativeCrashIntegration(
+    private val context: Context,
+    private val config: PostHogAndroidConfig,
+) : PostHogIntegration {
+    private var postHog: PostHogInterface? = null
+
+    private companion object {
+        private const val LAST_CAPTURED_TIMESTAMP_KEY = "nativeCrashLastCapturedTimestamp"
```
marandaneto: blocking: This watermark needs to be treated as SDK-internal, project-level state. Because the key is not included in `PostHogPreferences.ALL_INTERNAL_KEYS`, `PostHogSharedPreferences.getAll()` returns it as a registered property and `buildProperties()` attaches `nativeCrashLastCapturedTimestamp` to subsequent customer events. It is also absent from `PostHog.reset()`'s preserved keys, so reset/logout clears the deduplication cursor and the same retained tombstones are captured again after restart.

There is also a durability issue: `setValue()` ultimately uses `SharedPreferences.Editor.apply()`, which updates memory synchronously but writes to disk asynchronously without reporting failures. An abrupt process death before that write completes can lose the cursor and recapture an already queued crash on the next launch. Could we add the key to `ALL_INTERNAL_KEYS`, preserve it across `reset()`, and persist this marker synchronously in project-scoped storage before treating the record as acknowledged?

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,136 @@
...
+    @RequiresApi(Build.VERSION_CODES.S)
+    private fun scan(postHog: PostHogInterface) {
+        val activityManager =
+            context.getSystemService(Context.ACTIVITY_SERVICE) as? ActivityManager ?: return
+        val preferences = config.cachePreferences ?: return
+
+        val watermark = preferences.getValue(LAST_CAPTURED_TIMESTAMP_KEY) as? Long ?: 0L
+        val crashes =
+            activityManager
+                .getHistoricalProcessExitReasons(context.packageName, 0, MAX_EXIT_RECORDS)
```
marandaneto: blocking: Android applies `maxNum` before this code filters by exit reason. If a native crash is older than 20 newer non-native exits, it is excluded from every scan and will never be reported. Since the OS history is already a bounded ring buffer and `maxNum = 0` means all retained matching records, could we request all records here and filter by reason and watermark afterward?

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,136 @@
...
+                postHog.capture(
+                    PostHogEventName.EXCEPTION.event,
+                    properties = it,
+                    timestamp = Date(exitInfo.timestamp),
+                )
+            }
+
+            // Advance per record — unparsable ones too, retrying can't succeed —
+            // so dying mid-scan never re-captures already-reported crashes.
+            preferences.setValue(LAST_CAPTURED_TIMESTAMP_KEY, exitInfo.timestamp)
```
marandaneto: blocking: This advances the watermark without knowing whether `capture()` durably queued the event. `capture()` returns `Unit`, and queue submission/storage failures are swallowed, so a rejected executor, serialization/disk failure, or `close()` disabling the client while this scanner is still running can drop the event while permanently marking the crash as handled. Could we advance the watermark only after acknowledged durable queue persistence, and ensure uninstall/close cancels or waits for the scanner so it cannot acknowledge a dropped capture?

marandaneto:
A few additional follow-ups from the review:

- Debug images are currently deduplicated only by build/debug ID. If the same ELF is loaded at multiple base addresses, only one corresponding image entry survives and frames from the other base may not symbolicate. Please key these by `(debugId, imageAddr)` and cover the multiple-base case.
- Treating every `/data/` mapping as `in_app` is too broad. Please classify frames using app-specific native library, data, base APK, and split APK paths instead.
- ARM32 is enum value zero and may be omitted by proto3 as the default value. In that case the parser currently leaves `arch` unset; initialize it to `"arm"` and add coverage for an omitted architecture field.
- The final log reports every matching record as “Captured”, including records whose trace was missing or whose tombstone failed to parse. Please report captured/skipped counts accurately.
- Please add integration-level coverage around the scanner itself, especially history limits, reset/watermark behavior, internal-property leakage, close races, queue persistence failure, and concurrent installation.

marandaneto:
Two additional behavioral points to address before a stable release:

- Recovered tombstones are captured in a new process, so the current process's `$exception_steps` did not lead to the native crash and would be misleading in the resulting issue. Please suppress current-run exception steps for these events. Current identity/app-version enrichment can remain a documented limitation if avoiding it requires substantially more state.
- Please define and enforce a historical reporting policy. Reporting all unseen retained crashes is reasonable, but without an age limit, enabling the feature can emit arbitrarily old crashes with current-process metadata. A bounded lookback would make the behavior predictable and avoid misleading historical events.

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,136 @@
...
+        private const val MAX_EXIT_RECORDS = 20
+
+        @Volatile
+        private var integrationInstalled = false
+    }
+
+    override fun install(postHog: PostHogInterface) {
+        this.postHog = postHog
+
+        if (integrationInstalled || Build.VERSION.SDK_INT < Build.VERSION_CODES.S) {
```
marandaneto: blocking: `@Volatile` only guarantees visibility; the check followed by assignment is a compound operation and is not atomic, so concurrent installs can both observe `false` and start duplicate scanners. `uninstall()` also clears the process-wide flag from any integration instance, even one that did not acquire it, which can allow another scanner to start while the original is still active. Could we use atomic acquisition (for example `AtomicBoolean.compareAndSet`), track ownership per integration instance, and release the guard only from the owner after its scanner has terminated?

marandaneto:
left a few comments @cat-ph its in the right direction

marandaneto:
you can also check the https://github.com/abovevacant/epitaph TombstoneDecoder impl for parsing the exit metadata

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
cat-ph: this reuses autocaptureExceptions which is there on the remote config, but not a separate crash capture config; `captureNativeCrashes` is local

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
cat-ph: yep, should be! epoch millis -> UTC

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
marandaneto: yeah i meant the remote config API (/$token/config), sry, but it can be totally local as well

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
cat-ph: yeah, great catch; I added the separate own posthog-native-crash SharedPreferences file (NativeCrashWatermarkStore), written with commit()

_however_ I didn't do the durable capture() because I think right now it's fire-and-forget and we don't really ack that 🤔 I would've rathered duplicated than losing silently, I think we need to do some more work to have a ack way

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
cat-ph: I think this was copied from `PostHogErrorTrackingAutoCaptureIntegration` and might happen there too, but fixed it here and we can follow-up

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
cat-ph: yep good point, updated

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
cat-ph: yep, updated!

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
cat-ph: I have a draft docs PR and will mention it there: https://github.com/PostHog/posthog.com/pull/19035

I definitely agree and we can follow-up separately IMO, it _probably_ needs its own design 🤔 (I can open an issue and take a look after)

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingConfig.kt
cat-ph: ty for the reproduction and good catch! updated it

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
cat-ph: combined the reply in this one https://github.com/PostHog/posthog-android/pull/659#discussion_r3729711103 😆 but updated with a `NativeCrashWatermarkStore` SharedPreferences file, I _think_ that should work

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
cat-ph: ooh good point, added maxNum = 0 and filtering!

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,187 @@
...
+        val watermarkStore = NativeCrashWatermarkStore(context)
+        val watermark = watermarkStore.get()
+
+        // maxNum 0 returns every retained record. A positive cap is applied
+        // before the reason filter, so enough newer non-crash exits would
+        // starve an older native crash out of every scan.
+        val crashes =
+            activityManager
+                .getHistoricalProcessExitReasons(context.packageName, 0, 0)
+                .filter { it.reason == ApplicationExitInfo.REASON_CRASH_NATIVE && it.timestamp > watermark }
```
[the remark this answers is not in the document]
cat-ph: Good catch, and it is reachable: getHistoricalProcessExitReasons spans every process of the package, so sibling processes can die in the same millisecond. Fixed in 2ef1a51: the watermark now advances only past the last record of each distinct timestamp, so a scan dying mid-group re-captures the whole group instead of orphaning the tail (at-least-once, consistent with the durable-ack thread). Covered by the tied-timestamp test in d126335.

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,187 @@
...
+ * uploaded `.so` debug symbols.
+ *
+ * Requires Android 12 (API 31), where native crash records carry a tombstone.
+ * Enable with `errorTrackingConfig.captureNativeCrashes`; error tracking
+ * autocapture must also be enabled in the project settings.
+ */
+public class PostHogNativeCrashIntegration : PostHogIntegration {
+    private val context: Context
+    private val config: PostHogAndroidConfig
+    private val executor: ExecutorService
```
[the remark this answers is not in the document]
cat-ph: Skipping this one for consistency: no integration carries an internal marker today (PostHogErrorTrackingAutoCaptureIntegration, PostHogReplayIntegration and the rest are all public), wrapper SDKs construct integrations directly, and config.addIntegration is user-facing API. If we want SDK-managed integrations annotation-gated, that should be a deliberate sweep across all of them rather than only the new one.

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,187 @@
...
+                    timestamp = Date(exitInfo.timestamp),
+                )
+            }
+
+            // Advance only after capture returned, one record at a time:
+            // at-least-once. Losing a crash to an unacknowledged queue is worse
+            // than a rare duplicate from dying between the capture and this
+            // synchronous write.
+            watermarkStore.advance(exitInfo.timestamp)
+        }
```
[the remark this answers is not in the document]
cat-ph: Added in d126335: Robolectric tests (ApplicationExitInfoBuilder + setTraceInputStream) cover reason filtering, capture with the tombstone-derived timestamp, watermark acknowledgement and no-reprocessing, records without a trace, and a failed capture keeping tied records unacknowledged for the next scan.

posthog-android/src/main/java/com/posthog/android/PostHogAndroid.kt
```diff
@@ -183,6 +184,9 @@ public class PostHogAndroid private constructor() {
             if (config.capturePushNotificationSubscriptions) {
                 config.addIntegration(PostHogPushSubscriptionIntegration(config))
             }
+            if (config.errorTrackingConfig.captureNativeCrashes) {
+                config.addIntegration(PostHogNativeCrashIntegration(context, config))
+            }
```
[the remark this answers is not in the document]
cat-ph: Added in d126335: PostHogAndroidTest now asserts the integration is registered when errorTrackingConfig.captureNativeCrashes is enabled and absent by default.

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,187 @@
...
+    override fun uninstall() {
+        if (!installedByThisInstance) {
+            return
+        }
+        // Stop the scanner before releasing the process-wide guard, so a new
+        // install cannot start a second scanner while this one is still running.
+        executor.shutdownNow()
+        try {
+            executor.awaitTermination(1, TimeUnit.SECONDS)
+        } catch (e: InterruptedException) {
```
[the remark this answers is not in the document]
cat-ph: Keeping this as is: uninstall() only runs via PostHog.close(), a rare and effectively terminal operation, and one second is an upper bound only approached when a scan is actively mid-record at close time (the scanner checks for interruption between records). Releasing the guard without awaiting termination would reopen the overlapping-scanner race that double-captures crash records, which is the worse trade.

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,187 @@
...
+    override fun onRemoteConfig(loaded: Boolean) {
+        // Only react to a live config; a failed attempt applies no fresh values.
+        if (!loaded) {
+            return
+        }
+        if (config.remoteConfigHolder?.isNativeCrashCaptureEnabled() == true) {
+            postHog?.let { install(it) }
+        }
+    }
+
```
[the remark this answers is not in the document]
cat-ph: Covered by the tests in d126335: a second install does not start another scanner, a non-owner uninstall does not release the process-wide guard, the owner's uninstall does, and the remote toggle gates scheduling.

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,194 @@
...
+        installedByThisInstance = false
+        integrationInstalled.set(false)
+    }
+
+    override fun onRemoteConfig(loaded: Boolean) {
+        // Only react to a live config; a failed attempt applies no fresh values.
+        if (!loaded) {
+            return
+        }
+        if (config.remoteConfigHolder?.isNativeCrashCaptureEnabled() == true) {
```
marandaneto, line 160: blocking: If cached configuration enables capture, a scan can still be running when a live response disables autocapture. This branch does nothing when `isNativeCrashCaptureEnabled()` becomes false, so a slow or multi-record scan continues after the kill switch is applied. Please cancel or abort the active scan on a fresh disable while preserving the ability to schedule again after a later re-enable.

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,194 @@
...
+        val activityManager = getActivityManager(context) ?: return
+        val watermarkStore = NativeCrashWatermarkStore(context)
+        val watermark = watermarkStore.get()
+
+        // maxNum 0 returns every retained record. A positive cap is applied
+        // before the reason filter, so enough newer non-crash exits would
+        // starve an older native crash out of every scan.
+        val crashes =
+            activityManager
+                .getHistoricalProcessExitReasons(context.packageName, 0, 0)
```
marandaneto, line 189: blocking: This query returns exit records for every process in the package (`pid = 0`), while `integrationInstalled` is only process-local. In a multi-process app where PostHog initializes in more than one process, each scanner can read the same watermark and capture the same records concurrently; `SharedPreferences` does not provide atomic cross-process coordination. Please restrict scanning to one process or add cross-process locking and coherent deduplication.

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,194 @@
...
+            if (Thread.currentThread().isInterrupted) {
+                return
+            }
+
+            val properties =
+                try {
+                    exitInfo.traceInputStream?.use { stream ->
+                        coercer.toPostHogProperties(parser.parse(stream))
+                    }
+                } catch (e: Throwable) {
```
marandaneto, line 234: blocking: This catch turns any tombstone read/parser failure into `null`, but the loop still advances the timestamp watermark below. A transient `IOException` or unsupported tombstone is therefore permanently lost; for tied timestamps, a later successful record acknowledges the failed sibling too. Please distinguish a documented missing trace from a parse failure and stop or propagate on the latter so it is not acknowledged.

marandaneto:
@cat-ph Could you take another look at the four follow-ups in [this review comment](https://github.com/PostHog/posthog-android/pull/659#issuecomment-5192183834)? They still appear to be present on the current head:

- debug images are still keyed only by debug ID rather than `(debugId, imageAddr)`
- every `/data/` mapping is still classified as `in_app`
- an omitted proto3 architecture still leaves ARM32 as `null`
- the final “Captured” count still includes missing or failed tombstones

Could you either address these or reply with the rationale and intended follow-up for anything being deferred?

marandaneto:
did another pass after commits, a few small things to be fixed but almost there

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
cat-ph: ty, fixed + added a test!

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
cat-ph: ah good point, updated to use the first option (restrict to one process)

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
cat-ph: soooo I tried to do that in c93f163, just that when a tombstone is fully read and doesn't parse, I ack'd it so it doesn't poison-pill ourselves (since future reads still won't parse)

cat-ph:
> the four follow-ups in https://github.com/PostHog/posthog-android/pull/659#issuecomment-5192183834?

I .. totally missed that reply sorry @marandaneto  🤦 addressed all 4! plus the others and small test fixes

marandaneto (approved):

marandaneto:
once this is released, its worth exposing the config to flutter and react native

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
cat-ph: ah actually you fixed it with https://github.com/PostHog/posthog-android/pull/663 just realized I was looking at an earlier version, so no follow-up needed 😁

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
cat-ph: updated docs, opened https://github.com/PostHog/posthog-android/issues/700 for follow-up

posthog-android/src/main/java/com/posthog/android/errortracking/PostHogNativeCrashIntegration.kt
```diff
@@ -0,0 +1,239 @@
...
+            }
+
+            // Reading the trace can fail transiently (I/O), so abort without
+            // acknowledging and retry on the next launch. A tombstone that
+            // read fully but does not parse is deterministic: acknowledge and
+            // skip it, because retrying forever would hold the watermark below
+            // it and block every newer crash behind it.
+            val tombstoneBytes =
+                try {
+                    exitInfo.traceInputStream?.use { stream -> stream.readBytes() }
```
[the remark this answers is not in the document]
cat-ph: true.. but they aren't really attacker-controlled, written by the OS, still I added a 16mb cap should be big enough while remaining under heap problems

PR #660, opened 2026-07-28
feat(gradle-plugin): upload native debug symbols via symbol-sets upload

cat-ph:

Native crash capture (PostHog/posthog-android#659) needs the app's `.so` debug symbols uploaded at build time, the same way the plugin already uploads ProGuard mappings. This registers an `uploadPostHogNativeSymbols<Variant>` task that points `posthog-cli symbol-sets upload` (>= 0.7.32) at the variant's merged native libs intermediates: the unstripped libraries as built, before AGP strips them for packaging. The CLI scans the directory, uploads every library carrying debug info and a GNU build id, and triages the rest with guidance.

Wiring choices:

- Registered for every variant, independent of minification: native symbolication is orthogonal to JVM obfuscation, so the task also exists for non-minified builds.
- Automatic upload is an explicit `posthog { uploadNativeSymbols.set(true) }` opt-in and skips debuggable variants; the per-variant task remains directly invocable. It scans the complete merged native library directory, covering NDK builds, `jniLibs`, and libraries from dependencies.
- Skipped via `onlyIf` when the variant has no `.so` files at all.
- Reuses the existing CLI discovery, credential, and dotenv plumbing (`PostHogCliExecTask`), so `tasks.withType<PostHogCliExecTask>` configuration applies to it like the mapping upload.

- All 12 committed functional cases pass against AGP 8.0.2 and 8.9.1, covering merged-library dependencies, opt-in/out behavior, source bundling, debuggable-variant exclusion, and configuration-cache smoke coverage. A corrected local release-hook configuration-cache scenario fails on the previous head (`962b0a1`) for both AGP versions because Gradle cannot serialize the captured task, and passes on `ecc097a`; a corrected real `assembleRelease` failure also skips upload on the current head.
- Latest-head device/devbox E2E: `assembleDebug` skipped automatic upload, while `bundleRelease` ran `uploadPostHogNativeSymbolsRelease` exactly once. A fresh ELF plus its exact DWARF source tree was uploaded with CLI 0.10.0; the symbol set was stored without failure and marked used by Cymbal.
- On an API 35 ARM64 emulator, a real SIGSEGV produced one distinct `$exception` carrying that debug reference. Cymbal resolved the JNI, C++, and inline frames at lines 19, 12, and 8 with bundled `latest-v5` source context. Only an actual Play Console upload/distribution remains untested.

## :pencil: Checklist

- [ ] I updated the docs if needed. (docs PR is staged separately, gated on release)

### If releasing new changes


marandaneto:
we should add a config to enable|disable this in PostHogCliExecTask (or PostHogUploadNativeSymbolsTask not sure) since debug symbols are quite heavy and not all apps (the majority) dont need this
also, we should also provide another config to upload debug symbols w or w/o sources
i think both should be disabled by default

marandaneto:
not sure if this task is under isDebuggable already but we should not upload for debug builds

a known gap is that we still wont be able to symbolicate OS frames, which is ok for now

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadNativeSymbolsTask.kt
```diff
@@ -0,0 +1,101 @@
...
+        // Uploads have no outputs; without this the task would always rerun.
+        outputs.upToDateWhen { true }
+    }
+
+    /**
+     * Root of the variant's merged native libs intermediates. Not tracked as a
+     * task input: uploads are gated by [outputs.upToDateWhen] and snapshotting
+     * every `.so` would only slow the build.
+     */
+    @get:Internal
```
marandaneto: This directory cannot be `@Internal` while the task opts into up-to-date checking with `outputs.upToDateWhen { true }`. The only remaining tracked inputs are release metadata, so rebuilding native code without changing the app version can make Gradle report this task as `UP-TO-DATE` even though the `.so` and its build ID changed. The new symbols would then never be uploaded.

Please either declare this as an `@InputDirectory` with `@PathSensitive(PathSensitivity.RELATIVE)` so changes to the native libraries invalidate the task, or remove the `outputs.upToDateWhen { true }` optimization so the upload runs whenever it is scheduled. `@DisableCachingByDefault` only disables the build cache; it does not disable local up-to-date checks.

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
```diff
@@ -65,6 +70,40 @@ internal class PostHogAndroidGradlePlugin : Plugin<Project> {
...
+    ) {
+        val primaryOutput = variant.outputs.firstOrNull()
+        val uploadTask =
+            PostHogUploadNativeSymbolsTask.register(
+                project = project,
+                // The unstripped libraries as built; AGP strips them for
+                // packaging in a later task. The subdirectory layout varies
+                // across AGP versions, so point at the variant root and let
+                // the CLI scan recursively.
+                nativeLibsDirectory =
```
marandaneto: Both `intermediates/merged_native_libs/<variant>` and `merge<Variant>NativeLibs` are AGP implementation details rather than stable public contracts, so an AGP update can silently move the directory or rename the producer task. `SingleArtifact.MERGED_NATIVE_LIBS` provides a public artifact/provider on newer AGP versions, but it is unavailable on our AGP 8.0.2 compile baseline, so retaining this fallback is understandable. Could we add functional compatibility tests covering at least AGP 8.0.2 and the current supported AGP version, asserting that explicit upload depends on the producer and receives the expected merged native-libs directory?

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
```diff
@@ -65,6 +70,40 @@ internal class PostHogAndroidGradlePlugin : Plugin<Project> {
...
+            )
+
+        project.afterEvaluate {
+            PostHogTasksProvider.getMergeNativeLibsTask(project, variant.name)?.let { merge ->
+                uploadTask.configure { dependsOn(merge) }
+            }
+            // Auto-upload alongside assemble/install/bundle only when the app
+            // module builds native code itself. Apps that only bundle prebuilt
+            // `.so` files from dependencies can run the task explicitly.
+            if (variant.externalNativeBuild != null) {
```
marandaneto: Once native-symbol upload is explicitly opt-in and disabled by default, do we still need this `externalNativeBuild != null` gate? It excludes apps that package native libraries through `src/main/jniLibs`, custom native-producing tasks/plugins, or dependencies, even though their symbols may be present in the merged directory. The opt-in itself should prevent surprising uploads for React Native and other apps, while `onlyIf` already handles variants with no `.so` files. Removing this gate after adding the opt-in would make the enabled behavior cover all native-library sources consistently.

marandaneto:
left a few comments, the overall approach does whats supposed to do, noice!

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadNativeSymbolsTask.kt
cat-ph: good point, updated to use `@InputDirectory` and `@PathSensitive(PathSensitivity.RELATIVE)`

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
cat-ph: yep, added!

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
cat-ph: good point, got rid of it + added `posthog { uploadNativeSymbols }` disabled by default plus includeNativeSymbolSources as a separate flag

cat-ph:
> not sure if this task is under isDebuggable already but we should not upload for debug builds

great catch, updated this too!

cat-ph:
thanks @marandaneto for the thorough review 💙 updated! 🏓

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
```diff
@@ -65,6 +75,56 @@ internal class PostHogAndroidGradlePlugin : Plugin<Project> {
...
+            // Explicit opt-in rather than any capability heuristic: the merged
+            // output can carry `.so` files from the NDK, jniLibs, or plain
+            // dependencies, and only the app author knows whether their symbols
+            // belong in PostHog. Debuggable variants are excluded from the
+            // automatic hook so day-to-day debug builds don't upload heavy,
+            // unoptimized symbol sets. The task stays invocable explicitly for
+            // any variant either way.
+            if (extension.uploadNativeSymbols.get() && !variant.isDebuggable(project)) {
+                uploadTask.hookWithAssembleTasks(project, variant)
+            }
```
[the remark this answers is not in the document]
cat-ph: Right, finalizedBy runs regardless of the finalized task's outcome. Fixed in 962b0a1: the hook records its anchor tasks and the upload skips via onlyIf when any executed anchor failed, so a failed build uploads nothing and its error cannot be obscured by upload output. This applies to the mapping upload too, since it shares hookWithAssembleTasks. Covered by a functional test forcing a post-merge build failure and asserting the CLI never runs.

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/TasksUtils.kt
```diff
@@ -53,16 +53,26 @@ internal fun TaskProvider<out Task>.hookWithAssembleTasks(
...
+        bundleTask?.also { anchors.add(it) }?.configure {
             finalizedBy(this@hookWithAssembleTasks)
         }
+        // Finalizers run even when the build they finalize fails; skip the
+        // upload then, so artifacts of a failed build are not uploaded and the
+        // upload's own errors cannot obscure the original failure. Explicit
+        // invocations are unaffected: an unexecuted anchor carries no failure.
+        this@hookWithAssembleTasks.configure {
+            onlyIf("the finalized build succeeded") {
+                anchors.none { anchor -> anchor.get().state.failure != null }
```
marandaneto: blocking: Preserve Gradle configuration-cache compatibility

This execution-time predicate closes over `anchors` and resolves each `TaskProvider` with `get()`, causing configuration-cache storage to fail with `cannot serialize object of type DefaultTask`. This affects opted-in native-symbol uploads and the shared mapping-upload hook. Please keep the condition provider-safe without resolving `Task` instances during execution and add configuration-cache coverage.

posthog-android-gradle-plugin/build.gradle.kts
```diff
@@ -192,3 +192,29 @@ dependencies {
...
+
+// Functional tests run the plugin through Gradle TestKit against real AGP
+// versions (the oldest supported and a current one), because the merged
+// native libs directory and its producer task are AGP internals that can
+// move between releases.
+val functionalTest: SourceSet by sourceSets.creating
+
+dependencies {
+    "functionalTestImplementation"(gradleTestKit())
+    "functionalTestImplementation"("org.jetbrains.kotlin:kotlin-test-junit:${versions["kotlinVersion"]}")
```
marandaneto, line 208: suggestion: Regenerate dependency locks

This adds `kotlin-test-junit` (plus JUnit and Hamcrest transitively) to the functional-test classpath, but `gradle.lockfile` has no corresponding entries despite `lockAllConfigurations()`. Please resolve the functional-test dependencies with `--write-locks` and commit the lockfile updates.

marandaneto (approved):
left one blocker for config cache otherwise lgtm

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/TasksUtils.kt
cat-ph: fixed, ty! added a `PostHogTaskFailureTracker`

posthog-android-gradle-plugin/build.gradle.kts
cat-ph: regenerated, soz I missed it 🙈

PR #661, opened 2026-07-28
fix(android): preserve manually set releaseIdentifier

cat-ph:

`PostHogMetaPropertiesApplier.applyToConfig` had an inverted guard: the comment said "if releaseIdentifier is already set, we don't need to do anything", but the condition overwrote a manually configured `releaseIdentifier` with the auto-generated fallback (`applicationId@versionName+versionCode`). Anyone setting `releaseIdentifier` in `PostHogAndroidConfig` (as documented) lost it on init, breaking proguard symbolication due to the map-id mismatch.

This fixes the precedence to:

1. A pre-set `releaseIdentifier` is preserved (the asset read is skipped entirely).
2. Otherwise the map-id from `posthog-meta.properties` is used when present.
3. Otherwise the fallback is applied — including the previously broken case where the meta file exists but contains no map-id, which used to leave `releaseIdentifier` null.

- New unit tests in `PostHogMetaPropertiesApplierTest` covering all five paths (preserve manual value, use meta map-id, fallback on missing file / missing key / read error).
- Full unit test gate (`./gradlew testDebugUnitTest`) and `spotlessCheck` pass.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/posthog-android/src/test/java/com/posthog/android/internal/PostHogMetaPropertiesApplierTest.kt
+++ b/posthog-android/src/test/java/com/posthog/android/internal/PostHogMetaPropertiesApplierTest.kt
@@ -1,0 +1,88 @@
+package com.posthog.android.internal
+
+import android.content.Context
+import android.content.res.AssetManager
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.posthog.android.API_KEY
+import com.posthog.android.PostHogAndroidConfig
+import org.junit.runner.RunWith
+import org.mockito.kotlin.any
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.never
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+import java.io.ByteArrayInputStream
+import java.io.FileNotFoundException
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+@RunWith(AndroidJUnit4::class)
+internal class PostHogMetaPropertiesApplierTest {
+    private val context = mock<Context>()
+    private val assets = mock<AssetManager>()
+    private val config = PostHogAndroidConfig(API_KEY)
+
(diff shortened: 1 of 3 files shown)


posthog-android/src/test/java/com/posthog/android/internal/PostHogMetaPropertiesApplierTest.kt
```diff
@@ -0,0 +1,88 @@
...
+        mockMissingMetaProperties()
+
+        sut.applyToConfig(context, config, FALLBACK)
+
+        assertEquals(FALLBACK, config.releaseIdentifier)
+    }
+
+    @Test
+    fun `uses fallback when meta properties do not contain map id`() {
+        mockMetaProperties("some.other.property=value")
```
cat-ph, line 68: this is actually a more important case I saw in prod, tens of thousands of frames per week here at a first/quick look (although some could be older versions who don't have a releaseIdentifier in the first place).

the first one with manual releaseIdentifier isn't *as* common

marandaneto:
will check, i remember testing this so quite surprise its a bug, maybe a regression

marandaneto:
Was it a regression?

 Your memory is correct:

 - The initial implementation in PR #316, commit 00d4e33b from November 5, correctly did:
   ```kotlin
     if (config.releaseIdentifier.isNullOrEmpty()) {
         config.releaseIdentifier = fallback
     }
   ```
 - The November 28 refactor in commit bd1373f1 introduced PostHogMetaPropertiesApplier and accidentally inverted that
   behavior.
 - It shipped in posthog-android 3.27.0.

 So it was a regression during development of PR #316, likely after you tested it. Strictly speaking, it was broken in
 the first stable release containing releaseIdentifier, so it was not a regression between stable SDK releases.

marandaneto (approved):

PR #663, opened 2026-07-28
fix: make integration installation atomic

### LINKED ISSUE #662: `PostHogSendCachedEventsIntegration`: `@Volatile` guard is non-atomic — two concurrent `install()` calls both pass, flushing legacy events twice
## Bug

In `posthog/src/main/java/com/posthog/internal/PostHogSendCachedEventsIntegration.kt` (around lines 24 and 28–31), `integrationInstalled` is declared `@Volatile`:

```kotlin
companion object {
    @Volatile
    private var integrationInstalled = false
}
```

The guard uses a plain read-then-write:

```kotlin
if (integrationInstalled) return
integrationInstalled = true
flushLegacyEvents(config)
```

`@Volatile` guarantees visibility across threads but **not atomicity of the read-modify-write**. Two threads calling `install()` concurrently (e.g. two `PostHog` instances initialised in parallel on separate threads, each with its own `setupLock`) can both read `false`, both pass the check, both set `true`, and both call `flushLegacyEvents()` — sending cached legacy events twice.

## Fix

Replace `@Volatile` with `AtomicBoolean` and use `compareAndSet` for an atomic check-and-set:

```kotlin
companion object {
    private val integrationInstalled = AtomicBoolean(false)
}

// In install():
if (!integrationInstalled.compareAndSet(false, true)) return
flushLegacyEvents(config)
```

## Impact

Under concurrent initialisation, legacy cached events are sent twice, causing duplicate events in PostHog — inflating event counts and distorting analytics.

marandaneto:

Concurrent setup of multiple PostHog instances can race through the `@Volatile` read-then-write guards used by default integrations. `@Volatile` guarantees visibility but not an atomic check-and-set, allowing duplicate legacy flushes, lifecycle callbacks, replay listeners, logcat threads, and other integration side effects.

Closes #662.

This also fixes the same pattern in the seven other affected integrations found during the repository-wide audit. Each integration tracks whether its instance won installation, preventing a losing instance from clearing the winner's shared state during teardown. Install and uninstall are synchronized per instance so remote-config updates cannot race with teardown side effects. Cached-event integrations that lose installation also shut down their unused executor.

- Added concurrent-install regression tests for legacy event flushing, executor shutdown, and Android lifecycle callback registration.
- Added coverage proving a non-owner teardown does not clear the winning installation.
- Added explicit test isolation for the shared cached-events installation guard.
- `JAVA_TOOL_OPTIONS=-Dnet.bytebuddy.experimental=true make testJava`
- `./gradlew :posthog:apiCheck`
- Focused Android unit tests for all affected Android integrations, including session replay.
- `make checkFormat`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent using repository inspection, shell, editing, and test tools. The audit covered every production `@Volatile` Boolean and limited the change to the eight `integrationInstalled` guards that perform a check-and-set. Review feedback led to owner-aware teardown and per-instance synchronization across those guards; other volatile fields remain unchanged because they are visibility-only or already lock/thread confined.

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
@@ -7,6 +7,7 @@ import android.os.Bundle
 import com.posthog.PostHogIntegration
 import com.posthog.PostHogInterface
 import com.posthog.android.PostHogAndroidConfig
+import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * Captures deep link and screen view events
@@ -20,8 +21,7 @@ internal class PostHogActivityLifecycleCallbackIntegration(
     private var postHog: PostHogInterface? = null
 
     private companion object {
-        @Volatile
-        private var integrationInstalled = false
+        private val integrationInstalled = AtomicBoolean(false)
     }
 
     override fun onActivityCreated(
@@ -84,10 +84,9 @@ internal class PostHogActivityLifecycleCallbackIntegration(
     }
 
     override fun install(postHog: PostHogInterface) {
-        if (integrationInstalled) {
+        if (!integrationInstalled.compareAndSet(false, true)) {
             return
         }
-        integrationInstalled = true
 
         this.postHog = postHog
         application.registerActivityLifecycleCallbacks(this)
@@ -95,7 +94,7 @@ internal class PostHogActivityLifecycleCallbackIntegration(
 
     override fun uninstall() {
         this.postHog = null
-        integrationInstalled = false
+        integrationInstalled.set(false)
         application.unregisterActivityLifecycleCallbacks(this)
     }
 }
--- a/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
@@ -19,6 +19,7 @@ internal class PostHogActivityLifecycleCallbackIntegration(
     private val config: PostHogAndroidConfig,
 ) : ActivityLifecycleCallbacks, PostHogIntegration {
     private var postHog: PostHogInterface? = null
+    private var ownsInstallation = false
 
     private companion object {
         private val integrationInstalled = AtomicBoolean(false)
@@ -83,18 +84,28 @@ internal class PostHogActivityLifecycleCallbackIntegration(
     override fun onActivityDestroyed(activity: Activity) {
     }
 
+    @Synchronized
     override fun install(postHog: PostHogInterface) {
         if (!integrationInstalled.compareAndSet(false, true)) {
             return
         }
+        ownsInstallation = true
 
         this.postHog = postHog
         application.registerActivityLifecycleCallbacks(this)
     }
 
+    @Synchronized
     override fun uninstall() {
-        this.postHog = null
-        integrationInstalled.set(false)
-        application.unregisterActivityLifecycleCallbacks(this)
+        if (!ownsInstallation) {
+            return
+        }
+        try {
+            this.postHog = null
+            application.unregisterActivityLifecycleCallbacks(this)
+        } finally {
+            ownsInstallation = false
+            integrationInstalled.set(false)
+        }
     }
 }
--- a/posthog-android/src/test/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegrationTest.kt
+++ b/posthog-android/src/test/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegrationTest.kt
@@ -13,7 +13,9 @@ import org.junit.runner.RunWith
 import org.mockito.kotlin.any
 import org.mockito.kotlin.mock
 import org.mockito.kotlin.never
+import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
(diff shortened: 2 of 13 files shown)


marandaneto:
`setup` calls `install` so likely not happening because setup is behind a lock, but its good to fix it anyway

posthog/src/test/java/com/posthog/internal/PostHogSendCachedEventsIntegrationTest.kt
```diff
@@ -55,6 +61,55 @@ internal class PostHogSendCachedEventsIntegrationTest {
...
+                    integration.install(postHog)
+                }.apply { start() }
+            }
+
+        ready.await()
+        start.countDown()
+        threads.forEach { it.join() }
+
+        assertEquals(1, scheduledFlushes.get())
+        integrations.forEach { it.uninstall() }
```
turnipdabeets: [nit] If the assertion above ever fails, we never reach the `uninstall()` loop and the shared companion flag stays `true`, so the remaining installs in this class silently no-op with confusing secondary failures. Could we put the assert in a `try/finally` so a future regression fails cleanly?

posthog-android/src/test/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegrationTest.kt
```diff
@@ -40,6 +42,30 @@ internal class PostHogActivityLifecycleCallbackIntegrationTest {
...
+                    integration.install(fake)
+                }.apply { start() }
+            }
+
+        ready.await()
+        start.countDown()
+        threads.forEach { it.join() }
+
+        verify(application, times(1)).registerActivityLifecycleCallbacks(any())
+        integrations.forEach { it.uninstall() }
```
turnipdabeets: [nit] Same as the core test — a `try/finally` around the assertion would keep the static flag reset if this ever fails.

posthog/src/test/java/com/posthog/internal/PostHogSendCachedEventsIntegrationTest.kt
marandaneto: Good catch. Fixed in c279b8ee by moving integration cleanup into a finally block, so the shared state is reset even when the assertion fails.

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
```diff
@@ -54,17 +54,18 @@ public class PostHogErrorTrackingAutoCaptureIntegration : PostHogIntegration, Th
     }
 
     private fun installHandler() {
+        if (!integrationInstalled.compareAndSet(false, true)) {
+            return
+        }
         adapterExceptionHandler.setDefaultUncaughtExceptionHandler(this)
```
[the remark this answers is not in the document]
marandaneto: Fixed in c279b8ee. Error-tracking install and uninstall are now synchronized per instance, so handler registration and restoration cannot interleave; only the winning owner can tear down the handler.

posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
```diff
@@ -84,18 +84,17 @@ internal class PostHogActivityLifecycleCallbackIntegration(
...
-        integrationInstalled = true
 
         this.postHog = postHog
         application.registerActivityLifecycleCallbacks(this)
     }
 
     override fun uninstall() {
         this.postHog = null
-        integrationInstalled = false
+        integrationInstalled.set(false)
```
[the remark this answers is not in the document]
marandaneto: Fixed in c279b8ee. All eight integrations now track whether the current instance owns the shared installation, and non-owner teardown returns without clearing the flag or removing resources. Added a regression test covering this scenario.

posthog-android/src/test/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegrationTest.kt
marandaneto: Good catch. Fixed in c279b8ee with the same try/finally cleanup pattern in the Android concurrency test.

posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
```diff
@@ -83,19 +84,28 @@ internal class PostHogActivityLifecycleCallbackIntegration(
     override fun onActivityDestroyed(activity: Activity) {
     }
 
+    @Synchronized
     override fun install(postHog: PostHogInterface) {
-        if (integrationInstalled) {
+        if (!integrationInstalled.compareAndSet(false, true)) {
             return
```
ioannisj, line 90: We should probably also `executor.shutdown()` here since this is a no-install?

ioannisj (approved):

posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt
marandaneto: Good point. The inline location is the activity integration and has no executor, but the no-install executor leak applies to PostHogSendCachedEventsIntegration. Fixed there in f2d54d0a: losing instances now shut down their unused executor, with concurrent coverage asserting every executor is closed.

PR #666, opened 2026-07-30
fix: identify anonymous user when id already matches persisted distinct id

ioannisj:

Calling `identify(id)` where `id` already equals the persisted distinct ID left the user anonymous: no transition fired, no person-processed event was sent, and `$is_identified` stayed false. It's easy to hit after a non-identified bootstrap seeds the same ID.

The SDK now upgrades the user to identified and captures one person-processed `$set` (no `$identify`, since there's no anonymous ID to merge). The transition is decided and persisted atomically under `identifiedLock`, so two concurrent `identify()` calls can't both emit a person event for the same transition. Ports PostHog/posthog-js#4328.

Added a regression test for the matching-ID-while-anonymous path in `PostHogBootstrapTest`: the user ends up identified, one `$set` is captured (no `$identify`), and flags reload only when the caller supplies properties that affect evaluation. Ran the unit suite.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj

### DIFF
--- a/.changeset/eager-maps-leave.md
+++ b/.changeset/eager-maps-leave.md
@@ -1,4 +1,5 @@
 ---
+"posthog": patch
 "posthog-android": patch
 ---
 
--- a/.changeset/eager-maps-leave.md
+++ b/.changeset/eager-maps-leave.md
@@ -1,0 +1,5 @@
+---
+"posthog-android": patch
+---
+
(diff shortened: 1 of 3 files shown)


marandaneto (approved):
unblocking, check for concurrency issues if not yet

PR #668, opened 2026-07-31
fix: attach map_id to frames from stateless exception capture

cat-ph:

`captureExceptionStateless` did not pass `config.releaseIdentifier` to `ThrowableCoercer`, so exceptions captured through the stateless path never got `map_id` on their stack frames and could not be symbolicated against uploaded ProGuard/R8 mappings. The stateful `captureException` path already threads it through correctly.

This also affected `posthog-server`, whose `captureException` delegates to the stateless path — and additionally the server `PostHogConfig` had no way to set a `releaseIdentifier` at all (`asCoreConfig()` never assigned one).

Changes:

- Pass `config.releaseIdentifier` to `ThrowableCoercer` in `captureExceptionStateless`, matching the stateful path.
- Expose `releaseIdentifier` on `posthog-server`'s `PostHogConfig` (property + builder method) and propagate it in `asCoreConfig()`.

- New regression test in `PostHogStatelessTest` asserting `map_id` appears on every frame when `releaseIdentifier` is set.
- New end-to-end test in posthog-server's `PostHogTest` asserting frames captured via the server `captureException` carry `map_id`.
- New `PostHogConfigTest` cases for `asCoreConfig()` propagation and the builder method.
- `make test`, `make checkFormat`, and `apiCheck` pass; `apiDump` regenerated for the new server API.

## :pencil: Checklist

### If releasing new changes


ablaszkiewicz (approved):
🚢

hpouillot (approved):

posthog-server/src/main/java/com/posthog/server/PostHogConfig.kt
```diff
@@ -167,6 +167,15 @@ public open class PostHogConfig constructor(
...
     public var flagDefinitionCacheProvider: PostHogFlagDefinitionCacheProvider? = null
 
+    /**
+     * Identifier attached as `map_id` to exception stack frames so PostHog can symbolicate
+     * them against an uploaded ProGuard/R8 mapping. It must match the map-id used when
+     * uploading the mapping, e.g. via
+     * `posthog-cli exp proguard upload --path "mapping.txt" --map-id "<releaseIdentifier>"`.
+     * Defaults to null (no `map_id` attached).
+     */
+    public var releaseIdentifier: String? = null
```
marandaneto, line 177: not sure we need this in the server package, proguard/r8 are mainly android tools
you can minify server bundles with proguard but its not a common thing since the bundle lives in the server and its not exposed for reverse engineering

cat-ph: hmm @marandaneto I was thinking that it's just general JVM SDK, so it can get shipped in desktop apps, CLI tools etc

it should technically be ~negligible cost here to pass the map_id (since captureException already does it too), but also related that I was looking into bringing more of the ET mechanism in "core" so we can expose it directly here too (i.e. #669 #670 and some upcoming draft ones I was testing/working on [extracted this one from #670 tbh]) 😁

marandaneto: yeah its not a problem, but I doubt it'd be used

marandaneto (approved):

PR #669, opened 2026-08-03
feat(error-tracking): complete exception chain metadata and in-app classification

cat-ph:

First PR in a 4-PR stack that brings JVM/server error tracking up to parity with the other PostHog SDKs.

This one completes the exception-item model that the shared `ThrowableCoercer` emits. Today every `$exception_list` item is serialized in isolation: there are no chain ids (the code still carried a `// TODO: exception_id and parent_id`), suppressed exceptions are dropped, nothing bounds the payload, and there is no way to force third-party frames out of `in_app`.

What changed:

- **Chain metadata** — each item's mechanism carries `exception_id` (0-based position); cause items also get `parent_id` and mechanism `type: "chained"`, while the primary item keeps its own mechanism type. A single-item list carries no ids at all, matching `posthog-rs` (which only links a chain when there is more than one exception). The ids are emitted on the wire; persisting the relationships needs the server-side mechanism-schema change (see "Review round 1" below).
- **Suppressed exceptions** — `Throwable.suppressed` (one level, bounded) is serialized after the cause chain with mechanism `type: "suppressed"` and the holder's `parent_id`.
- **Caps** — at most 50 items per `$exception_list` and 64 frames per stacktrace (keeping the frames nearest the crash). The 50-item cap bounds the traversal itself, not just the output.
- **Synthetic frames** — compiler-generated noise (JVM and Kotlin lambdas, Android D8/R8 desugared lambdas and outlines, Spring CGLIB proxies, reflection accessors, dynamic proxies) is flagged `method_synthetic: true` rather than dropped.
- **`inAppExcludes`** — new `PostHogErrorTrackingConfig.inAppExcludes` forces frames out of `in_app`; excludes win over `inAppIncludes`.

Notes for reviewers:

- Frame **ordering** is not part of this PR — crash-last ordering already shipped in #603, and this PR keeps that code and its comment as-is.
- The wire additions are additive: no existing key changes meaning, `platform: "java"` is unchanged, and the new fields are omitted rather than sent as `false`/`null` when they do not apply.
- `inAppExcludes` is a **body property**, not a constructor param, so every constructor descriptor of `PostHogErrorTrackingConfig` — including the Kotlin `$default` synthetic — is byte-identical to `main`; the only API-dump change for that class is the added getter.
- `ThrowableCoercer.fromThrowableToPostHogProperties` gains a trailing defaulted `inAppExcludes` param. Kotlin callers are source-compatible; the JVM descriptor changes, which is fine for a `@PostHogInternal` entry point.

- New `ThrowableCoercerTest` (11 tests) covering single-item id omission, a 3-deep cause chain, suppressed exceptions, both caps, bounded traversal of an endless cause chain, suppressed-fills-leftover-capacity, every synthetic-frame heuristic (with negatives), and excludes-beat-includes.
- Updated the existing `PostHogTest` exception assertions for the new mechanism fields (including the single-item case), on top of the ordering assertions from #603.
- `./gradlew :posthog:test` and `:posthog:apiCheck` pass; `posthog/api/posthog.api` regenerated with `apiDump` and the diff is additive apart from the documented `@PostHogInternal` defaulted-arg descriptors. `spotlessCheck` clean.

## :pencil: Checklist

### If releasing new changes

## :link: Stacked PR

Position 1 of 4. Base: `main`.

1. **this PR** — core exception chain metadata + in-app classification
2. `cat/java-et-server-config` — server error-tracking config and `captureException` options
3. `cat/java-et-uncaught` — opt-in server uncaught-exception capture

Please review and merge in stack order; each PR targets the previous branch.

### DIFF
--- a/.changeset/exception-chain-metadata.md
+++ b/.changeset/exception-chain-metadata.md
@@ -4,10 +4,10 @@
 
 Completes the exception-item model in the shared `ThrowableCoercer`:
 
-- Each `$exception_list` item's mechanism now carries `exception_id` (0-based position); cause items get `parent_id` and mechanism `type: "chained"`, the primary item keeps its existing type. A single-item list carries no ids, matching the other SDKs.
+- Each `$exception_list` item's mechanism now carries `exception_id` (0-based position); cause items get `parent_id` and mechanism `type: "chained"`, the primary item keeps its existing type. A single-item list carries no ids, matching the other SDKs. The chain metadata is emitted on the wire; persisting the relationships requires the PostHog-side mechanism-schema change (in flight), which today drops the ids on ingestion.
 - Suppressed exceptions (`Throwable.suppressed`) are serialized with mechanism `type: "suppressed"` and the holder's `parent_id`.
-- Caps: at most 50 items per `$exception_list` (keeps the primary + nearest causes) and 64 frames per stacktrace (keeps the frames nearest the crash).
-- JVM-synthesized frames (lambdas, Spring CGLIB proxies, reflection accessors, dynamic proxies) are flagged with `synthetic: true` rather than dropped.
-- New `PostHogErrorTrackingConfig.inAppExcludes` to force frames out of `in_app` (excludes win over `inAppIncludes`).
+- Caps: at most 50 items per `$exception_list` and 64 frames per stacktrace (keeps the frames nearest the crash). The 50-item cap bounds the traversal itself, so a pathological cause chain cannot be walked without limit.
+- JVM-synthesized frames (lambdas, Spring CGLIB proxies, reflection accessors, dynamic proxies) are flagged with `method_synthetic: true` rather than dropped.
+- New `PostHogErrorTrackingConfig.inAppExcludes` to force frames out of `in_app` (excludes win over `inAppIncludes`). Matching happens against runtime class names before symbolication, so on minified (ProGuard/R8) builds excludes generally will not match and server-side deobfuscation may reclassify frames afterwards; a deobfuscation-aware in-app contract is a follow-up.
 
 All field/key names and `platform: "java"` are unchanged; the additions are backwards compatible on the wire.
--- a/.changeset/exception-chain-metadata.md
+++ b/.changeset/exception-chain-metadata.md
@@ -7,7 +7,7 @@ Completes the exception-item model in the shared `ThrowableCoercer`:
 - Each `$exception_list` item's mechanism now carries `exception_id` (0-based position); cause items get `parent_id` and mechanism `type: "chained"`, the primary item keeps its existing type. A single-item list carries no ids, matching the other SDKs. The chain metadata is emitted on the wire; persisting the relationships requires the PostHog-side mechanism-schema change (in flight), which today drops the ids on ingestion.
 - Suppressed exceptions (`Throwable.suppressed`) are serialized with mechanism `type: "suppressed"` and the holder's `parent_id`.
 - Caps: at most 50 items per `$exception_list` and 64 frames per stacktrace (keeps the frames nearest the crash). The 50-item cap bounds the traversal itself, so a pathological cause chain cannot be walked without limit.
-- JVM-synthesized frames (lambdas, Spring CGLIB proxies, reflection accessors, dynamic proxies) are flagged with `method_synthetic: true` rather than dropped.
+- Compiler-generated frames (JVM and Kotlin lambdas, Android D8/R8 desugared lambdas and outlines, Spring CGLIB proxies, reflection accessors, dynamic proxies) are flagged with `method_synthetic: true` rather than dropped.
 - New `PostHogErrorTrackingConfig.inAppExcludes` to force frames out of `in_app` (excludes win over `inAppIncludes`). Matching happens against runtime class names before symbolication, so on minified (ProGuard/R8) builds excludes generally will not match and server-side deobfuscation may reclassify frames afterwards; a deobfuscation-aware in-app contract is a follow-up.
 
 All field/key names and `platform: "java"` are unchanged; the additions are backwards compatible on the wire.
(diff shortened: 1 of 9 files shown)


posthog/src/main/java/com/posthog/internal/errortracking/ThrowableCoercer.kt
```diff
@@ -22,17 +38,212 @@ public class ThrowableCoercer {
...
+     */
+    private fun isSyntheticFrame(
+        className: String,
+        methodName: String,
+    ): Boolean {
+        // lambdas: javac's synthetic `lambda$...` method, Kotlin's indy-generated `foo$lambda$N`,
+        // or the invokedynamic `$$Lambda` class (also the legacy D8 `-$$Lambda$Foo$hash` form)
+        if (methodName.startsWith(LAMBDA_METHOD_PREFIX) ||
+            methodName.contains(LAMBDA_METHOD_MARKER) ||
+            className.contains(LAMBDA_CLASS_MARKER)
```
marandaneto, line 59: suggestion: With Kotlin’s class-based lambda backend, frames commonly look like `ExampleKt$run$1.invoke`. None of these heuristics match that form, so `method_synthetic` is omitted in non-minified/debug Android builds despite the stated Kotlin-lambda support. Could we detect this generated class shape, guarded by an `invoke`-style method, and add a regression test?

marandaneto (approved):

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingConfig.kt
```diff
@@ -7,6 +7,9 @@ import com.posthog.PostHog
  */
 public class PostHogErrorTrackingConfig
     @JvmOverloads
+    // Do NOT add new options as constructor params: appending even a defaulted param rewrites the
+    // Kotlin `$default` synthetic constructor descriptor, so already-compiled consumers hit a
+    // NoSuchMethodError on `PostHogErrorTrackingConfig()`. New options go in the class body.
```
marandaneto, line 12: In a major release, remove constructor defaults and declare every trailing overload explicitly. Then no $default constructor is generated, and future
 options can add a new full constructor while retaining all previously shipped overloads. For arbitrary named optional settings, a builder/configuration
 block is the better API.

posthog/src/main/java/com/posthog/internal/errortracking/ThrowableCoercer.kt
cat-ph: Good catch — fixed in 32939f81.

The heuristics only covered the indy/desugared shapes, so a lambda compiled by Kotlin's CLASS backend (older Kotlin/AGP, or `-Xlambdas=class`) fell through: those become nested classes like `ExampleKt$run$1` with the body in `invoke`, which matched none of the existing markers.

New clause in `isSyntheticFrame`: mark the frame when the class name ends in a `$<digits>` segment AND the method is `invoke` or `invokeSuspend` (the coroutine variant of the same generated shape). The method guard is the deliberate conservative line — a plain anonymous class (`MyClass$1` running user-written `run`/`call`/whatever) stays unmarked, since that body is the user's code, not compiler output.

Four cases added to the parameterized heuristic test:

- `com.app.ExampleKt$run$1` + `invoke` -> marked
- `com.app.ExampleKt$main$1$2` + `invokeSuspend` (nested) -> marked
- `com.app.MyClass$1` + `run` -> NOT marked
- `com.app.MyClass$1` + `handleMessage` -> NOT marked

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingConfig.kt
cat-ph: Agreed, and worth writing down. The synthetic `$default` constructor is exactly the hazard you describe: adding a defaulted parameter changes its descriptor, so every previously compiled call site breaks at runtime — this area has already bitten us twice.

The body-property shape here was the deliberate choice for that reason (properties are additive, no constructor descriptor to break), but it does not solve the general case.

No code change on this PR. Noting the explicit-trailing-overloads / builder-or-configuration-block approach for the next major, where we can drop the constructor defaults outright.

cat-ph: 🤖 ^ but true, didn't change anything, did you want me to add a note in the comments here?

hpouillot:
🤖 I’m a robot reviewer. I found one reliability issue in the latest head.

`captureException` calls `isIgnoredThrowable` before the new bounded `ThrowableCoercer` traversal. When `ignoredExceptionTypes` is non-empty, `findIgnoredTypeInCauseChain` in `PostHogStateless.kt` still walks causes without the 50-item cap. A custom throwable whose `cause` returns a fresh instance on each read can therefore loop indefinitely here, including on an uncaught-exception thread, before the bounded coercer runs.

This prefilter also uses equality-based `hashSetOf`, while `ThrowableCoercer` now uses identity-based cycle detection. Distinct but value-equal causes can stop the prefilter early and prevent an ignored cause from being detected.

Could both paths use the same identity-based, 50-item bounded traversal? A regression test through the public capture path with a non-empty ignored-type list would cover the gap.

cat-ph:
thanks @hpouillot, fixed! going through `ThrowableCoercer.walkCauseChain,` and capped, plus 2 regression tests

hpouillot (approved):

.changeset/exception-chain-metadata.md
```diff
@@ -0,0 +1,13 @@
+---
+'posthog': minor
```
marandaneto, line 2: @cat-ph if you want this android and/or server, you also need the android/server changesets
now you need a new pr if thats the case

cat-ph: @marandaneto ah sorry, android was just the test file changed but we want it there actually good point T_T opening a new PR for the changesets

PR #670, opened 2026-08-03
feat(server): expose in-app frame config and captureException options

cat-ph:

Second PR in the 4-PR JVM error-tracking stack. It exposes the in-app classification configuration surface on the server SDK and adds the missing `captureException` overloads.

`releaseIdentifier` exposure + stateless parity fix shipped separately in #668 (merged); this PR now carries only the in-app classification config and capture options.

Contains:

1. **`feat(server): expose in-app frame config and captureException options`**:
   - `PostHogConfig.inAppIncludes` / `inAppExcludes` control `in_app` classification of captured frames (class-name prefix match; excludes always win). `inAppExcludes` defaults to a new `PostHogConfig.DEFAULT_IN_APP_EXCLUDES` list of common JVM/framework prefixes (JDK, Kotlin, Spring, Netty, servlet containers, HTTP clients, the PostHog SDK itself) so zero-config users get a sensible your-code vs framework split. Assigning your own list replaces the defaults.
   - Both are available on the config `Builder`.
   - New `captureException(exception, distinctId, options)` / `captureException(exception, options)` overloads take `PostHogCaptureOptions` with the same merging semantics as `capture(..., options)`: custom properties, `$groups`, `$set`/`$set_once`, timestamp, and feature-flag enrichment via a pre-evaluated `flags` snapshot or `appendFeatureFlags`. Reserved exception properties (e.g. `$exception_level`, `$exception_fingerprint`) can be overridden through options properties. Request-context distinct-id resolution and personless fallback behave exactly like the existing overloads. Because the options overload sits next to the existing properties overload at the same arity, Java callers that passed an explicit untyped `null` as the third argument now need to cast it — called out in the changeset and the KDoc.
2. **`docs(server): use the canonical proguard upload command in releaseIdentifier KDoc`** — the KDoc landed in #668 cites the hidden `posthog-cli exp proguard upload` alias; use the canonical `posthog-cli proguard upload`.

- 13 new tests in `posthog-server` (7 in `PostHogConfigTest`, 6 in `PostHogTest`): defaults and overrides for the new config, `DEFAULT_IN_APP_EXCLUDES` contents, exclude-beats-include classification, and each new `captureException` overload including options merging, reserved-property overrides, personless fallback, and `map_id` presence/absence on the options path.
- `./gradlew :posthog:test :posthog-server:test :posthog:apiCheck :posthog-server:apiCheck` pass; `spotlessCheck` clean. API dumps regenerated; the `posthog-server` diff is additive only on top of #668's entries, which now come from `main`.

## :pencil: Checklist

### If releasing new changes

## :link: Stacked PR

Position 2 of 4. Base: `cat/java-et-coercer` (PR #669).

1. PR #669 — core exception chain metadata + in-app classification
2. **this PR** — server in-app classification config and `captureException` options
3. `cat/java-et-uncaught` — opt-in server uncaught-exception capture

Review after its parent; the diff shown here is only this PR's own commits once the parent merges.


posthog-server/src/main/java/com/posthog/server/PostHogInterface.kt
```diff
@@ -817,4 +817,43 @@ public sealed interface PostHogInterface {
...
+     *   `$exception_level` can be overridden via the options properties.
+     *
+     * Java callers passing an explicit untyped `null` third argument must cast it
+     * (`captureException(e, id, (Map<String, Object>) null)`) since it now matches both this
+     * overload and the properties one.
+     */
+    public fun captureException(
+        exception: Throwable,
+        distinctId: String?,
+        options: PostHogCaptureOptions,
```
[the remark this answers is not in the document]
cat-ph: leaving unresolved for thoughts, but I think it makes sense and simpler than 'polluting' with another method etc

posthog-server/src/main/java/com/posthog/server/PostHog.kt
```diff
@@ -247,6 +270,56 @@ public class PostHog : PostHogStateless(), PostHogInterface {
...
+    override fun captureException(
+        exception: Throwable,
+        distinctId: String?,
+        options: PostHogCaptureOptions,
+    ) {
+        if (!enabled) {
+            return
+        }
+
+        try {
```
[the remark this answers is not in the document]
cat-ph: Valid finding, thanks — fixed in 9ee09cbeb3a1dbe799e6ebae5d24d0d0add29604.

Rather than adding an `isIgnoredThrowable` call to the options path (which would re-diverge the next time the guarded path grows a step), both paths now go through one shared pre-capture route in core: `PostHogStateless.captureExceptionEvent` owns the `ignoredExceptionTypes` prefilter, the coerce-then-merge property order and the personless distinct-id fallback, and additionally accepts the event fields `captureExceptionStateless` cannot carry (groups, `$set`/`$set_once`, timestamp). `captureExceptionStateless` delegates to it and the server options overload calls it instead of coercing and calling `captureStateless` itself, so the two paths cannot drift apart again.

Regression tests in `PostHogTest`:

- `captureException with options drops throwables matching ignoredExceptionTypes`
- `captureException with options drops a throwable whose cause matches ignoredExceptionTypes`
- `captureException with options merges options for types outside ignoredExceptionTypes`

posthog-server/src/main/java/com/posthog/server/PostHog.kt
```diff
@@ -247,6 +269,46 @@ public class PostHog : PostHogStateless(), PostHogInterface {
...
+            val captureContext = PostHogRequestContext.resolveCaptureContext(distinctId, options.properties)
+            val callerProperties =
```
hpouillot: 🤖 **Robot review:** the suppression/opt-out gates run after option enrichment. `mergeCaptureProperties()` can perform a remote `/flags` request when `appendFeatureFlags=true`; only afterwards does `captureExceptionEvent()` drop an ignored throwable, and `captureStateless()` checks `optOut`. An ignored or opted-out exception can therefore still make a network request (including its distinct ID and flag-evaluation person properties) even though no exception event will be queued.

Please run the shared exception/opt-out preflight before `mergeCaptureProperties()`, or make enrichment lazy so it runs only after those gates. This is separate from the fixed bypass bug: the exception is now dropped correctly, but only after avoidable work/data transmission.

posthog-server/src/main/java/com/posthog/server/PostHog.kt
```diff
@@ -247,6 +269,46 @@ public class PostHog : PostHogStateless(), PostHogInterface {
...
+            // captureExceptionStateless cannot carry groups/user properties/timestamp, so this
+            // overload takes the shared core route instead: it runs the same ignoredExceptionTypes
+            // prefilter and merges callerProperties AFTER the coerced exception properties (so
+            // options can still override reserved keys like $exception_level).
+            super.captureExceptionEvent(
+                exception,
+                distinctId = captureContext.distinctId,
+                properties = callerProperties,
+                userProperties = options.userProperties,
+                userPropertiesSetOnce = options.userPropertiesSetOnce,
```
hpouillot: 🤖 **Robot review:** `userProperties` and `userPropertiesSetOnce` should not be forwarded on a `$exception` event. The dedicated exception ingestion pipeline is read-only, and its prepare step explicitly deletes both `$set` and `$set_once` before emission ([ingestion code](https://github.com/PostHog/posthog/blob/master/nodejs/src/ingestion/pipelines/errortracking/prepare-event-step.ts#L43-L52)); `$exception` is denied person updates because the separate pipeline cannot preserve ordering with normal person updates ([person-update contract](https://github.com/PostHog/posthog/blob/master/nodejs/src/ingestion/common/persons/person-update.ts#L55-L57)).

As written, the SDK wire tests pass but neither value ever updates the person, so the KDoc/changeset promise of `$set`/`$set_once` parity is misleading and the payload is wasted. `options.userProperties` can still be used above as feature-flag evaluation input, but I would pass neither map into `captureExceptionEvent` (and remove those helper parameters/claims).

posthog-server/src/main/java/com/posthog/server/PostHog.kt
```diff
@@ -247,6 +269,46 @@ public class PostHog : PostHogStateless(), PostHogInterface {
...
+        options: PostHogCaptureOptions,
+    ) {
+        if (!enabled) {
+            return
+        }
+
+        try {
+            val captureContext = PostHogRequestContext.resolveCaptureContext(distinctId, options.properties)
+            val callerProperties =
+                mergeCaptureProperties(
```
marandaneto, line 297: blocking: Filter ignored exceptions before feature-flag enrichment

This calls `mergeCaptureProperties` before `captureExceptionEvent` applies `ignoredExceptionTypes`. With `appendFeatureFlags=true`, an ignored exception can still trigger a `/flags` request before being dropped, creating avoidable load. Please prefilter before enrichment (or defer enrichment until core accepts the exception), and cover that ignored exceptions issue no `/flags` request.

posthog-server/src/main/java/com/posthog/server/PostHogConfig.kt
```diff
@@ -279,6 +306,33 @@ public open class PostHogConfig constructor(
...
         public const val DEFAULT_POLL_INTERVAL_SECONDS: Int = 30
 
+        /**
+         * Default [inAppExcludes] prefixes: common JVM, Kotlin, and server-framework packages
+         * whose frames are noise in application stack traces. Assigning your own list to
+         * [inAppExcludes] replaces these defaults.
+         */
+        @JvmField
+        public val DEFAULT_IN_APP_EXCLUDES: List<String> =
+            listOf(
```
marandaneto: blocking: Do not expose a process-wide mutable default list

`listOf(...)` is exposed to Java here as a mutable fixed-size `Arrays$ArrayList`, and configs/builders share this same instance. A Java caller can call `set()` through the public field/getter and mutate defaults process-wide. Please expose an unmodifiable default and defensively copy it per config/builder, with mutation-isolation or Java-interop coverage.

marandaneto (approved):
still some comments, but unblocking

.changeset/core-shared-exception-capture-route.md
```diff
@@ -0,0 +1,13 @@
+---
+'posthog': patch
+'posthog-android': patch
```
marandaneto, line 3: should it be server here instead

posthog-server/src/main/java/com/posthog/server/PostHog.kt
cat-ph: yep thanks, fixed + added test!

posthog-server/src/main/java/com/posthog/server/PostHog.kt
cat-ph: thanks, fixed + test!

posthog-server/src/main/java/com/posthog/server/PostHog.kt
cat-ph: verified + fixed!

posthog-server/src/main/java/com/posthog/server/PostHogConfig.kt
cat-ph: good catch, ty, fixed!

.changeset/core-shared-exception-capture-route.md
cat-ph: not instead, but I added it under too (good spot, ty!), since this was for the core change that bumped android/server too

marandaneto: ok, i got confused with the pr title (feat/server)

hpouillot (approved):

PR #671, opened 2026-08-03
feat(server): opt-in uncaught exception capture

cat-ph:

Third PR in the 4-PR JVM error-tracking stack: opt-in capture of uncaught JVM exceptions for the server SDK.

- New `PostHogConfig.captureUncaughtExceptions` (also on the `Builder`), default off. When enabled, `PostHog.setup` installs a global `Thread.defaultUncaughtExceptionHandler` that captures the crashing exception as a fatal, unhandled `$exception` event (mechanism `UncaughtExceptionHandler`), flushes, then delegates to the previously registered handler. `close()` removes it again.
- Unlike the Android SDK, this is gated purely on the local flag: the server SDK never fetches remote config, so the remote autocapture toggle can never fire. To support that, the shared `PostHogErrorTrackingAutoCaptureIntegration` takes an optional caller-supplied `enabledGate` instead of always using the built-in gate (local `errorTrackingConfig.autoCapture` + remote config as a kill-switch, i.e. the behavior from #648 is preserved for Android).
- The handler now delivers captures through an internal `CaptureTarget` seam (`installWith`), because the core `PostHogInterface` and the stateless server client share no capture supertype.
- With no previous default handler present, the handler reproduces the JVM's own `Exception in thread "..."` stderr output, so opting into capture never hides a crash from stderr log collection.
- New internal `PostHogCapturedThrowables` guard: a weak, identity-keyed, process-wide set letting independent capture paths avoid double-reporting the same `Throwable` instance. It is deliberately directional — the uncaught handler only *marks* (a crash is always captured as the authoritative fatal/unhandled record, even if the same instance was logged first), while log-mirror paths (the appender in PR 4) *consult* it and skip instances already reported. It never keeps a throwable or its stack alive.

Delivery caveat documented in the KDoc and changeset: `flush()` drains the queue synchronously on the crashing thread, but the preceding capture enqueues asynchronously, so under an immediate hard exit the final exception is best-effort — the same guarantee the Android SDK gives. Services that care should keep `flushAt` low.

- 15 new tests: 7 in the core `PostHogErrorTrackingAutoCaptureIntegrationTest` (local-only gate installs/refuses, `CaptureTarget` capture+flush, capture-not-suppressed-by-dedup, marking for later log mirrors, JVM stderr fallback, non-owning instance cannot tear down the installed handler), 2 in `PostHogCapturedThrowablesTest` (identity keying, value-equal throwables both captured), and 6 in the new server `PostHogUncaughtExceptionTest` (off by default, install+chain+restore on close, fatal/unhandled/mechanism assertions over a real `/batch` request, repeated setup keeps ownership, second client does not stack a handler, works with no remote config).
- `./gradlew :posthog:test :posthog-server:test` pass (`PostHogErrorTrackingAutoCaptureIntegrationTest` 33 tests, `PostHogCapturedThrowablesTest` 2, `PostHogUncaughtExceptionTest` 6), plus `:posthog:apiCheck` / `:posthog-server:apiCheck` and `spotlessCheck`. API dumps regenerated; additive only.

## :pencil: Checklist

### If releasing new changes

## :link: Stacked PR

Position 3 of 4. Base: `cat/java-et-server-config` (PR #670).

1. PR #669 — core exception chain metadata + in-app classification
2. PR #670 — server error-tracking config and `captureException` options
3. **this PR** — opt-in server uncaught-exception capture


posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
```diff
@@ -158,13 +237,33 @@ public class PostHogErrorTrackingAutoCaptureIntegration : PostHogIntegration, Th
...
+                // itself: this is the authoritative fatal/unhandled record for the crash and must not
+                // be downgraded by an earlier handled capture of the same instance
+                // (`logger.error(..., e); throw e`).
+                PostHogCapturedThrowables.markAndCheck(throwable)
+                target.capture(PostHogThrowable(throwable, thread))
+                // The queue sends a fatal exception event synchronously on this (the crashing)
+                // thread, bypassing the flushAt threshold; this flush covers anything else still
+                // pending. Delivery is still best-effort under an immediate hard exit — same
+                // guarantee as the Android SDK.
+                target.flush()
```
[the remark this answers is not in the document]
cat-ph: Good catch, and you were right that my "the queue sends fatal exception events synchronously" claim was wrong for this SDK. I had verified it against core `PostHogQueue`, which routes fatal records through `submitSyncSafely`, but the server client never uses that queue — it always runs `PostHogMemoryQueue` (wired in `PostHogConfig.asCoreConfig`), where `add()` enqueues on the serial queue executor and `flush()` ran inline on the crashing thread with an early return when the deque looked empty. So the crash path really did race its own enqueue and usually sent nothing.

Fixed in d2ced6cdff27cc2531a384675d359b095487ea1c. `PostHogMemoryQueue.flushBlocking(timeoutMs)` submits a barrier task onto the same single-threaded executor that `add()` uses and waits on a latch bounded by the timeout, so the send is ordered after the pending enqueue instead of racing it, and it ignores `flushAt`. Only the crash path uses it: the `CaptureTarget.flush()` wired in `PostHog.setup` calls `flushBlocking(CRASH_FLUSH_TIMEOUT_MS)` (2000 ms, mirroring the Rust SDK's bounded panic-hook flush); the public `flush()` is unchanged.

The docs are corrected to best-effort framing rather than a synchronous-delivery promise: the config KDoc, the integration comment and the changeset now say the capture is enqueued asynchronously and the crash path performs a bounded blocking flush so the event gets a network attempt before the JVM exits, with the timeout, HTTP failure and hard-exit caveats spelled out. Making the fatal event itself a synchronous, threshold-independent send is recorded in the changeset as a follow-up.

Regression coverage in `PostHogUncaughtExceptionTest` and `PostHogMemoryQueueTest`: `crash event is flushed even while its enqueue is still pending` parks the queue thread inside a gated batch request, runs the handler on its own thread, asserts it is still blocked in the flush with the crash batch not yet sent, then releases the gate and asserts the crash batch arrived before the handler returned; `flushBlocking sends an event whose enqueue has not run yet` pins the inline-flush blindness at the queue level; `flushBlocking gives up after the timeout instead of hanging` covers the stalled-executor path; and `flushBlocking still flushes when the calling thread is already interrupted` covers a crash on an interrupted thread. The first two fail against the old inline flush.

posthog-server/src/main/java/com/posthog/server/PostHog.kt
```diff
@@ -27,11 +28,54 @@ public class PostHog : PostHogStateless(), PostHogInterface {
...
+                // The uncaught Throwable is a PostHogThrowable carrying fatal/handled=false/mechanism;
+                // routing it through captureException preserves those via the shared coercer, and the
+                // queue sends fatal exception events synchronously on the crashing thread.
+                integration.installWith(
+                    object : PostHogErrorTrackingAutoCaptureIntegration.CaptureTarget {
+                        override fun capture(throwable: Throwable) {
+                            captureException(throwable)
+                        }
+
+                        override fun flush() {
```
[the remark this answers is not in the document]
cat-ph: Confirmed, and fixed in d2ced6cdff27cc2531a384675d359b095487ea1c. The enqueue-ordering race is closed by a barrier rather than by hoping the scheduler cooperates: `PostHogMemoryQueue.flushBlocking(timeoutMs)` submits a task onto the very same single-threaded executor that `add()` posts to, so the flush happens-after the pending enqueue, and it drains ignoring `flushAt`. The crashing thread then waits on that barrier for at most a named `CRASH_FLUSH_TIMEOUT_MS` (2000 ms, mirroring the Rust SDK's bounded panic-hook flush), waiting out the full budget even if the thread was already interrupted and restoring the interrupt flag afterwards. Only the crash path uses this; the public `flush()` keeps its inline behaviour.

Delivery is documented as bounded and best-effort rather than guaranteed: the flush can hit the timeout, the batch is capped at `maxBatchSize`, a flush already in progress on the timer thread makes the barrier a no-op, and a hard exit can still cut the HTTP attempt short.

Two new tests pin it. `crash event is flushed even while its enqueue is still pending` parks the queue thread inside a gated `/batch` request so the crash enqueue is provably pending, runs the handler on a separate thread, asserts that thread is still blocked inside the flush and that the crash batch has not been sent, then opens the gate and asserts the crash batch arrived before the handler returned — it fails against the old inline flush with "Expected the handler to still be blocked inside the crash flush". `flushBlocking gives up after the timeout instead of hanging` stalls the executor and asserts the call returns false at roughly the timeout instead of hanging. I also dropped the `flushAt(1)` from the existing fatal-event test, since that threshold was masking the race.

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
```diff
@@ -158,13 +237,33 @@ public class PostHogErrorTrackingAutoCaptureIntegration : PostHogIntegration, Th
...
             }
         }
 
         // Always delegate: we may still be mid-chain even while dormant.
-        defaultExceptionHandler?.uncaughtException(thread, throwable)
+        val previousHandler = defaultExceptionHandler
+        if (previousHandler != null) {
+            previousHandler.uncaughtException(thread, throwable)
+        } else if (throwable !is ThreadDeath) {
```
marandaneto: blocking: Deprecated `ThreadDeath` reference fails compilation

This project treats warnings as errors, so this check makes `:posthog:compileKotlin` fail with `ThreadDeath is deprecated`. Please avoid referencing `ThreadDeath`, or suppress the deprecation at the narrowest appropriate scope.

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
```diff
@@ -158,13 +237,33 @@ public class PostHogErrorTrackingAutoCaptureIntegration : PostHogIntegration, Th
...
-                postHog.captureException(PostHogThrowable(throwable, thread))
-                postHog.flush()
+            captureTarget?.let { target ->
+                // Mark the throwable so post-crash log mirrors of this exact instance (e.g. a
+                // shutdown hook logging the crash) don't re-report it — but never skip the capture
+                // itself: this is the authoritative fatal/unhandled record for the crash and must not
+                // be downgraded by an earlier handled capture of the same instance
+                // (`logger.error(..., e); throw e`).
+                PostHogCapturedThrowables.markAndCheck(throwable)
+                target.capture(PostHogThrowable(throwable, thread))
```
marandaneto: blocking: Telemetry failures can prevent uncaught-exception delegation

If `capture` or `flush` throws, execution never reaches the previous handler or JVM stderr fallback, so enabling PostHog can hide or replace the application’s original crash handling. Please guarantee delegation from a `finally` path.

posthog-server/src/main/java/com/posthog/server/internal/PostHogMemoryQueue.kt
```diff
@@ -76,6 +78,84 @@ internal class PostHogMemoryQueue(
...
+
+        try {
+            executor.execute {
+                try {
+                    flushIgnoringThreshold()
+                } finally {
+                    drained.countDown()
+                }
+            }
+        } catch (e: Throwable) {
```
marandaneto: blocking: Broad catch hides unrelated executor failures

Catching `Throwable` here turns VM errors and programming failures into a logged `false` result that the caller ignores. Catch only the expected stable exception type, such as `RejectedExecutionException`, and let unexpected failures propagate to the uncaught-handler boundary.

posthog/src/main/java/com/posthog/internal/errortracking/PostHogCapturedThrowables.kt
```diff
@@ -0,0 +1,69 @@
...
+    // synchronized on the set. Cleared entries are pruned opportunistically via [queue].
+    private val seen = HashSet<IdentityWeakKey>()
+
+    /**
+     * Records [throwable] as captured and reports whether the caller should capture it.
+     *
+     * @return true if this is the first time the instance has been marked (the caller should
+     *   capture it), false if it was already marked (the caller should skip it).
+     */
+    public fun markAndCheck(throwable: Throwable): Boolean =
```
marandaneto: suggestion: Defer the unused process-wide captured-throwable guard

No production log-mirror consumer consults `markAndCheck`, and the only production call discards its result. This adds synchronized process-wide state and public API solely for hypothetical future use; please add the concrete consumer in this change or defer this guard until it is required.

.changeset/core-captured-throwables.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog': patch
```
marandaneto: we need only 1 posthog changeset, we have 2

posthog-server/src/main/java/com/posthog/server/internal/PostHogMemoryQueue.kt
```diff
@@ -76,6 +78,84 @@ internal class PostHogMemoryQueue(
...
+            // RejectedExecutionException and friends: the executor is gone, there is nothing to await
+            config.logger.log("Blocking flush could not be scheduled: $e.")
+            return false
+        }
+
+        return awaitUninterruptibly(drained, timeoutMs)
+    }
```
marandaneto, line 241: on android we do this with an exception type marker
https://github.com/PostHog/posthog-android/blob/dea84f90a9acd86dd5ec053cdfe6e026ea48558b/posthog/src/main/java/com/posthog/internal/PostHogQueue.kt#L138
i wonder if that would be easier?

posthog-server/src/main/java/com/posthog/server/PostHog.kt
```diff
@@ -26,12 +28,80 @@ public class PostHog : PostHogStateless(), PostHogInterface {
             }
         }
 
+    /**
+     * Uncaught-exception integration installed when [PostHogConfig.captureUncaughtExceptions] is
+     * enabled, retained so it can be uninstalled on [close].
+     */
+    private var uncaughtExceptionIntegration: PostHogErrorTrackingAutoCaptureIntegration? = null
```
marandaneto, line 35: i dont think this should live here, the integration should rather call flush directly instead

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
```diff
@@ -158,13 +237,33 @@ public class PostHogErrorTrackingAutoCaptureIntegration : PostHogIntegration, Th
...
+        val previousHandler = defaultExceptionHandler
+        if (previousHandler != null) {
+            previousHandler.uncaughtException(thread, throwable)
+        } else if (throwable !is ThreadDeath) {
+            // No previous default handler: reproduce the JVM's built-in crash output that
+            // ThreadGroup would have printed had we not installed ourselves as the default handler,
+            // so opting into capture never hides crashes from stderr log collection. ThreadDeath is
+            // excluded because ThreadGroup stays silent for it.
+            System.err.print("Exception in thread \"${thread.name}\" ")
+            throwable.printStackTrace(System.err)
```
marandaneto, line 291: probably has to be removed or log with the logger

posthog-server/src/test/java/com/posthog/server/PostHogUncaughtExceptionTest.kt
```diff
@@ -0,0 +1,379 @@
...
+        // Already sent when the handler returned, so this should not have to wait.
+        val request = mockServer.takeRequest(1, TimeUnit.SECONDS)
+        assertNotNull(request, "Expected the crash event to be flushed before the handler returned")
+
+        val batch = request.parseBatch()
+        val exceptionEvent = batch.findEvent("\$exception")
+        assertNotNull(exceptionEvent, "Expected an \$exception event")
+
+        val props = batch.eventProperties("\$exception")
+        assertEquals("fatal", props["\$exception_level"], "Uncaught exceptions must be fatal")
```
dustinbyrne, line 142: Consider not classifying all thread failures as fatal

The default handler receives uncaught exceptions from every thread. A worker-thread exception can terminate only that thread while the JVM continues.

Please emit `error` for a boundary that can continue. Emit `fatal` only when process termination is expected. Add tests for both cases.

posthog-server/src/test/java/com/posthog/server/PostHogUncaughtExceptionTest.kt
```diff
@@ -0,0 +1,379 @@
...
+        val exceptionList = props["\$exception_list"] as? List<Map<String, Any?>>
+        assertNotNull(exceptionList, "Expected a \$exception_list")
+        assertTrue(exceptionList.isNotEmpty())
+
+        @Suppress("UNCHECKED_CAST")
+        val mechanism = exceptionList.first()["mechanism"] as? Map<String, Any?>
+        assertNotNull(mechanism, "Expected a mechanism on the first exception item")
+        assertEquals(false, mechanism["handled"], "Uncaught exceptions must be marked handled=false")
+        assertEquals(
+            "UncaughtExceptionHandler",
```
dustinbyrne: Consider canonical automatic exception metadata

This assertion records a noncanonical mechanism type. Use `mechanism.type="onuncaughtexception"`. Add a stable lowercase `$exception_source` for the JVM hook.

Please also correct the exception-list metadata:

- Always include `value`.
- Include `exception_id=0` for one exception.
- Mark causes with `source="cause"`.
- Do not copy the outer handled state to nested entries.
- Use `type="chained"` and `source="suppressed"` for suppressed exceptions.
- Traverse all suppressed exception trees in depth-first order.

The source and type gaps come from this integration. The shared coercer already contains the other gaps, but this path publishes them.

posthog-server/src/main/java/com/posthog/server/PostHog.kt
```diff
@@ -26,12 +28,80 @@ public class PostHog : PostHogStateless(), PostHogInterface {
...
+            val alreadySetUp = isEnabled()
+            super.setup(config.asCoreConfig())
+            if (alreadySetUp || !isEnabled()) {
+                return
+            }
+
+            // Core setup never installs integrations for the stateless base, so wire the uncaught
+            // handler explicitly. Gate purely on the local server flag — the server SDK never fetches
+            // remote config, so the remote-config gate the Android SDK uses can never fire here.
+            // Single-owner by design: the handler is process-wide, so only the first client that opts in
```
dustinbyrne, line 56: suggestion: Define ownership after the owner closes

This policy stops capture while another opted-in client remains active. This can affect server processes that replace a client.

Please transfer ownership when the owner closes, or document the first-owner restriction in the public configuration documentation.

posthog-server/src/test/java/com/posthog/server/PostHogUncaughtExceptionTest.kt
```diff
@@ -0,0 +1,379 @@
...
+                PostHogConfig.builder(TEST_API_KEY)
+                    .host(mockServer.url("/").toString())
+                    .captureUncaughtExceptions(true)
+                    .build(),
+            )
+
+        val handler = Thread.getDefaultUncaughtExceptionHandler()
+        assertTrue(handler is PostHogErrorTrackingAutoCaptureIntegration)
+
+        handler.uncaughtException(Thread.currentThread(), IllegalStateException("kaboom"))
```
dustinbyrne: suggestion: Add real JVM process tests

This test calls the handler directly. It cannot prove the real process exit, worker-thread continuation, stderr output, or delivery timing before termination.

Please add subprocess tests for these cases:

- The main failure terminates the process.
- A worker-thread failure does not terminate the process.
- The SDK preserves JVM stderr output.
- The SDK calls the previous handler.
- The event uses canonical metadata.

posthog-server/src/test/java/com/posthog/server/PostHogUncaughtExceptionTest.kt
dustinbyrne: via https://github.com/PostHog/sdk-specs/blob/main/openspec/specs/exception-event-metadata/spec.md

dustinbyrne: also https://github.com/PostHog/sdk-specs/blob/main/openspec/specs/capture-exception/spec.md defines list order, frame order, and stack preservation

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
cat-ph: good catch! they log-and-fall-through now

posthog-server/src/main/java/com/posthog/server/internal/PostHogMemoryQueue.kt
cat-ph: fixed, ty! narrowed to RejectedExecutionException (and reworked slightly)

posthog/src/main/java/com/posthog/internal/errortracking/PostHogCapturedThrowables.kt
cat-ph: I dropped it, was planning to use in #672 (unsure yet if we want that, but separate discussion, moved it there)

.changeset/core-captured-throwables.md
cat-ph: fixed, dropped the captured-throwable stuff

posthog-server/src/main/java/com/posthog/server/internal/PostHogMemoryQueue.kt
cat-ph: great point and thanks! yeah that turned out easier, updated

posthog-server/src/main/java/com/posthog/server/PostHog.kt
cat-ph: fair, updated!

posthog-server/src/test/java/com/posthog/server/PostHogUncaughtExceptionTest.kt
cat-ph: updated, ty! main thread is fatal, others are error

posthog-server/src/test/java/com/posthog/server/PostHogUncaughtExceptionTest.kt
cat-ph: fixed what was in scope here, mechanism.type and source; I'll take a look at the metadata/coercer in a separate PR

posthog-server/src/main/java/com/posthog/server/PostHog.kt
cat-ph: documented for now, the transfer felt a bit overengineering but happy to rework/revisit if you feel strongly about this!

posthog-server/src/test/java/com/posthog/server/PostHogUncaughtExceptionTest.kt
cat-ph: yep, ty! added some more tests (and caught another issue 😆)

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
cat-ph: fixed! moved to a narrow scoped suppression

posthog/src/main/java/com/posthog/errortracking/PostHogErrorTrackingAutoCaptureIntegration.kt
cat-ph: so as I understand it in the JVM, it won't print it anymore if we set an uncaught-exception handler and then there are 2 cases:

1. if it already had a handler, we delegate to it
2. if it had **no** handler, we'd need to print it here as usual, otherwise it's silently dropped

AIUI the config.logger is default off unless debug logging is enabled, and it goes with the SDK logs, but this was keeping the JVM behavior

I added a test to assert the stderr contains the trace (Dustin also asked for one)

posthog-server/src/main/java/com/posthog/server/PostHog.kt
```diff
@@ -26,12 +27,94 @@ public class PostHog : PostHogStateless(), PostHogInterface {
...
+            if (config.captureUncaughtExceptions) {
+                getConfig<com.posthog.PostHogConfig>()?.let { coreConfig ->
+                    val integration =
+                        PostHogErrorTrackingAutoCaptureIntegration(coreConfig, { true }, ::isProcessFatal)
+                    // The uncaught Throwable is a PostHogThrowable carrying fatal/handled=false/mechanism;
+                    // routing it through captureException preserves those via the shared coercer. A
+                    // fatal-level event takes PostHogMemoryQueue's bounded blocking fatal path inside
+                    // add() (same fatal-record marker the core queue keys on), so capture() itself
+                    // delivers the crash — and everything queued ahead of it — before returning.
+                    integration.installWith(
```
marandaneto: blocking: Roll back a denied handler installation — `Thread.setDefaultUncaughtExceptionHandler` can throw `SecurityException` under a `SecurityManager`. Because this runs after `super.setup` enables the client and `installHandler` sets its process-wide ownership flag, `PostHog.with` can throw while leaving a live half-initialized client and preventing later installations. Make this installation transactional or clean up before propagating.

.changeset/core-uncaught-gate.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog': patch
+---
+
+`PostHogErrorTrackingAutoCaptureIntegration` can now be gated on a caller-supplied strategy instead of the built-in gate (local `errorTrackingConfig.autoCapture` with remote config as a kill-switch): a new `PostHogErrorTrackingAutoCaptureIntegration(config, enabledGate)` constructor lets SDK layers that never fetch remote config (e.g. the server SDK) decide autocapture purely from local config. The uncaught handler also delivers captures through an internal `CaptureTarget` seam (`installWith`) so it can drive clients that are not a core `PostHogInterface`, and when no previous default handler exists it now reproduces the JVM's own `Exception in thread ...` stderr output, so installing capture never hides a crash from log collection. Uncaught-exception events now carry the canonical `mechanism.type` `onuncaughtexception` (previously the noncanonical `UncaughtExceptionHandler`), per the sdk-specs exception-event-metadata spec. Android behavior and the existing `install(PostHogInterface)` path are otherwise unchanged; the additions are internal (`@PostHogInternal`) and visible only because of the multi-module architecture.
```
marandaneto: nit: Document the actual constructor signature — this changeset advertises `(config, enabledGate)`, but the API dump exposes the new overload as `(config, enabledGate, fatalPolicy)`. Please update the release note to match the published signature.

marandaneto (commented):
Automated advisory code review.

.changeset/core-uncaught-gate.md
marandaneto: i think changesets should be one liner, only with whats the value for the customer rather than text wall with internal details

.changeset/server-uncaught-exceptions.md
```diff
@@ -0,0 +1,7 @@
...
...s` (also on the config `Builder`). When enabled, `PostHog` installs a global `Thread.defaultUncaughtExceptionHandler` on setup that captures the crashing exception as an unhandled `$exception` event (mechanism `onuncaughtexception`, `$exception_source: jvm.uncaught_exception_handler`; `$exception_level` is `fatal` for a main-thread crash and `error` for a worker-thread one, whose death the process survives), flushes, and then delegates to the previously registered handler; the handler is removed again on `close()`. Unlike the Android SDK this is gated purely on the local flag — the server SDK never fetches remote config. A fatal `$exception` event takes a dedicated queue path (keyed off the same fatal-event marker the Android SDK's queue uses): the enqueue and the send run as one ordered task on the queue executor, draining the queue batch by batch and ignoring `flushAt`, so a backlog larger than one batch cannot strand the crash event, and the crashing thread blocks on it for at most a bounded timeout (2s). Delivery stays best-effort, the same guarantee class as the Android SDK: the drain can hit the timeout and each HTTP attempt can fail or be cut short by an immediate hard exit.
```
marandaneto: same as https://github.com/PostHog/posthog-android/pull/671/changes#r3851626914

posthog-server/src/main/java/com/posthog/server/PostHog.kt
marandaneto: mm not yet?

marandaneto (approved):
approving to unblock, still a few comments left

posthog-server/src/main/java/com/posthog/server/PostHog.kt
cat-ph: great catch, ty, rolling the ownership back + regression test

.changeset/core-uncaught-gate.md
cat-ph: dropped all this, simplified changesets

cat-ph: yeah totally agree, I simplified them!

posthog-server/src/main/java/com/posthog/server/PostHog.kt
cat-ph: 🤔 might have dropped the commit while playing with jj, sorry

updated! moved to PostHogUncaughtExceptionCapture, just calls install here

..but I left the flush no-op because I think capture() already uses the fatal path and the sync flush could block the crashing thread (well, no timeout as I see)

PR #680, opened 2026-08-06
feat(surveys): add opt-in requireDeviceTypeTargeting config

dmamujee:

Android currently treats surveys with missing or empty `conditions.deviceTypes` as matching every device. That makes it easy for surveys intended for other platforms (for example web surveys that only set URL/selector conditions) to become eligible on Android, because those web-only conditions are not evaluated by the Android SDK.

This change adds an opt-in `PostHogSurveysConfig.requireDeviceTypeTargeting` flag so apps can require explicit device-type targeting before a survey is eligible.

When `false` (default): preserve existing behavior — missing/empty `deviceTypes` still match.

When `true`:
- Missing `conditions.deviceTypes` does not match
- Empty `conditions.deviceTypes` does not match
- Non-empty conditions continue to use the existing match operators against `Mobile`, `Tablet`, or `TV`

- Added `PostHogSurveysDeviceTypeTargetingTest` covering default allow, require-deny for missing/empty, matching and non-matching explicit device types, and existing match operators (`EXACT`, `IS_NOT`, `I_CONTAINS`)
- Ran `:posthog-android:testDebugUnitTest --tests "com.posthog.android.surveys.PostHogSurveysDeviceTypeTargetingTest"`
- Ran `:posthog:test`, `:posthog:apiDump`, and `make checkFormat`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with Cursor agent assistance. Approach follows an opt-in SDK config on `PostHogSurveysConfig` rather than a custom survey predicate, so eligibility stays in the existing device-type matching layer and default behavior remains unchanged for other consumers.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/require-device-type-targeting.md
+++ b/.changeset/require-device-type-targeting.md
@@ -1,0 +1,6 @@
+---
+'posthog': minor
+'posthog-android': minor
+---
+
+Add opt-in `PostHogSurveysConfig.requireDeviceTypeTargeting` to exclude surveys without explicit device-type targeting on Android. When enabled, surveys with missing or empty `conditions.deviceTypes` are ineligible; surveys with a non-empty device-type condition continue to use the existing match operators against `Mobile`, `Tablet`, or `TV`.
--- a/posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
@@ -183,8 +183,10 @@ public class PostHogSurveysIntegration(
     }
 
     private fun doesSurveyDeviceTypesMatch(survey: Survey): Boolean {
-        val deviceTypes = survey.conditions?.deviceTypes ?: return true
-        if (deviceTypes.isEmpty()) return true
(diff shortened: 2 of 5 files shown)


dustinbyrne:
thanks for the pull request @dmamujee!

at first glance this looks reasonable. i'm curious if @PostHog/team-surveys has an opinion on this? i could see this being classified as a bug, in which case we might want to just fix it without the additional configuration option

lucasheriques:
hey @dmamujee thanks for the PR! I'm Lucas and I work on the Surveys product.

We have a feature request that I think might be better suited for this, potentially. [Take a look here](https://github.com/PostHog/posthog/issues/41318)

It's about having surveys be able to target specifically SDKs/platforms (i.e. only mobile native, mobile web, etc).

would that fix your use case? because I think having a `requireDeviceTypeTargeting` option is a little bit odd, and I worry about how to make sure this option is available on the UI too, because it doesn't really seem to fit the current options we have.

Also, if you want to make sure your survey displays only on mobile, doesn't the option `device type is Mobile` is enough for you now? Can you tell more about the actual product use case you're looking for? I think we can figure out a better solution (maybe it's that feature request)

PR #681, opened 2026-08-06
fix(server): keep locally evaluated flags when one flag is inconclusive

matheus-vb:

`evaluateFlags` discarded the entire locally computed flag map as soon as one flag definition was inconclusive, then replaced the snapshot with a `/flags` response. One foreign flag gated on a person property the caller never passes therefore forced a billed request per identity, `flagKeys` could not prevent it because it was applied only after a successful local pass, and during a `/flags` outage trivially resolvable flags read `false` for the whole cache window.

`evaluateFlags` now evaluates locally first and keeps what it resolved, asking `/flags` only for the keys that stayed unresolved, which is what every other server SDK already does. `flagKeys` scopes the local evaluation loop rather than the definitions map, so flag dependencies still resolve. `onlyEvaluateLocally` is now strictly local and never serves cached remote values. `posthog-server.api` is unchanged.

### Request volume

Exactly one new path reaches the billed `/flags` endpoint: a non `InconclusiveMatchException` error while evaluating a flag definition now falls back instead of propagating into the caller. That only fires for a customer whose local evaluation is already throwing, so it trades a crash for one request rather than quietly adding volume.

Every other input state is less than or equal to today, and two strictly decrease. The inconclusive trigger is unchanged, `flagKeys` only shrinks the evaluated set, the cache is still consulted before every request with failures honoured, and customers without a personal API key never reach the new code. Verified across fifteen input states against a worktree at the base commit: thirteen identical, two lower. `$feature_flag_called` volume is unchanged, since values are recomputed per call but consistently.

The change most worth reviewer attention is not billing: a group aggregated flag evaluated without `groups` resolves locally to `false`, and that now takes precedence over the server's answer.

Added coverage in `PostHogEvaluateFlagsTest` and `PostHogFeatureFlagsTest` for the local wins merge, `flagKeys` scoping, flag dependencies under scoping, outage plus negative caching, strict `onlyEvaluateLocally`, undefined requested keys, empty definitions, group flags, and definitions that fail to load. Each assertion was checked by mutating the implementation and confirming only the intended test fails. Full module suite is green at 442 tests, plus `make checkFormat` and `apiDump` with no diff.

## :pencil: Checklist

### If releasing new changes


posthog-server/src/main/java/com/posthog/server/PostHogInterface.kt
```diff
@@ -647,19 +647,25 @@ public sealed interface PostHogInterface {
     )
 
     /**
-     * Evaluate every feature flag for [distinctId] in a single `/flags` round-trip and return a
-     * snapshot. Repeat lookups against the snapshot do not make additional network requests, and
+     * Evaluate every feature flag for [distinctId] and return a snapshot. With local evaluation
+     * configured, flags resolvable from the definitions in memory are answered locally and keep
+     * those values; a single `/flags` request fills in only the keys that stayed unresolved.
```
dustinbyrne, line 652: isn't this effectively the same from the end user perspective? i.e., we're just swapping out locally evaluated flags for remote flags - ideally they _should_ evaluate to be the same things, but i'd consider `/flags` to be the correct canonical source

this is still a billable request either way

posthog-server/src/main/java/com/posthog/server/PostHogInterface.kt
```diff
@@ -647,19 +647,25 @@ public sealed interface PostHogInterface {
...
      *   distinct ID is used; if none exists, an empty snapshot is returned.
      * @param groups groups for group-based flags
      * @param personProperties person properties for flag evaluation
      * @param groupProperties group properties for flag evaluation
-     * @param flagKeys when non-empty, restricts the underlying request to the given keys; this is
-     *   distinct from [PostHogFeatureFlagEvaluations.only] which filters in memory after the call
-     * @param onlyEvaluateLocally when true, do not fall back to a `/flags` request if local
-     *   evaluation cannot resolve every flag
+     * @param flagKeys when non-empty, restricts both local evaluation and the underlying request to
+     *   the given keys, so a flag you did not ask for cannot force a request. Requested keys with no
```
dustinbyrne, line 662: this looks like the real bug fix imo

posthog-server/src/main/java/com/posthog/server/PostHogInterface.kt
dustinbyrne: i suppose which set takes precedence is [ambiguous in the sdk specification](https://github.com/PostHog/sdk-specs/blob/77bcedf8e55e00363543c1cf1c2efb7f79a5d130/openspec/specs/get-feature-flags-and-payloads/spec.md)

> ### Server-side flow
>
> 1. **Resolve evaluation context** from `distinct_id` plus optional groups, person properties, group properties, device id, and geoip settings.
> 2. **Attempt local evaluation** of all requested flags when definitions are available.
> 3. **Collect values and payloads** from the local evaluation result.
> 4. **Fall back to remote evaluation** if local evaluation is unavailable, incomplete, or explicitly bypassed.
> 5. **Return both maps together** in one result object.

marandaneto:
moving to draft until comments are resolved

marandaneto:
@matheus-vb are you gonna follow up on this?

posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
```diff
@@ -274,15 +296,48 @@ internal class PostHogFeatureFlags(
...
+        if (flagKeys != null) {
+            val undefined = flagKeys.filterNot { currentFlagDefinitions.containsKey(it) }
+            if (undefined.isNotEmpty()) {
+                config.logger.log(
+                    "No local definition for requested flag(s) ${undefined.joinToString(", ")} - " +
+                        "they will be absent from locally-evaluated snapshots; " +
+                        "check for deleted flags or typos",
+                )
+            }
+        }
```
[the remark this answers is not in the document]
gustavohstrassburger: Keeping this behavior, and I don't think it's a regression — it's deliberate parity with posthog-python, which is the reference server SDK here.

In `posthog/client.py`, `_get_all_flags_and_payloads_locally` filters the definitions by `flag_keys_to_evaluate` *before* the evaluation loop:

```python
flags_to_process = self.feature_flags
if flag_keys_to_evaluate:
    flag_keys_set = set(flag_keys_to_evaluate)
    flags_to_process = [
        flag for flag in self.feature_flags if flag["key"] in flag_keys_set
    ]
```

A requested key with no local definition is dropped by that filter, so it never reaches a `try` block and never sets `fallback_to_flags`. Same outcome as here: absent from the snapshot, no `/flags` request bought on its own.

The definitions-lag case you describe is real, and I've made it explicit rather than leaving it implied — the log line now reads "check for deleted flags, typos, or a flag created since the last definitions poll", and the same caveat is on `PostHogInterface.evaluateFlags`, `PostHogEvaluateFlagsOptions`, and the changeset. The tradeoff is deliberate: making an unknown key force a billed request per identity is the exact failure mode this PR exists to remove, and it would make a typo in `flagKeys` unboundedly expensive.

If an unresolved flag already forces the request, the undefined key does ride along and gets filled — covered by `a requested undefined key is filled by a request an unresolved flag already forced`.

posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
```diff
@@ -943,19 +1003,28 @@ internal class PostHogFeatureFlags(
...
+                entry.flags
+            } else {
+                getFeatureFlagsFromRemote(
+                    distinctId,
+                    groups,
+                    personProperties,
+                    groupProperties,
+                    flagKeys,
+                    disableGeoip,
+                ).also { entry = cache.getEntry(cacheKey) }
```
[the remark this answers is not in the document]
gustavohstrassburger: Agreed with your own downgrade note: the accurate fix is documentation, not excluding the keys from the request. Addressed in 5e13e06 — the KDoc on `evaluateFlags`/`PostHogEvaluateFlagsOptions` and the changeset no longer claim undefined keys are unconditionally absent, they now say the key is absent *unless an unresolved flag already requires the `/flags` call, which then also fills it*.

Not narrowing the request body to unresolved keys, for two reasons: it would make server-only and newly-created flags permanently invisible (the regression raised in the sibling thread), and the keys ride a request `needsRemote` already forced, so there's no extra round-trip and no extra billed request.

Covered by `a requested undefined key is filled by a request an unresolved flag already forced`.

posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
```diff
@@ -274,15 +296,48 @@ internal class PostHogFeatureFlags(
...
             } catch (e: InconclusiveMatchException) {
                 config.logger.log("Local evaluation inconclusive for flag '$key': ${e.message}")
-                if (!onlyEvaluateLocally) {
-                    // Allow fallback to remote evaluation
-                    return null
-                }
+                needsRemote = true
+            } catch (e: Throwable) {
+                config.logger.log("Local evaluation failed for flag '$key': ${e.message}")
+                needsRemote = true
```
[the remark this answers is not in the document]
gustavohstrassburger: Added in 5e13e06: `an error thrown while evaluating one flag falls back for that flag instead of crashing`. It pairs a conclusive definition with `throwingFlagDefinition` (a multivariate definition holding a null variant, which makes `computeFlagLocally` NPE) and asserts evaluation doesn't propagate, the local value wins, the broken key is filled by exactly one `/flags` request, and the two keys' `locally_evaluated` markers differ.

While here I also narrowed the branch from `catch (Throwable)` to `catch (Exception)` (4220fa0) — `Throwable` would have swallowed `OutOfMemoryError`/`StackOverflowError` and then issued a network request on the way out. posthog-python catches `Exception` at the same spot. The test is unaffected; an NPE is an `Exception`.

posthog-server/src/main/java/com/posthog/server/PostHogInterface.kt
gustavohstrassburger: Not quite the same, and this is the part I'd most like a second opinion on.

On billing you're right that a request is a request — the win isn't cheaper requests, it's *fewer* of them. Before this change one inconclusive definition discarded the whole locally-computed batch, so a single foreign flag gated on a person property the caller never passes forced a `/flags` request per identity, forever. After it, that flag alone is unresolved and the rest keep their local values. When every requested flag resolves locally the request count goes to zero, which is the case `flagKeys` scoping now makes reachable.

Where it's genuinely *not* equivalent: local and remote can disagree, and the sharpest case is a group-aggregated flag evaluated without `groups`. `computeFlagLocally` answers `false` rather than throwing, so the flag counts as resolved and that `false` now beats the server's `true`. I called it out in the PR body and pinned it with `a group flag with no groups supplied resolves locally to false and wins over the server` so it's a decision rather than a surprise. It's also exactly how posthog-python behaves today.

The other non-equivalence is a `/flags` outage: locally-resolvable flags used to read `false` for the whole cache window and now hold their real values.

posthog-server/src/main/java/com/posthog/server/PostHogInterface.kt
gustavohstrassburger: Agreed — that's the one with teeth. `flagKeys` used to be applied only *after* a successful local pass, so a flag you never asked about could still make the pass inconclusive and force a request. It now scopes the evaluation loop, which is what makes "ask for the two flags I care about, pay nothing" actually work.

Worth noting it scopes the loop and not the definitions map — `computeFlagLocally` resolves flag dependencies through the full map, so narrowing the map would make every dependent flag inconclusive. Pinned by `flagKeys scoping leaves flag dependencies resolvable`.

posthog-server/src/main/java/com/posthog/server/PostHogInterface.kt
gustavohstrassburger: You're right that the spec doesn't settle it — step 4 says "fall back to remote evaluation if local evaluation is unavailable, incomplete, or explicitly bypassed" without saying which set wins on overlap. So I took posthog-python's `evaluate_flags` as the tiebreaker, since it's the SDK furthest along on this flow. It records local results first, then on fallback skips any key it already resolved:

```python
for key, detail in response.get("flags", {}).items():
    if key in locally_evaluated_keys:
        continue
```

So: local wins, remote fills gaps only. That's what this PR now does, and I've put the reasoning in a comment next to the merge (4220fa0) so it isn't re-litigated on the next pass.

Happy to flip it to remote-wins if you'd rather treat `/flags` as canonical — it's a one-line change to the merge direction plus the group-flag test. But it would put android out of step with python, and it would mean local evaluation is never authoritative, which makes `onlyEvaluateLocally` a bit incoherent. Might be worth pinning down in sdk-specs either way so the next SDK doesn't have to guess.

posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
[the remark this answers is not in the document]
gustavohstrassburger: valid — the definitions-lag scenario is real. Resolved by documenting the tradeoff rather than changing the behavior, for posthog-python parity.

posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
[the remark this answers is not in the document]
gustavohstrassburger: valid — and the downgrade note was right that the fix is documentation, not excluding the keys.

posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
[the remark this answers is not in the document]
gustavohstrassburger: valid — genuine coverage gap, test added.

gustavohstrassburger:
@marandaneto @dustinbyrne — review comments are addressed, back out of draft.

Since your last pass:

- `5e13e06` — clarified the undefined-key log line and the KDoc/changeset contract, which no longer claims undefined keys are unconditionally absent. Added `an error thrown while evaluating one flag falls back for that flag instead of crashing`.
- `4220fa0` — narrowed the new catch from `Throwable` to `Exception`, so an `OutOfMemoryError` is no longer swallowed into a network request, and documented the merge precedence next to the merge itself.

CI is green on `4220fa0`. Four of the five threads are resolved.

The one left open is the merge precedence thread on `PostHogInterface.kt` — @dustinbyrne raised that `/flags` is arguably the canonical source, and the sdk-specs wording does not settle which set wins on overlap. This PR currently does local-wins with remote filling gaps, following posthog-python. Flipping it is a one-line change to the merge direction plus the group-flag test, so it just needs a call.

posthog-server/src/main/java/com/posthog/server/PostHogInterface.kt
dustinbyrne: this is fine. we actually had a gap in the SDK specifications where `evaluate_flags`/`evaluateFlags` wasn't spec'd out, so I've done that here: https://github.com/PostHog/sdk-specs/blob/main/openspec/specs/evaluate-flags/spec.md

given the implementation across our server-side SDKs, we settled on:
> The SDK SHALL use a cached evaluated result or attempt local evaluation when its architecture supports either source. When local evaluation does not produce the required set and local-only mode is false, the SDK MAY make at most one direct remote /flags (or equivalent) evaluation request for the call. A locally resolved value SHALL NOT be overwritten by a remote fallback value for the same key in the resulting snapshot.

so we're good

dustinbyrne (approved):
👍 👍

PR #687, opened 2026-08-07
fix(error-tracking): bound the ignoredExceptionTypes cause-chain walk

cat-ph:

The `ignoredExceptionTypes` prefilter (`findIgnoredTypeInCauseChain`, consulted by both `PostHog.captureException` and `captureExceptionStateless` before coercion) walked `throwable.cause` with equality-based cycle detection and no depth bound. Two ways that hangs the capture caller:

- a throwable whose `getCause()` returns a fresh throwable on every call defeats *any* visited-set cycle detection — the walk loops forever and allocates unboundedly, worst on synchronous crash paths;
- a throwable overriding `equals`/`hashCode` can make the hash-set report a false cycle on genuinely distinct links (early stop, wrong answer) or fail to detect a real one.

(Shipped in #608, spotted during review of #669.)

## Fix semantics

- **Identity-based cycle detection**: `Collections.newSetFromMap(IdentityHashMap())` instead of `hashSetOf`, so user-defined `equals`/`hashCode` can't confuse the walk.
- **Hard bound of 50 links** (`MAX_IGNORED_CHECK_CHAIN_LENGTH`): every traversal of a user-provided exception graph must be bounded independently, since `getCause()` is user code.
- **Proceed on limit**: hitting the bound without a match returns "not ignored", so the capture proceeds rather than dropping the event.

**Stacked on #669** (base `cat/java-et-coercer`; will restack onto `main` after it merges). On that base the coercer's own walk is already identity-based and bounded by `MAX_EXCEPTION_LIST_SIZE = 50`, so this closes the last unbounded traversal of user-provided exception graphs on the capture path. The bound here mirrors that cap; unifying both walks into one shared bounded helper is left as a follow-up so this stays a surgical fix.

New unit tests in `PostHogStatelessTest`:

- regenerating-cause throwable (fresh object per `getCause()` read) with an invocation counter — cause reads stay ≤ the bound;
- self-cycle and two-object mutual cycle terminate;
- `equals`/`hashCode`-colliding throwables are still walked by identity — a distinct-but-equal second link matching an ignored type is found;
- match at the root and mid-chain still short-circuits;
- ignored type parked beyond the bound → prefilter returns no match AND `captureExceptionStateless` still enqueues the `$exception` event (truncated chain);
- existing test: a matching ignored type still skips capture.

`make checkFormat`, `make testJava`, and `apiDump` (no public API diff) all pass.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.changeset/bound-ignored-exception-walk.md
+++ b/.changeset/bound-ignored-exception-walk.md
@@ -1,0 +1,5 @@
+---
+"posthog": patch
+---
+
+Bound the `ignoredExceptionTypes` prefilter's cause-chain walk. It previously walked `throwable.cause` with no depth limit and equality-based cycle detection, so a throwable whose `getCause()` returns a fresh object on every call looped forever in the prefilter, and one whose `equals`/`hashCode` misreport identity could stop the walk early and miss an ignored cause. The walk is now identity-based and capped at 50 links; hitting the cap without a match lets the capture proceed rather than dropping the event.
--- a/.changeset/exception-chain-metadata.md
+++ b/.changeset/exception-chain-metadata.md
@@ -4,10 +4,10 @@
 
 Completes the exception-item model in the shared `ThrowableCoercer`:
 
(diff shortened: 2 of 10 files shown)


cat-ph:
fixed in #669, closing

Issue #675, opened 2026-08-05
Push: setOptOut(true) can strand a queued unregister DELETE, and optIn() never re-registers the token


ioannisj:
### Problem

`unregisterPushNotificationToken()` resolves its caller before the DELETE actually runs: `unregisterCurrent()` queues `performUnregister` on the subscription manager's executor, and `performUnregister` starts with `if (closed || config.optOut) return`. So the sequence `unregisterPushNotificationToken()` then `setOptOut(true)`, which is exactly what a wrapper SDK's opt-out flow produces, can flip `config.optOut` before the executor runs, and the DELETE silently parks. The pending unregister is persisted, but the same guard applies on later launches, so the server-side subscription stays active for the entire opted-out period and Workflows keep targeting a device that opted out.

The reverse transition has a gap too: when a DELETE succeeds the stored token is cleared, but `optIn()` only flips consent, it doesn't refetch the FCM token or re-register. A user who opts out and back in stays unsubscribed until the app restarts (or `onNewToken` happens to fire).

Found while reviewing the React Native push integration (PostHog/posthog-js#4415, [thread](https://github.com/PostHog/posthog-js/pull/4415#discussion_r3719629071)): the wrapper dispatches unregister before `setOptOut` in call order, but bridge-call ordering can't order the executor work behind it, so this isn't fixable from the wrapper side.

### Proposal

Treat consent transitions as first-class in the subscription manager:

- Let a pending unregister execute while opted out. A DELETE removes data rather than collecting it, so gating it on `config.optOut` inverts the intent of opt-out; the guard should distinguish "don't register/send" from "don't clean up".
- On `optIn()` with `capturePushNotificationSubscriptions` enabled, refetch the token and re-register, so opt-in re-arms push without an app restart.

PR #688, opened 2026-08-10
fix: send the push unregister DELETE even after opt-out (#675)

dmarchuk:
## Problem

Opting out did not stop Workflows push. `unregisterPushNotificationToken()` resolves before the DELETE runs, and `performUnregister` began with `if (closed || config.optOut) return`. The normal opt-out flow — `unregisterPushNotificationToken()` then `setOptOut(true)` — let opt-out flip `config.optOut` before the executor ran the unregister, so the DELETE silently parked. The server-side subscription stayed active for the whole opted-out period and Workflows kept delivering to a device that opted out (#675).

## Change

`performUnregister` no longer gates on `config.optOut`, only on `closed`. A DELETE removes data rather than collecting it, so opt-out must block registration and sends but not cleanup. The register and send paths keep their `config.optOut` guards unchanged.

- New unit test `opt-out during an in-flight unregister strands the DELETE (#675)`: registers a token, blocks the manager's single executor thread, queues `unregisterCurrent()` behind it, flips `config.optOut` before it runs, then releases. Asserts a DELETE reaches the mock server. Fails on the pre-fix code (reproduces the bug), passes with the fix.
- Updated the existing opted-out unregister test to assert the DELETE is sent, not skipped.
- Full `PostHogPushSubscriptionManagerTest` suite passes via `./gradlew :posthog:test`.
- Emulator E2E against a local stack (via `adb reverse`). Registered a token, then ran `unregisterPushNotificationToken()` then `optOut()`. The DELETE reached the backend (200). Rebuilt with the pre-fix `config.optOut` guard and the same flow logged the unregister but sent no DELETE.

## Not in this PR

#675 also reports that `optIn()` does not refetch the token or re-register. That half is fixed separately in #689.

> [!NOTE]
> Draft. Directed by @dmarchuk; implemented and verified locally by Claude while @ioannisj is out.

### DIFF
--- a/.changeset/push-unregister-optout.md
+++ b/.changeset/push-unregister-optout.md
@@ -1,5 +1,6 @@
 ---
 "posthog": patch
+"posthog-android": patch
 ---
 
 Fix: opting out no longer strands an in-flight push unregister. The unregister `DELETE` is data removal, so it now completes even after `setOptOut(true)` instead of leaving the server-side subscription active for the whole opted-out period (#675).
--- a/.changeset/push-unregister-optout.md
+++ b/.changeset/push-unregister-optout.md
@@ -1,5 +1,5 @@
 ---
-"posthog-android": patch
+"posthog": patch
 ---
 
 Fix: opting out no longer strands an in-flight push unregister. The unregister `DELETE` is data removal, so it now completes even after `setOptOut(true)` instead of leaving the server-side subscription active for the whole opted-out period (#675).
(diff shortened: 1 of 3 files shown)


.changeset/push-unregister-optout.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog": patch
```
marandaneto, line 2: you need posthog-android here as well

marandaneto (approved):

.changeset/push-unregister-optout.md
dmarchuk: Added, thank you!

PR #689, opened 2026-08-10
fix(push): re-register the device token on opt-in (#675)

dmarchuk:
## Problem

The opt-in half of #675. After a logout `unregisterPushNotificationToken()` clears the stored device token, `optIn()` only flips consent back on. It never refetches the FCM token or re-registers the device, so a user who opts out and back in stays unsubscribed and Workflows can't target them until the next app launch re-runs registration (or `onNewToken` happens to fire).

## Change

Adds an `onOptIn()` lifecycle hook to `PostHogIntegration` (default no-op). `PostHog.optIn()` now notifies every integration, mirroring how `onRemoteConfig` is dispatched. `PostHogPushSubscriptionIntegration.onOptIn()` refetches the FCM token and re-registers the device, reusing the same executor and token fetcher as install.

The push integration is only registered when `capturePushNotificationSubscriptions` is enabled, so the refetch is gated to auto-capture apps. Manual push mode is unchanged: the host owns the token lifecycle.

## How tested

- Unit test: `onOptIn` refetches the token and re-registers the device (a second registration after install), in `PostHogPushSubscriptionIntegrationTest`.
- Emulator E2E against a local stack with a real FCM token (Firebase on the sample's classpath, `adb reverse` to the backend). The app auto-registered its FCM token on launch, opt out cleared it, then opt in refetched the FCM token and re-registered the device (`POST /api/push_subscriptions`, 200). Rebuilt with the opt-in dispatch disabled and the same opt-in produced no re-register.
- The iOS counterpart (PostHog/posthog-ios#760) is confirmed the same way on a real device.

## Not in this PR

The strand-the-unregister-DELETE half of #675 is fixed separately in #688. This PR is independent and targets `main`.

> [!NOTE]
> Draft. Directed by @dmarchuk; implemented and verified locally by Claude while @ioannisj is out.


posthog/src/main/java/com/posthog/PostHogIntegration.kt
```diff
@@ -29,4 +29,13 @@ public interface PostHogIntegration {
...
     public fun onRemoteConfig(loaded: Boolean = true) {
     }
+
+    /**
+     * Called when the user opts back in via [PostHogInterface.optIn], so an integration that stood
+     * down while opted out can re-arm. The push integration uses this to refetch the device token and
+     * re-register, since a prior logout unregister cleared it and opt-in alone would leave the device
+     * unsubscribed until the next app launch.
+     */
+    public fun onOptIn() {
```
marandaneto: blocking: Preserve interface compatibility

This no-op body is emitted through `PostHogIntegration.DefaultImpls`, while `onOptIn()` remains abstract in the JVM interface. Java implementations will fail to recompile, and previously compiled integrations can throw `AbstractMethodError` when this callback is dispatched. Please avoid extending the public interface in this patch release, or route the notification through an internal or genuinely JVM-compatible default mechanism.

dmarchuk: I see, thanks for the catch! 🙏

marandaneto (approved):

PR #690, opened 2026-08-10
fix(surveys): exclude web-only surveys on Android

dustinbyrne:

**Why:** Surveys scoped to the web via a CSS `selector` or `url` display condition were leaking onto native Android and rendering as inert prompts with non-functional targeting.

`PostHogSurveysIntegration.getActiveMatchingSurveys()` did not check `conditions.url` or `conditions.selector`. Android cannot evaluate either condition, so matching native device-type or event targeting is not enough to make such a survey eligible.

This brings posthog-android in line with the canonical `surveys` contract in [PostHog/sdk-specs](https://github.com/PostHog/sdk-specs/commit/4623e11a32687e1ee8abe73ca2f051f19ea5662a) and the behavior already shipped in posthog-ios and posthog-react-native.

Analogous of https://github.com/PostHog/posthog-ios/pull/733

### What changed

- Added a filter to exclude surveys with any non-empty `url` or `selector` condition.
- Kept surveys without web conditions eligible for the existing native targeting checks.
- Added regression coverage for URL, selector, device-type, and event-targeting combinations.

- `./gradlew :posthog-android:testDebugUnitTest --tests 'com.posthog.android.surveys.PostHogSurveysWebConditionsTest'`
- `make test`
- `make checkFormat`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

**Autonomy:** Fully autonomous

Authored by PostHog Code. A Changeset entry was added since the repo uses Changesets rather than direct CHANGELOG edits.

---

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
@@ -191,6 +191,16 @@ public class PostHogSurveysIntegration(
     }
 
     /**
+     * Returns whether a survey has CSS selector or URL conditions that cannot be
+     * evaluated outside a browser DOM. These surveys are excluded from native
+     * matching even when they also have device-type or event conditions.
+     */
+    private fun hasWebConditions(survey: Survey): Boolean {
+        val conditions = survey.conditions ?: return false
+        return !conditions.url.isNullOrEmpty() || !conditions.selector.isNullOrEmpty()
+    }
+
+    /**
      * Get surveys enabled for the current user.
      * Uses the cached surveys pushed from remote config and filters for active matching surveys.
      *
@@ -232,6 +242,10 @@ public class PostHogSurveysIntegration(
 
             // 2. Filter out surveys that don't match device type
             if (!doesSurveyDeviceTypesMatch(survey)) return@filter false
+
+            // 2.5. Filter out surveys with CSS selector / URL conditions that
+            // cannot be evaluated natively, so they must never display on Android
+            if (hasWebConditions(survey)) return@filter false
 
             // 3. Filter out seen surveys (unless they can activate repeatedly)
             if (getSurveySeen(survey)) return@filter false
--- a/.changeset/exclude-web-only-surveys.md
+++ b/.changeset/exclude-web-only-surveys.md
@@ -1,0 +1,5 @@
+---
+"posthog-android": patch
+---
+
+Fix surveys scoped to web via a CSS selector or URL display condition leaking onto native Android. Surveys with `conditions.url` and/or `conditions.selector` are now excluded from active matching surveys, since those conditions can only be evaluated in a browser DOM. This mirrors the exclusion already shipped in posthog-ios and posthog-react-native.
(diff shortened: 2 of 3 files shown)


posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
```diff
@@ -190,6 +190,30 @@ public class PostHogSurveysIntegration(
...
+     */
+    private fun hasOnlyWebConditions(survey: Survey): Boolean {
+        val conditions = survey.conditions ?: return false
+
+        val hasWebCondition = !conditions.url.isNullOrEmpty() || !conditions.selector.isNullOrEmpty()
+        if (!hasWebCondition) return false
+
+        val hasNativeTargeting = !conditions.deviceTypes.isNullOrEmpty() || hasEvents(survey)
+        return !hasNativeTargeting
+    }
```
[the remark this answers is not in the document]
dustinbyrne: valid — regression tests for URL + matching device targeting and selector + activated event targeting failed against the previous implementation. The filter now excludes any survey with a non-empty URL or selector, and the focused and full unit test suites pass in 8ee7722c.

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
```diff
@@ -190,6 +190,16 @@ public class PostHogSurveysIntegration(
...
 
+    /**
+     * Returns whether a survey has CSS selector or URL conditions that cannot be
+     * evaluated outside a browser DOM. These surveys are excluded from native
+     * matching even when they also have device-type or event conditions.
+     */
+    private fun hasWebConditions(survey: Survey): Boolean {
+        val conditions = survey.conditions ?: return false
+        return !conditions.url.isNullOrEmpty() || !conditions.selector.isNullOrEmpty()
+    }
```
marandaneto, line 201: i think this makes sense but it'd also mean that a survey that is permissive for mobile users (intentionally) would stop appearing, so it's better to confirm with @PostHog/team-surveys for its impact
@PostHog/team-surveys you can approve if thats ok

marandaneto: oh looks like the pr is already merged for ios
we probably need this in react native and flutter then, but i'd still check with surveys team

marandaneto (approved):

posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
dustinbyrne: yeah this was originally initiated by SDK compliance monitor

via [surveys/spec.md](https://github.com/PostHog/sdk-specs/blob/c59b528b58c775295b19297858946d0b62de715a/openspec/specs/surveys/spec.md)
> 5. **Filter active matching surveys.** Eligibility checks include active/running status, device type, wait-period rules, linked/targeting/internal feature flags, optional event/action activation conditions, and platform-specific display constraints. Non-web/native SDKs MUST exclude surveys whose only display-targeting conditions are web-only (a CSS `selector` and/or `url` match) — those conditions are unevaluable outside a browser DOM, so displaying such a survey natively renders an inert prompt with no working targeting.

arnohillen (approved):
[arno's agent] Reviewed at head 8ee7722c: change verified against head code, CI classified, no blocking findings.

arnohillen:
[arno's agent] The showNextSurvey test exercises the onSurveysLoaded auto-trigger rather than calling showNextSurvey directly, so a regression in showNextSurvey itself would not fail it.

lucasheriques (approved):

PR #696, opened 2026-08-11
fix: include publishing docs in assemble

marandaneto:

The `posthog` 6.29.2 release failed while creating its Javadoc artifact. Three Dokka dependencies had been removed from the Gradle lockfiles for `posthog` and `posthog-server`. The regular PR build did not create the published documentation artifacts, so it did not resolve those dependencies or catch the incomplete lockfiles.

This restores the missing lock entries. It also adds the Javadoc and HTML documentation jars to `assemble` for `posthog`, `posthog-server`, and `posthog-android-gradle-plugin`. The existing `make compile` CI step now creates the same documentation artifacts used during publishing and fails if their dependency locks are incomplete.

No changeset is included because this does not change SDK code or package behavior.

- `make compile`
- `make updateLocks`
- `make checkFormat`
- `make dryReleaseCore`
- `make dryReleaseServer`
- `make dryReleaseAndroid`
- `make dryReleaseAndroidPlugin`
- Removed the three Dokka entries locally and verified that `./gradlew :posthog:assemble --rerun-tasks` failed with the same missing-lock error as the release
- `git diff --check`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi coding agent investigated the failed release logs and implemented the fix in a dedicated worktree. We chose to make published documentation part of each relevant module's normal assembly instead of adding special cases to the root `Makefile` or CI workflow. The final autoreview was skipped at the human driver's request.


marandaneto:
https://github.com/PostHog/posthog-android/actions/runs/31481312761/job/93749612468

dmarchuk (approved):

PR #698, opened 2026-08-14
fix: normalize SDK timestamps to UTC

marandaneto:

SDK dates were formatted with the process default time zone. The JSON represented the correct instant, but it could use a local offset instead of the canonical UTC `Z` form. Session replay log capture also used logcat's local `threadtime` output, so parsing depended on the device time zone.

This change formats SDK dates in UTC while preserving the exact instant. Session replay now requests logcat's epoch format, passes the capture start time as epoch seconds with milliseconds, and converts each epoch timestamp directly into a UTC calendar. Timestamp API documentation now explains that non-UTC inputs keep the same instant and are serialized in UTC.

- `GsonDateTypeAdapterTest` sets the default time zone to `America/Los_Angeles` and verifies that `2023-07-15T08:30:45.123-07:00` serializes as `2023-07-15T15:30:45.123Z` without changing the input instant.
- `PostHogSerializerTest` sets the default time zone to `America/New_York` and verifies the exact UTC values for both `timestamp` and `sent_at` near a daylight saving transition.
- `LogcatParserTest` verifies exact UTC epoch parsing, independence from the process default time zone, conversion of sub-millisecond epoch fractions to milliseconds, log level and payload parsing, and rejection of ambiguous `threadtime` lines.
- `PostHogLogCatIntegrationTest` verifies the exact `logcat -v epoch -T 1721057445.123 *:E` command.
- Ran the focused core and Android unit tests, `make checkFormat`, `pnpm changeset status --since=origin/main`, and committed-HEAD autoreview against `origin/main`.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

The Pi worker agent prepared the existing implementation for review, added the required changeset, and ran the repository checks and isolated autoreview. The human directed the branch, scope, commit message, PR title, and required validation. This local agent session does not have a shareable session link.

### DIFF
--- a/.changeset/calm-clocks-align.md
+++ b/.changeset/calm-clocks-align.md
@@ -1,0 +1,7 @@
+---
+"posthog": patch
+"posthog-android": patch
+"posthog-server": patch
+---
+
+Normalize SDK event and replay log timestamps to UTC while preserving their exact instants.
--- a/.changeset/calm-clocks-align.md
+++ b/.changeset/calm-clocks-align.md
@@ -1,7 +1,5 @@
 ---
-"posthog": patch
 "posthog-android": patch
-"posthog-server": patch
 ---
 
-Normalize SDK event and replay log timestamps to UTC while preserving their exact instants.
+Make session replay log timestamp handling more robust by parsing timezone-independent logcat epoch timestamps instead of local wall-clock timestamps.
--- a/.changeset/calm-clocks-align.md
+++ b/.changeset/calm-clocks-align.md
@@ -1,5 +1,6 @@
 ---
+"posthog": patch
 "posthog-android": patch
 ---
 
-Make session replay log timestamp handling more robust by parsing timezone-independent logcat epoch timestamps instead of local wall-clock timestamps.
+Clarify that event timestamps are serialized in UTC, and make session replay log timestamp handling more robust by parsing timezone-independent logcat epoch timestamps instead of local wall-clock timestamps.
--- a/posthog-android/src/main/java/com/posthog/android/replay/internal/LogLine.java
+++ b/posthog-android/src/main/java/com/posthog/android/replay/internal/LogLine.java
@@ -12,7 +12,7 @@ import java.util.GregorianCalendar;
 @PostHogInternal
 public class LogLine {
     /**
-     * The timestamp of the event. In UTC even though the device might not have been.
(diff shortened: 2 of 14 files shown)


arnohillen:
<small>(created by claude code)</small>

[arno's agent] Real logcat -v epoch lines start with leading spaces because liblog prints the seconds right-aligned via snprintf("%19lld") (verified in AOSP logprint from android-7.0.0_r1 through Android 14), so LOG_LINE_RE with matcher.matches() never matches and every console log line is dropped; the pattern needs a leading \s* (or trim the line before parse) and a test with a padded line.

arnohillen:
<small>(created by claude code)</small>

[arno's agent] The epoch format modifier only exists from Android 7.0 (absent in android-6.0.1_r81 logprint.c, added in android-7.0.0_r1, same for the -T %s.%q epoch parsing), so on minSdk 23 devices logcat exits with 'Invalid parameter epoch to -v' and console log capture stops working entirely; this needs an SDK_INT gate or a minSdk bump.

marandaneto:
@arnohillen Good catch on the padding. Fixed in d7c357be: the epoch parser now accepts logcat’s leading whitespace, with a regression test using a right-aligned epoch line.

marandaneto:
@arnohillen Session Replay is only supported on API 26+ (`PostHogReplayIntegration.isSupported()` requires `Build.VERSION.SDK_INT >= Build.VERSION_CODES.O`). Console log events are only captured while Session Replay is active, so the API 23 `logcat -v epoch` limitation is outside the supported Session Replay range and does not require an additional gate here.

arnohillen (approved):
<small>(created by claude code)</small>

[arno's agent] Verified at d7c357be: LOG_LINE_RE now tolerates the right-aligned epoch padding with a regression test on a 19-char padded line, and your API 26 argument for the second point checks out (console lines only emit while isSessionReplayActive() is true, and isSupported() requires SDK_INT >= O in install()).

arnohillen:
<small>(created by claude code)</small>

[arno's agent] isSessionReplayActive() can still return true below API 26 because start() sets the active flag without an isSupported() check, so API 23 devices that previously emitted console-only events now stop silently; those recordings have no snapshots so the loss looks acceptable, noting it only because the API 26+ gate is enforced in install() rather than start().

PR #699, opened 2026-08-17
fix: support fractional rollout percentages in local evaluation

marandaneto:

Local feature flag definitions can contain fractional condition rollout percentages such as `0.1` or `0.5`. The condition model stored this field as an integer, so Gson rejected fractional definitions before they could be evaluated locally.

This changes the internal condition model to store `rolloutPercentage` as a `Double`, allowing Gson and the local evaluator to preserve fractional values.

- `./gradlew :posthog:test :posthog-server:test` using JDK 17
- `make checkFormat`
- `make api`
- `pnpm changeset status`
- Added local evaluator coverage for `0.1`, `0.5`, `100.0`, `0.0`, and `null`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi was used to inspect the local-evaluation path, implement the fix, run the test and formatting checks, and perform an isolated autoreview. Autoreview flagged the changed JVM signature, but `FlagConditionGroup` is annotated `@PostHogInternal` and is public only for the SDK's multi-module architecture. The human DRI chose the direct `Double` type instead of adding a compatibility-only integer property to this internal model.

### DIFF
--- a/.changeset/fractional-rollout-percentages.md
+++ b/.changeset/fractional-rollout-percentages.md
@@ -1,0 +1,6 @@
+---
+"posthog": patch
+"posthog-server": patch
+---
+
+Support fractional rollout percentages when evaluating feature flags locally.
--- a/posthog-server/src/test/java/com/posthog/server/internal/FlagEvaluatorTest.kt
+++ b/posthog-server/src/test/java/com/posthog/server/internal/FlagEvaluatorTest.kt
@@ -658,6 +658,45 @@ internal class FlagEvaluatorTest {
     }
 
     @Test
+    internal fun testMatchFeatureFlagPropertiesWithFractionalRollout() {
+        val testCases =
+            listOf(
+                Triple(0.1, "user-2912", true),
+                Triple(0.1, "user-212", false),
+                Triple(0.5, "user-212", true),
+                Triple(100.0, "user-2912", true),
+                Triple(0.0, "user-2912", false),
+                Triple(null, "user-2912", true),
+            )
+
+        for ((rolloutPercentage, distinctId, expected) in testCases) {
+            val json =
+                """
+                {
+                  "id": 1,
+                  "name": "Fractional Rollout Flag",
+                  "key": "fractional-rollout",
+                  "active": true,
+                  "filters": {
+                    "groups": [
+                      {
+                        "properties": [],
+                        "rollout_percentage": $rolloutPercentage
+                      }
+                    ]
+                  },
+                  "version": 1
+                }
+                """.trimIndent()
+
+            val flag = config.serializer.gson.fromJson(json, FlagDefinition::class.java)
+            val result = evaluator.matchFeatureFlagProperties(flag, distinctId, emptyMap())
+
+            assertEquals("rollout $rolloutPercentage for $distinctId", expected, result)
+        }
+    }
+
+    @Test
     internal fun testMatchFeatureFlagPropertiesWithVariant() {
         val flag = createMultiVariateFlag()
         val result = evaluator.matchFeatureFlagProperties(flag, "user-123", emptyMap())
--- a/posthog/src/main/java/com/posthog/internal/PostHogLocalEvaluationModels.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogLocalEvaluationModels.kt
@@ -49,15 +49,10 @@ public class FlagFilters(
 @PostHogInternal
 public class FlagConditionGroup(
     public val properties: List<FlagProperty>?,
-    rolloutPercentage: Int?,
-    public val variant: String?,
-) {
     @SerializedName("rollout_percentage")
-    public val rolloutPercentageDecimal: Double? = rolloutPercentage?.toDouble()
-
-    public val rolloutPercentage: Int?
-        get() = rolloutPercentageDecimal?.toInt()
-}
+    public val rolloutPercentage: Double?,
+    public val variant: String?,
+)
 
 /**
  * A property condition for flag evaluation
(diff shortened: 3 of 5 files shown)


marandaneto:
follow up https://github.com/PostHog/posthog-dotnet/issues/291

posthog-server/src/test/java/com/posthog/server/internal/FlagEvaluatorTest.kt
```diff
@@ -657,6 +657,45 @@ internal class FlagEvaluatorTest {
         assertTrue("Expected ~500 matches, got $matchCount", matchCount in 400..600)
     }
 
+    @Test
+    internal fun testMatchFeatureFlagPropertiesWithFractionalRollout() {
+        val testCases =
+            listOf(
+                Triple(0.1, "user-2912", true),
```
haacked: `suggestion`: Every expected value in this table is tied to the flag key at [line 678](https://github.com/PostHog/posthog-android/blob/30ca802e8f9822a96c1764c2fd1be03ec53e6a5b/posthog-server/src/test/java/com/posthog/server/internal/FlagEvaluatorTest.kt#L678), and the test doesn't document it. For the key `fractional-rollout`, `user-2912` hashes to about 0.0014% and `user-212` to about 0.43%, which is why `user-212` falls outside a 0.1 rollout but inside a 0.5 one. Rename the key and all six rows become wrong with no explanation.

The rest of this file explains non-obvious expectations in comments (lines [650](https://github.com/PostHog/posthog-android/blob/30ca802e8f9822a96c1764c2fd1be03ec53e6a5b/posthog-server/src/test/java/com/posthog/server/internal/FlagEvaluatorTest.kt#L650) and [704](https://github.com/PostHog/posthog-android/blob/30ca802e8f9822a96c1764c2fd1be03ec53e6a5b/posthog-server/src/test/java/com/posthog/server/internal/FlagEvaluatorTest.kt#L704)), and this is the least obvious table in it.

```kotlin
// Expectations are tied to the flag key "fractional-rollout":
// hash("fractional-rollout.user-2912") ~ 0.0000142, hash("fractional-rollout.user-212") ~ 0.00429.
// Renaming the key changes both hashes and invalidates every row below.
val testCases =
    listOf(
        // 0.0000142 <= 0.001, so user-2912 falls inside a 0.1% rollout
        Triple(0.1, "user-2912", true),
        // 0.00429 > 0.001, so user-212 falls outside 0.1% ...
        Triple(0.1, "user-212", false),
        // ... but 0.00429 <= 0.005, so it falls inside 0.5%
        Triple(0.5, "user-212", true),
        Triple(100.0, "user-2912", true),
        Triple(0.0, "user-2912", false),
        // A null rollout means no restriction, so everyone matches
        Triple(null, "user-2912", true),
    )
```

.changeset/fractional-rollout-percentages.md
```diff
@@ -0,0 +1,6 @@
+---
+"posthog": patch
+"posthog-server": patch
+---
+
+Support fractional rollout percentages when evaluating feature flags locally.
```
haacked: `suggestion`: A consumer who upgrades `posthog` without upgrading `posthog-server` hits a `NoSuchMethodError` on the first local flag evaluation, and nothing in the build prevents that pairing. The two artifacts carry independent version lines ([`gradle.properties:25`](https://github.com/PostHog/posthog-android/blob/30ca802e8f9822a96c1764c2fd1be03ec53e6a5b/gradle.properties#L25) and [`:28`](https://github.com/PostHog/posthog-android/blob/30ca802e8f9822a96c1764c2fd1be03ec53e6a5b/gradle.properties#L28)), [`.changeset/config.json`](https://github.com/PostHog/posthog-android/blob/30ca802e8f9822a96c1764c2fd1be03ec53e6a5b/.changeset/config.json) has empty `linked` and `fixed` groups, and [`posthog-server/build.gradle.kts:94`](https://github.com/PostHog/posthog-android/blob/30ca802e8f9822a96c1764c2fd1be03ec53e6a5b/posthog-server/build.gradle.kts#L94) declares `api(project(":posthog"))` which publishes as a required rather than a strict version, so Gradle's highest-wins resolution can pair a new core with an old server. The old `FlagEvaluator` was compiled against `getRolloutPercentage()Ljava/lang/Integer;`, which this PR removes. Patch bumps are the ones dependency bots auto-merge, so the mismatch lands quietly.

The changeset text becomes the published changelog entry for both packages, so noting it there is the cheapest fix. A `minor` bump on both packages would instead put the upgrade in front of a human.

```suggestion
Support fractional rollout percentages when evaluating feature flags locally.

`posthog` and `posthog-server` must be upgraded together for this release. The internal `FlagConditionGroup.rolloutPercentage` changed from `Integer` to `Double`, so a mismatched pair throws `NoSuchMethodError` during local flag evaluation.
```

posthog/src/main/java/com/posthog/internal/PostHogLocalEvaluationModels.kt
```diff
@@ -50,7 +50,7 @@ public class FlagFilters(
 public class FlagConditionGroup(
     public val properties: List<FlagProperty>?,
     @SerializedName("rollout_percentage")
-    public val rolloutPercentage: Int?,
+    public val rolloutPercentage: Double?,
```
haacked: `nit`: Now that this field is a `Double`, `0.1` is expressible for the first time and reads like "10%", but the evaluator divides by 100, so it means one tenth of a percent. Add a note to the KDoc at [line 47](https://github.com/PostHog/posthog-android/blob/30ca802e8f9822a96c1764c2fd1be03ec53e6a5b/posthog/src/main/java/com/posthog/internal/PostHogLocalEvaluationModels.kt#L47): "A condition group with properties and a rollout percentage on a 0-100 scale, so 0.1 means one tenth of a percent."

arnohillen (approved):
<small>(created by claude code)</small>

Verified the only consumer (FlagEvaluator.isConditionMatch) already divides by 100.0 so whole-number behavior is unchanged, Gson reads lossless doubles into Int fields so mixed-version shared-cache fleets stay safe for integral values, and the new test covers the exact fractional failure path; the ABI change is on a @PostHogInternal class with the api dump updated.

dustinbyrne (approved):

PR #703, opened 2026-08-18
feat: skip push registration when the project has no integration for the app_id

dmarchuk:

A device on a project without push configured makes a registration request on every app open that the server can only throw away, and a project that turns push on afterwards never reaches the devices it already has.

- The SDK registers its token at startup whenever push capture is on, against every project it reports to. It has no way to know whether that project opted in.
- `/api/push_subscriptions/` answers those with `200 {"stored": false, "push_enabled": false}` and discards the token.
- Because any 2xx writes `deliveredForDistinctId`, the device then records a success for a token the server never kept and stops asking. No response reaches a client that has stopped asking, so a project that configures push later reaches nothing installed before it — until a token rotation, an `identify()` to a different distinct id, or a reinstall.

Server side is merged: PostHog/posthog#83148 publishes the list, and the contract this implements is `docs/internal/push-subscription-registration.md` there (plus PostHog/posthog#83941, which adds the caching requirement below).

## Changes

Remote config now carries `push.appIds` — the Firebase project ids and APNs bundle ids a project accepts registrations for.

- **The send is gated in `attempt()`**, the choke point that already covers `register`, flush-driven `retryPending`, and the identify resend. The record is still persisted when the send is gated, so `onPushAppIdsChanged` has a token to register later rather than waiting for the app to hand us one again.
- **The list is cached and preloaded on launch**, mirroring `errorTracking`. Registration runs at startup and remote config resolves asynchronously, so without the cache the gate would only take effect from the second launch onward.
- **Absent stays permissive.** A server older than the key sends nothing, which an SDK cannot tell apart from a project with push disabled, so `null` means attempt. Only a published empty list means skip.
- **`onPushAppIdsChanged` clears the delivered marker** for an `app_id` that just became registerable, and re-registers. It fires only on that transition — firing on every config load would put the request back on every launch, which is what the marker exists to prevent.

On the first launch that ever sees the key, the previous list is treated as empty rather than unknown, so every configured `app_id` counts as new. That costs one request per device, once, and it is the only way to reach a device whose project was configured before it updated to an SDK that reads this key.

New unit tests in `PostHogPushSubscriptionManagerTest`:

- An `app_id` outside the list sends nothing but still persists the record. Catches the gate being dropped, and catches the record not being persisted, which would leave `onPushAppIdsChanged` with no token.
- A null list still registers. Catches the gate failing closed against a server older than the key, which would silently disable push on every such deployment.
- An `app_id` inside the list registers. Catches the gate being inverted.
- An `app_id` becoming registerable clears the delivered marker and re-registers. This is the un-stranding path.
- An unrelated `app_id` becoming registerable does not re-register. Catches over-firing.

### Manual E2E on the emulator against a local server

Ran the six-scenario verification brief on an Android emulator (posthog-android-sample as `com.posthog.pushtest`, real FCM token, Firebase project `posthog-dev`) against a local PostHog serving real `push.appIds` remote config for a team with a togglable Firebase integration. `:posthog:test` (62 push-manager cases) and the full suite pass.

- Configured launch: exactly one POST, server store, person property `$device_push_subscription_posthog-dev` set (verified in ClickHouse and the persons UI).
- Unconfigured launch: no POST, "skipped: posthog-dev not configured", `push_subscription.pending` written without a delivered marker.
- Cold start unconfigured: skip logged ~7s before a deliberately delayed `/config` responded, so the disk-cached slice gated it, not the fresh fetch.
- Un-stranding: after the integration was added, the next launch logged "became registerable; re-registering" and sent exactly one POST.
- No repeat: three further relaunches sent nothing.
- Old server: with the `push` key removed from the response, the SDK registered.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Authored with Claude Code, as the reference implementation of the three client rules; iOS mirrors it next, matching how PostHog/posthog-android#656 and PostHog/posthog-ios#735 were kept in step.

Two design points worth a reviewer's attention. The gate lives in `attempt()` rather than `performRegister` so flush and identify resends are covered by the same check. And the transition set is computed in `processPushConfig` from the previously cached list, then drained once by `consumeNewlyRegisterablePushAppIds`, so a device re-registers on the transition and not on every subsequent config load.

I read the server behavior from the merged endpoint rather than assuming it, and the delivered-marker mechanics from this repo's own `performRegister` and `currentRecord`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Bbav6vE7Tm24KkdRhZmvNM)_

### DIFF
--- a/.changeset/push-app-ids-gate.md
+++ b/.changeset/push-app-ids-gate.md
@@ -1,0 +1,5 @@
+---
+"posthog": patch
+---
+
+Skip push token registration when the project has no push integration for the app_id, using the `push.appIds` list published in remote config. A device whose project configures push later re-registers on the next config load rather than staying unreachable.
--- a/.changeset/push-app-ids-gate.md
+++ b/.changeset/push-app-ids-gate.md
@@ -1,5 +1,6 @@
 ---
 "posthog": patch
+"posthog-android": patch
 ---
 
 Skip push token registration when the project has no push integration for the app_id, using the `push.appIds` list published in remote config. A device whose project configures push later re-registers on the next config load rather than staying unreachable.
--- a/posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
+++ b/posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
@@ -13,6 +13,7 @@ import com.posthog.internal.PostHogPreferences.Companion.FEATURE_FLAG_EVALUATED_
 import com.posthog.internal.PostHogPreferences.Companion.FEATURE_FLAG_REQUEST_ID
 import com.posthog.internal.PostHogPreferences.Companion.FLAGS
 import com.posthog.internal.PostHogPreferences.Companion.MINIMAL_FLAG_CALLED_EVENTS
+import com.posthog.internal.PostHogPreferences.Companion.PUSH
 import com.posthog.internal.PostHogPreferences.Companion.SESSION_REPLAY
 import com.posthog.internal.PostHogPreferences.Companion.SURVEYS
 import com.posthog.surveys.Survey
@@ -130,6 +131,14 @@ public class PostHogRemoteConfig(
     @Volatile
     private var autoCaptureExceptions = false
 
+    // The app_ids this project accepts push registrations for, or null when we have never heard
+    // from a server that sends the key. Null means "attempt the registration" — an SDK cannot tell an
+    // unconfigured project from a server older than the key, so absent has to stay permissive.
+    @Volatile private var pushAppIds: List<String>? = null
+
+    // app_ids that became registerable on the last /config read, drained by the push manager.
+    @Volatile private var newlyRegisterablePushAppIds: Set<String> = emptySet()
+
     // Set by preloadErrorTrackingConfig when a disk-cached error-tracking config exists at startup.
     // Survives clear()/reset(): it's project-level config, not user data, like the cached config.
     @Volatile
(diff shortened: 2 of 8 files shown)


posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
```diff
@@ -639,6 +650,70 @@ public class PostHogRemoteConfig(
...
+        val appIds =
+            when (push) {
+                is Map<*, *> -> (push["appIds"] as? List<*>)?.mapNotNull { it as? String }
+                else -> null
+            }
+
+        pushAppIds = appIds
+        if (persist) {
+            if (appIds != null) {
+                config.cachePreferences?.setValue(PUSH, mapOf("appIds" to appIds))
```
marandaneto: blocking: Persist the transition only after the delivered marker is durably cleared

This persists the new `appIds` before `onRemoteConfigLoaded` asynchronously queues clearing `deliveredForDistinctId`. If the process exits or the SDK closes in that window, the next launch preloads the new list, no longer detects the transition, and the stale delivered marker permanently suppresses registration. Please persist the transition itself, or delay advancing the cached `appIds` until the marker is durably cleared.

.changeset/push-app-ids-gate.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog": patch
```
marandaneto, line 2: you need an android changeset here as well

marandaneto (approved):

posthog/src/main/java/com/posthog/internal/PostHogRemoteConfig.kt
dmarchuk: Updated, thank you!

PR #704, opened 2026-08-18
feat(gradle-plugin): add event release mode via posthog.releaseMode

ablaszkiewicz:

The mapping upload binds the release it creates to the mapping. Map ids are derived from the mapping's content, so two releases shipping the same mapping both report whichever one uploaded it first.

`posthog.releaseMode=event` (or `POSTHOG_RELEASE_MODE=event`, which posthog-cli and the bundler plugins already read) passes `--release-mode event` to posthog-cli, leaving the mapping unbound. Nothing has to be injected into the app: the server resolves each event's release from the `$app_namespace` / `$app_version` / `$app_build` the SDK already sends — same shape as posthog-ios with `dsym upload --no-release-bind`.

The flag is only passed outside the default mode, so symbol-set builds keep working against a posthog-cli predating it. An unrecognised mode fails the build rather than silently falling back.

Needs PostHog/posthog#84478.

### Before

| Symol set (bound to a release) | Issue shows release |
|--------|--------|
| CleanShot 2026-08-18 at 16 28 51 | CleanShot 2026-08-18 at 16 29 19 | 

### After

| Symol set (NOT bound to a release) | Issue STILL shows release (resolved via cymbal) |
|--------|--------|
| CleanShot 2026-08-18 at 16 31 31 | CleanShot 2026-08-18 at 16 30 39
 | 
 
 ### CLI output AFTER

In the new mode I uploaded source maps and then modified version without changing anything in the code. That means we should create new release but still skip the upload. In previous version this was not possible.

| New release but same map skips upload |
|--------|
| CleanShot 2026-08-18 at 16 35 12@2x | 

Assembled the sample app in each mode and checked the CLI invocation: `-Pposthog.releaseMode=event` and `POSTHOG_RELEASE_MODE=event` both append `--release-mode event`, the default appends nothing, and `=events` fails the build. No tests — the module has no test infrastructure.

## :pencil: Checklist

### DIFF
--- a/posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadProguardMappingsTask.kt
+++ b/posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadProguardMappingsTask.kt
@@ -62,6 +62,12 @@ public abstract class PostHogUploadProguardMappingsTask : PostHogCliExecTask() {
         if (!mappingsFiles.isPresent || mappingsFiles.get().isEmpty) {
             error("[PostHog] Mapping files are missing!")
         }
+        // posthog-cli reads POSTHOG_RELEASE_MODE itself when --release-mode is absent, and the
+        // Gradle daemon's environment may carry one that contradicts the mode resolved here —
+        // where the gradle property is supposed to win. Pin the resolved mode in the child's
+        // environment so it cannot be overridden by inheritance. A posthog-cli predating the flag
+        // has no such argument and ignores the variable, so this stays safe for old CLIs.
+        releaseMode.orNull?.let { environment(POSTHOG_RELEASE_MODE_ENV, it) }
         super.exec()
     }
 
--- a/posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadProguardMappingsTask.kt
+++ b/posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadProguardMappingsTask.kt
@@ -53,6 +53,11 @@ public abstract class PostHogUploadProguardMappingsTask : PostHogCliExecTask() {
     @get:Optional
     public abstract val build: Property<Int>
 
+    /** [PostHogReleaseMode.cliValue], as the enum itself is internal to the plugin. */
+    @get:Input
+    @get:Optional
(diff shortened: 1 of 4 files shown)


posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadProguardMappingsTask.kt
```diff
@@ -94,6 +99,12 @@ public abstract class PostHogUploadProguardMappingsTask : PostHogCliExecTask() {
             args.add("--build")
             args.add(it.toString())
         }
+        // Passed only outside the default mode, so a symbol-set build keeps working against a
+        // posthog-cli predating the flag.
+        releaseMode.orNull?.takeIf { it != PostHogReleaseMode.SYMBOL_SET.cliValue }?.let {
```
marandaneto, line 110: blocking: Gradle property cannot override an inherited event release mode

When `POSTHOG_RELEASE_MODE=event` is inherited and `-Pposthog.releaseMode=symbol-set` is explicitly set, this condition omits the CLI argument, so the inherited environment still selects `event`. That violates the documented Gradle-property precedence and can upload an unbound mapping. Please explicitly enforce `symbol-set` in this conflict and add coverage for the property/environment combination.

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadProguardMappingsTask.kt
```diff
@@ -94,6 +99,12 @@ public abstract class PostHogUploadProguardMappingsTask : PostHogCliExecTask() {
...
+            args.add("--release-mode")
+            args.add(it)
+        }
```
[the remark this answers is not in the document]
ablaszkiewicz: Fixed

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadProguardMappingsTask.kt
ablaszkiewicz: Fixed

marandaneto (approved):

cat-ph (approved):
let's go!

PR #717, opened 2026-08-22
fix(server): respect network and retry status during flush

marandaneto:

An explicit server queue flush could send events while the configured network status was offline or while the queue was waiting after a failed request. Timer and threshold flushes already respected both conditions. Explicit flushes should follow the same eligibility rules without dropping queued events.

This change checks queue eligibility before every batch in an explicit drain. Rechecking between batches also stops a long flush if connectivity changes while it is running. The existing flushing guard is still reset after every exit path.

- `./gradlew :posthog-server:test --tests 'com.posthog.server.internal.PostHogMemoryQueueTest' --rerun-tasks`
- `make checkFormat`
- Autoreview against `origin/main`

Added tests for an initially offline explicit flush, a disconnect between drained batches, and a retry pause that later expires.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

The developer asked Pi's worker agent to fix the queue eligibility findings from a duplication report. The implementation follows the existing file-backed queue behavior. Pi autoreview identified that connectivity also needed to be checked between drained batches, so the code and regression coverage were updated before opening this PR.

### DIFF
--- a/posthog-server/src/main/java/com/posthog/server/internal/PostHogMemoryQueue.kt
+++ b/posthog-server/src/main/java/com/posthog/server/internal/PostHogMemoryQueue.kt
@@ -72,6 +72,14 @@ internal class PostHogMemoryQueue(
     }
 
     override fun flush() {
+        flush(ignoreRetryPause = false)
+    }
+
+    override fun flushForShutdown() {
+        flush(ignoreRetryPause = true)
+    }
+
+    private fun flush(ignoreRetryPause: Boolean) {
         if (isFlushing.getAndSet(true)) {
             config.logger.log("Queue is flushing.")
             return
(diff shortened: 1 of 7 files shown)


posthog-server/src/main/java/com/posthog/server/internal/PostHogMemoryQueue.kt
```diff
@@ -81,8 +81,8 @@ internal class PostHogMemoryQueue(
         // rather than racing ahead of them and seeing an empty queue
         executor.submitSyncSafely {
             try {
-                while (isAboveThreshold(1) && executeBatch()) {
-                    // Keep draining successful batches until the queue is empty.
+                while (isAboveThreshold(1) && canFlushBatch() && executeBatch()) {
```
ioannisj: [suggestion] I think for `close()` path we may want to skip the pause here? There's no next tick for close() and if this was a transient 5xx error we would lose those events?

ioannisj (approved):

posthog-server/src/main/java/com/posthog/server/internal/PostHogMemoryQueue.kt
marandaneto: mm good point

marandaneto: Good catch! Fixed in 3411b641 — shutdown now uses a queue-level `flushForShutdown()` hook that bypasses retry backoff while still respecting network availability and stopping after a failed request. Added regression coverage for retrying a pending event during `close()`.

PR #722, opened 2026-08-23
fix(replay): keep manually started replay active after session rotation

### LINKED ISSUE #721: Session replay captures zero snapshots from 3.43.2 onward, while isSessionReplayActive() reports true
### Version

3.43.2 (regression from 3.43.1; still present in 3.60.4)

### Steps to Reproduce

Session replay stops capturing entirely in 3.43.2. `3.43.1` and `3.43.2` differ by a single commit — `590d694`, *"Enforce 24-hour maximum session duration and 30-minute inactivity rotation with automatic session rotation, mirroring iOS"* — which is the only entry in the 3.43.2 changelog.

Our setup defers replay start rather than enabling it at init:

```kotlin
PostHogAndroid.setup(
    this,
    PostHogAndroidConfig(apiKey = "<key>", host = "https://us.i.posthog.com").apply {
        sessionReplay = false               // started manually below
        sessionReplayConfig.screenshot = false   // wireframe mode
        sessionReplayConfig.maskAllImages = false
        sessionReplayConfig.maskAllTextInputs = false
        sampleRateProvider = { 1.0 }        // forced to 1.0 for this test
    }
)

// from Application.onCreate, in a coroutine
PostHog.startSessionReplay(false)
```

1. Build an app with the above against **3.43.1**, on a device/emulator (Android 14, API 34, x86_64).
2. Launch it, interact for >30s (so the project's `minimumDurationMilliseconds` of 30000 is satisfied), background it.
3. Observe `$snapshot` events in logcat and a recording in the project.
4. Change only the dependency version to **3.43.2**. Rebuild, uninstall, reinstall, repeat the identical flow.

Everything else held constant: same app code, same device, same interaction flow, same forced sample rate, same project.

### Expected Result

`$snapshot` events are captured and POSTed to `/s/`, and a recording appears — as happens on 3.43.1:

```
D PostHog : Queued Event $snapshot: ....event.      (x20)
D PostHog : Request https://us.i.posthog.com/s/}: [{"api_key":"<redacted>", ...
            "event":"$snapshot","properties":{"$lib":"posthog-android",
            "$lib_version":"3.43.1","$snapshot_source":"mobile",
            "$snapshot_data":[{"data":{"wireframes":[...]}}]}}]
```

Recording retrievable, 62s and 20s in two separate runs.

### Actual Result

On 3.43.2 (and 3.60.4): **zero `$snapshot` events, no request to `/s/`, no recording** — while ordinary analytics events queue and flush normally in the same session.

Critically, the SDK reports replay as running:

```
D ForceReplay : replayActive=true after 1 attempt
D PostHog     : [Session Replay] Session changed. Re-initializing recording for new session.
```

`PostHog.isSessionReplayActive()` returns `true`, `startSessionReplay()` throws nothing, and no error or warning is logged. Nothing distinguishes this from a working setup except that no snapshot is ever produced. That silence is most of the cost of the bug — we shipped a version bump past 3.43.1 and lost all replay in production without any signal.

Ruled out during investigation:

- **Sampling.** Forced via `sampleRateProvider = { 1.0 }` (the property `makeSamplingDecision` actually reads, taking precedence over remote config). The SDK does not log `Sample rate (x) has determined that this sessionId (y) will not be sent to the server`.
- **Session duration.** Sessions ran 45–75s against a 30s `minimumDurationMilliseconds`.
- **Network.** Ordinary events POST successfully in the same session; remote config is fetched and applied.
- **Replay never starting.** Retried until `isSessionReplayActive()` returned `true` before asserting.

### Possible mechanism

The 3.43.2 changelog notes that `PostHogSessionManager.isAppInBackground` now **defaults to `true`** until the first lifecycle `onStart` flips it. We call `startSessionReplay(false)` from a coroutine launched in `Application.onCreate`, which can run before that first `onStart` — and our launch path enters an Activity directly via an intent rather than through a launcher icon. If the new rotation logic treats a session begun while `isAppInBackground == true` as ineligible, that would match the symptom exactly: replay reports active, the session rotates, and no frame is ever captured.

That is a hypothesis from the changelog and the observed behaviour, not something we have confirmed in your code.

### Environment

- `com.posthog:posthog-android` 3.43.1 (working) vs 3.43.2 / 3.60.4 (broken)
- Android 14 (API 34), x86_64 emulator
- Kotlin, AGP 8.13.2, Gradle 8.13
- US Cloud, wireframe replay mode (`screenshot = false`)

marandaneto:

Calling `PostHog.startSessionReplay(false)` with automatic replay disabled creates a new session. The session-change callback is queued before the manual start completes. When that callback runs, it sees `sessionReplay = false` and stops the recording, so `isSessionReplayActive()` briefly reports `true` but no snapshots are captured.

This change tracks whether recording started while automatic replay was disabled. Session changes and remote-config updates preserve that manual recording, including the first live config response during cold start. An inactive recorder is not started when `sessionReplay` is false, and an automatically started recording still stops if that setting changes to false. Project replay flags and sampling decisions continue to apply.

Fixes #721.

- `./gradlew :posthog-android:test`
- `make checkFormat`
- Added regression tests for manual replay during session rotation and the queued callback ordering.
- Android 15 emulator in wireframe mode with `sessionReplay = false` and a main-thread `startSessionReplay(false)` call. Before the fix, replay reported active but sent no snapshot requests. With the fix, it sent the full wireframe and interaction snapshots to `/s/`.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi investigated the regression, reproduced both the queued callback and first remote-config paths, and implemented the fix. Recording state is published atomically so background callers cannot expose a partially updated manual-start state. The project replay flag and sampling checks remain in place.

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -2418,14 +2418,14 @@ public class PostHogReplayIntegration(
 
     /**
      * Whether the live (or, on a fetch failure, disk-cached) remote config permits recording the
-     * current session right now. Mirrors the decision [reevaluateRecordingState] makes — master
-     * switch, session-replay flag, event triggers, and the sampling decision — so the buffered
+     * current session right now. Mirrors the decision [reevaluateRecordingState] makes — automatic
+     * start setting, session-replay flag, event triggers, and the sampling decision — so the buffered
      * opening window is migrated only when the session is genuinely recordable, never for one the
      * fresh config samples out.
      */
     private fun isRecordingPermittedForCurrentSession(): Boolean {
         val remoteConfig = config.remoteConfigHolder ?: return false
-        if (!config.sessionReplay || !remoteConfig.isSessionReplayFlagActive()) {
+        if ((!config.sessionReplay && !isSessionReplayActive) || !remoteConfig.isSessionReplayFlagActive()) {
             return false
         }
         if (shouldWaitForEventTriggers()) {
@@ -2436,9 +2436,10 @@ public class PostHogReplayIntegration(
     }
 
     /**
-     * Re-evaluate recording against the live remote config: stop when session replay is no longer
-     * permitted (master switch off, flag off, or sampled out) and resume when it turns on. Without
-     * this, a fresh-`false` would keep recording until the next session rotation.
+     * Re-evaluate recording against the live remote config: preserve a manual recording when
+     * automatic replay is off, stop when the project flag turns off or the session is sampled out,
+     * and automatically resume only when automatic replay is enabled. Without this, a fresh-`false`
+     * would keep recording until the next session rotation.
      *
      * The very first delivery is exempt from the **flag-off** stop only: on that delivery
      * [resolveFirstRemoteConfig] already owns the stop-and-drop decision for the buffered opening
@@ -2449,7 +2450,7 @@ public class PostHogReplayIntegration(
         val postHog = this.postHog ?: return
         val remoteConfig = config.remoteConfigHolder ?: return
 
-        if (!config.sessionReplay || !remoteConfig.isSessionReplayFlagActive()) {
+        if ((!config.sessionReplay && !isSessionReplayActive) || !remoteConfig.isSessionReplayFlagActive()) {
             if (!isFirstDelivery) {
                 stopIfActive("Remote config disabled recording. Stopping.")
             }
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -180,6 +180,9 @@ public class PostHogReplayIntegration(
     @Volatile
     private var isSessionReplayActive: Boolean = false
 
+    @Volatile
+    private var isManualSessionReplayActive: Boolean = false
+
     // Event triggers for session recording
     private val eventTriggersLock = Any()
 
@@ -2091,6 +2094,9 @@ public class PostHogReplayIntegration(
         resetSessionStateIfNeeded(currentSessionId, force = !resumeCurrent)
 
         isSessionReplayActive = true
+        // Automatic setup never starts while this setting is false. A start in that state comes
+        // from the manual API or an event trigger and must survive automatic-start checks.
+        isManualSessionReplayActive = !config.sessionReplay
 
         if (!resumeCurrent) {
             // Without this, on a static UI the first user-driven onDraw can be tens of seconds
@@ -2115,6 +2121,7 @@ public class PostHogReplayIntegration(
 
     override fun stop() {
         isSessionReplayActive = false
+        isManualSessionReplayActive = false
         synchronized(decorViews) {
             decorViews.values.forEach { it.drawState.invalidateMaskCapture() }
         }
@@ -2204,7 +2211,8 @@ public class PostHogReplayIntegration(
         mainHandler.handler.post {
             // config.sessionReplay controls automatic starts. An active replay may have been
             // started manually, so preserve it across rotation even when automatic replay is off.
-            if (!config.sessionReplay && !isSessionReplayActive) {
+            if (!config.sessionReplay && !isManualSessionReplayActive) {
+                if (isSessionReplayActive) stop()
                 return@post
             }
             if (remoteConfig?.isSessionReplayFlagActive() != true) {
@@ -2425,7 +2433,7 @@ public class PostHogReplayIntegration(
      */
     private fun isRecordingPermittedForCurrentSession(): Boolean {
         val remoteConfig = config.remoteConfigHolder ?: return false
-        if ((!config.sessionReplay && !isSessionReplayActive) || !remoteConfig.isSessionReplayFlagActive()) {
+        if ((!config.sessionReplay && !isManualSessionReplayActive) || !remoteConfig.isSessionReplayFlagActive()) {
             return false
         }
         if (shouldWaitForEventTriggers()) {
@@ -2450,7 +2458,7 @@ public class PostHogReplayIntegration(
         val postHog = this.postHog ?: return
         val remoteConfig = config.remoteConfigHolder ?: return
 
-        if ((!config.sessionReplay && !isSessionReplayActive) || !remoteConfig.isSessionReplayFlagActive()) {
+        if ((!config.sessionReplay && !isManualSessionReplayActive) || !remoteConfig.isSessionReplayFlagActive()) {
             if (!isFirstDelivery) {
                 stopIfActive("Remote config disabled recording. Stopping.")
             }
@@ -2494,6 +2502,7 @@ public class PostHogReplayIntegration(
             // persisting immediately (PostHogReplayQueue.shouldPersist reads isActive), instead of
             // leaking snapshots to the send queue in the window before the posted stop() runs on main.
             isSessionReplayActive = false
+            isManualSessionReplayActive = false
             mainHandler.handler.post { stop() }
         }
     }
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -177,11 +177,20 @@ public class PostHogReplayIntegration(
             color = Color.BLACK
         }
 
-    @Volatile
-    private var isSessionReplayActive: Boolean = false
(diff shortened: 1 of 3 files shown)


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -2202,11 +2202,9 @@ public class PostHogReplayIntegration(
...
         mainHandler.handler.post {
-            // config.sessionReplay is the customer-facing master switch: a config-level disable
-            // must not be overridden by remote flag + sampling. Manual PostHog.startSessionReplay
-            // calls and trigger-matched starts go through different code paths and are unaffected.
-            if (!config.sessionReplay) {
-                if (isSessionReplayActive) stop()
+            // config.sessionReplay controls automatic starts. An active replay may have been
+            // started manually, so preserve it across rotation even when automatic replay is off.
+            if (!config.sessionReplay && !isSessionReplayActive) {
                 return@post
```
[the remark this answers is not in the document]
marandaneto: Fixed in ed2ea375. Manual recording state is now tracked explicitly and preserved through both the first and later remote-config deliveries. The project replay flag and sampling checks still apply.

posthog-android/src/test/java/com/posthog/android/replay/PostHogReplayIntegrationTest.kt
```diff
@@ -435,7 +433,39 @@ internal class PostHogReplayIntegrationTest {
             sut.onSessionIdChanged()
             shadowOf(Looper.getMainLooper()).idle()
 
-            assertFalse(sut.isActive())
+            assertTrue(sut.isActive())
+        } finally {
+            sut.uninstall()
+        }
```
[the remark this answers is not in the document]
marandaneto: No code change needed here. PostHogReplayIntegration has no clear() method, and the existing tests for this integration consistently use uninstall() for cleanup. These tests also await replay-executor work before assertions, matching the surrounding test pattern.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -2114,7 +2128,7 @@ public class PostHogReplayIntegration(
     }
 
     override fun stop() {
```
ioannisj: [suggestion] I think the manual intent needs to live somewhere `stop()` doesn't touch. `recordingState` goes back to `INACTIVE` here, so the fact that someone explicitly asked for this recording is gone with it.

That means the fix only holds until the first stop:
- in `onSessionIdChanged`, manual survives the `config.sessionReplay` gate, but `makeSamplingDecision` right below it can still stop us, and on the next rotation `isManualSessionReplayActive` is already false so we take the `!config.sessionReplay` branch and return
- in `reevaluateRecordingState`, `stopIfActive` on a flag-off clears it too, and the resume path is gated on `config.sessionReplay`, which is false in exactly this scenario, so it never comes back

posthog-js keeps this as sticky persisted state rather than a live flag, `session-recording.ts:346-370` registers the sampling/linked-flag/trigger overrides into persistence so a stop then re-evaluate still sees the intent.

Would a separate `@Volatile manualStartRequested`, set by the manual entry point and only cleared by an explicit `PostHog.stopSessionReplay()`, work here? `isManualSessionReplayActive` would read that instead of `recordingState == MANUAL`.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -2090,7 +2102,9 @@ public class PostHogReplayIntegration(
         val currentSessionId = postHog?.getSessionId()?.toString()
         resetSessionStateIfNeeded(currentSessionId, force = !resumeCurrent)
 
-        isSessionReplayActive = true
+        // Automatic setup never starts while this setting is false. A start in that state comes
+        // from the manual API or an event trigger and must survive automatic-start checks.
```
ioannisj: [nit] This labels any start as `MANUAL` whenever the switch is off, so an event-trigger start gets tagged manual too. Harmless today since the trigger path already stops at the `activatedSession != currentSessionId` guard above before we reach here, but the enum ends up claiming something we didn't actually observe. Worth passing the reason into `start()` (or reading the `manualStartRequested` flag from the other comment) so it's true by construction?

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -177,8 +177,20 @@ public class PostHogReplayIntegration(
             color = Color.BLACK
         }
 
+    private enum class RecordingState {
+        INACTIVE,
+        AUTOMATIC,
+        MANUAL,
```
ioannisj, line 183: If there's a flag call in flight (that will activate the recording) it will undo a manual stop when it lands right? Very small window but I think the code allows for that since a stop transitions to INACTIVE, then on flag response we basically check (recordingState == RecordingState.INACTIVE) and resume

ioannisj (approved):

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
marandaneto: Yes, you're right. This behavior predates this PR: the previous boolean state also let reevaluateRecordingState restart any inactive recorder when an in-flight flag response arrived. Fixing it correctly requires tracking manual-stop provenance so automatic stops can still resume while explicit stops cannot. I'll keep that as a separate change rather than expand this session-rotation regression fix.

marandaneto: @ioannisj @turnipdabeets 
i think this should be tested on ios, flutter and rn as well since the changes related to https://github.com/PostHog/posthog-android/pull/722 is not this repo only
pls double check all SDKs

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -2114,7 +2128,7 @@ public class PostHogReplayIntegration(
     }
 
     override fun stop() {
-        isSessionReplayActive = false
+        recordingState = RecordingState.INACTIVE
```
turnipdabeets: [question] `stop()` clears the MANUAL marker, so once anything stops us the `!config.sessionReplay` gate locks manual replay out for the rest of the process. Confirmed on a Pixel 9 with `SESSION_INACTIVITY_DURATION` shortened to 30s: after 45s backgrounded I get `Session cleared. Stopping recording.` and then zero `/s/` requests for the rest of the run, while ordinary events keep flowing — the automatic path recovers from the same clear. Is that in scope here, or worth a follow-up?

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -2202,10 +2216,9 @@ public class PostHogReplayIntegration(
...
-            // config.sessionReplay is the customer-facing master switch: a config-level disable
-            // must not be overridden by remote flag + sampling. Manual PostHog.startSessionReplay
-            // calls and trigger-matched starts go through different code paths and are unaffected.
-            if (!config.sessionReplay) {
+            // config.sessionReplay controls automatic starts. An active replay may have been
+            // started manually, so preserve it across rotation even when automatic replay is off.
+            if (!config.sessionReplay && !isManualSessionReplayActive) {
                 if (isSessionReplayActive) stop()
                 return@post
             }
```
turnipdabeets: [suggestion] This reads `recordingState` twice — once via `isManualSessionReplayActive`, once via `isSessionReplayActive` — so a manual `start()` landing between them would stop a replay that just started. Very narrow window compared to what this PR closes, but a local read makes it airtight:
```suggestion
            val state = recordingState
            if (!config.sessionReplay && state != RecordingState.MANUAL) {
                if (state != RecordingState.INACTIVE) stop()
                return@post
            }
```

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -177,8 +177,20 @@ public class PostHogReplayIntegration(
             color = Color.BLACK
         }
 
+    private enum class RecordingState {
```
turnipdabeets: [nit] `MANUAL` also covers event-trigger starts — the comment at the assignment says so, but the name reads narrower. Worth putting it on the enum where people will look it up:
```suggestion
    // MANUAL covers any start made while config.sessionReplay is false — the manual
    // startSessionReplay API and event-trigger starts alike.
    private enum class RecordingState {
```

PR #732, opened 2026-08-25
fix(server): bound missing flag probes

marandaneto:

The server SDK retains clean remote omissions so a missing flag does not trigger a `/flags` request on every evaluation. That retained state was unbounded, and a later response returning the flag did not clear an earlier omission. A delayed omission could also replace newer positive evidence.

This completes the bounded missing-key probe behavior defined in [sdk-specs#48](https://github.com/PostHog/sdk-specs/pull/48).

The missing and remotely returned key stores now use finite-capacity, access-ordered maps. Evicted keys become eligible for another probe. Every current-generation remote response clears retained omissions for keys it returns, including unscoped responses. Evidence sequence numbers prevent delayed responses from replacing newer positive evidence.

Concurrent callers for the same unknown key now remain coordinated until the owner request finishes. An interrupted waiter returns its partial local result instead of starting a duplicate `/flags` request.

- `./gradlew :posthog-server:test --tests "com.posthog.server.internal.PostHogFeatureFlagsTest"`
- `JAVA_TOOL_OPTIONS=-Dnet.bytebuddy.experimental=true ./gradlew :posthog-server:test`
- `make checkFormat`
- `git diff --check`

The full server suite uses the Byte Buddy experimental flag because the currently pinned Byte Buddy version otherwise rejects Java 21.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent after auditing the merged specification and the existing `posthog-server` probe coordination. Reviewer feedback led to using the package's existing access-ordered map pattern, reconciling positive evidence from every remote response, and removing the timeout path that could start a duplicate probe. Pi's bundled autoreview checked the final committed branch and reported no actionable findings.

### DIFF
--- a/posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
+++ b/posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
@@ -44,8 +44,8 @@ internal class PostHogFeatureFlags(
         )
 
     private val missingFlagKeysLock = Object()
-    private val knownMissingFlagKeys = linkedSetOf<String>()
-    private val knownRemoteFlagKeys = linkedMapOf<String, Long>()
+    private val knownMissingFlagKeys = boundedFlagEvidenceMap<Unit>()
+    private val knownRemoteFlagKeys = boundedFlagEvidenceMap<Long>()
     private val inFlightMissingFlagProbes = mutableMapOf<String, MissingFlagProbe>()
     private var missingFlagKeysGeneration: Long = 0
     private var remoteFlagEvidenceSequence: Long = 0
@@ -260,7 +260,8 @@ internal class PostHogFeatureFlags(
 
         fun await(): Boolean =
             try {
-                done.await(MISSING_FLAG_PROBE_WAIT_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+                done.await()
+                true
             } catch (_: InterruptedException) {
                 Thread.currentThread().interrupt()
                 false
@@ -412,6 +413,7 @@ internal class PostHogFeatureFlags(
         }
 
         return try {
+            val responseGeneration = synchronized(missingFlagKeysLock) { missingFlagKeysGeneration }
             val response =
                 api.flags(
                     distinctId,
@@ -432,6 +434,7 @@ internal class PostHogFeatureFlags(
                 computeResponseError(response),
             )
             if (response != null) {
+                reconcileReturnedFlagEvidence(responseGeneration, response)
                 onResponse?.invoke(response)
             }
             flags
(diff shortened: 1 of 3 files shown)


posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
```diff
@@ -1230,52 +1234,80 @@ internal class PostHogFeatureFlags(
...
-                    knownRemoteFlagKeys.remove(key)
-                    knownMissingFlagKeys.add(key)
-                }
+            plan.currentlyMissing.filterTo(mutableSetOf()) { it in returned }.forEach {
+                retainKnownRemoteFlagKeyLocked(it)
             }
         }
     }
 
+    private fun retainKnownMissingFlagKeyLocked(key: String) {
```
ioannisj, line 1259: [suggestion] Would it be worth switching both stores to the `LinkedHashMap(16, 0.75f, true)` + `removeEldestEntry` shape that `PostHogFeatureFlagCache.kt:12-21` already uses in this package? It'd drop both hand-rolled eviction loops and the remove-then-re-add.

ioannisj (approved):
Just one non-blocking suggestion my agent picked up

dustinbyrne (commented):
Most of the specification work looks correct. Capacity limits, eviction, refresh invalidation, retries, and delayed-response protection are present.

I found two edge cases that prevent full compliance with `sdk-specs#48`:

### Clear stored missing-key state for all remote responses

The remote call at `PostHogFeatureFlags.kt:1096` does not supply `onResponse`. Therefore, it does not clear stored missing-key state when the response returns that key.

Example:

1. The SDK asks `/flags` for a specific key.
2. The response does not contain the key.
3. The SDK stores the key as missing.
4. A later unscoped `/flags` response contains the key.
5. A later scoped evaluation still treats the key as missing.

The specification requires any current response that contains the key to clear the older missing-key state.

Could we reconcile returned keys for every remote response while keeping the existing generation check?

### Do not start a second probe while the first probe is active

When the ten-second wait expires at `PostHogFeatureFlags.kt:1141`, this path starts another `/flags` request through `fetchUncoordinatedFlags`.

I reproduced this by keeping the first request active. The second request started after approximately 10.9 seconds.

The specification requires concurrent evaluations for the same key to share one probe while that probe is active.

Could this path continue waiting or return the partial local result instead of starting another request?

These conditions are edge cases, but they are reachable and conflict with the specification.

Human-driven, agent-assisted review.

posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
marandaneto: Good suggestion. Fixed in c3f77f41 by switching both evidence stores to bounded access-ordered `LinkedHashMap` instances with `removeEldestEntry`.

marandaneto:
@dustinbyrne Thanks for catching both edge cases. Fixed in c3f77f41. Every remote response now captures its definitions generation and reconciles returned keys with retained evidence, including unscoped responses. Same-key waiters now remain coordinated until the owner finishes; an interrupted waiter returns its partial local result instead of starting a duplicate request. I added regression tests for both paths.

dustinbyrne (approved):

PR #735, opened 2026-08-27
fix(flags): align local string matching with the flags service

### LINKED ISSUE #78019: Case folding and numeric stringification are inconsistent between SDK local evaluation and the flags service
While shipping `starts_with`/`ends_with` local evaluation across the SDKs (#72992), a pattern emerged: our case-insensitive property matching is inconsistent between the SDKs and the flags service, and within the flags service itself. Filing this so we can settle it once instead of per code review.

## How the flags service folds case

In `rust/feature-flags/src/properties/property_matching.rs`:

- `icontains` / `starts_with` / `ends_with` (and negations): `to_ascii_lowercase()`, ASCII-only. The code comment says it's a deliberate perf choice.
- `exact` / `is_not`: full Unicode `to_lowercase()`.

So the server already applies two different folding rules depending on operator.

## What the SDKs use for the icontains/starts_with/ends_with family

| SDK | Folding | Matches server? |
|---|---|---|
| PHP | `strtolower` (ASCII-only) | ✅ |
| Ruby | `downcase(:ascii)` as of PostHog/posthog-ruby#229 | ✅ |
| Elixir | ASCII-only in PostHog/posthog-elixir#192 (pending) | ✅ |
| Python | `casefold()` (most aggressive Unicode; `"ss"` matches `"ß"`) | ❌ |
| Go | `strings.ToLower` (Unicode) | ❌ |
| Node | `toLowerCase()` (Unicode) | ❌ |
| Rust SDK | `to_lowercase()` (Unicode) | ❌ |
| .NET | `OrdinalIgnoreCase` (Unicode simple folding) | ❌ |

Concretely: person property `"Äbc"` with `starts_with: "ä"` matches locally in Python/Go/Node/Rust/.NET and does **not** match on `/flags`. Same flag, same person, different answer depending on evaluation path — exactly the divergence local evaluation is supposed to avoid.

## Adjacent inconsistencies in the same code

- The Rust SDK's `exact` operator uses `eq_ignore_ascii_case` while the server's `exact` is Unicode — the mirror image of the divergence above.
- Numeric stringification: the server renders a float-typed `323.0` as `"323.0"` (serde `Display`), as do Ruby and Python, but Go and Node render `"323"`, so a JSON `323.0` property matches `ends_with: "3"` locally but not remotely. .NET additionally used the host locale for decimal separators (`3.14` → `"3,14"` on de-DE), fixed in PostHog/posthog-dotnet#270.

## Options

1. **Standardize the SDKs on ASCII-only folding** for the icontains/starts/ends family to match the server. PHP and Ruby are already there; the rest are one-line changes plus tests. The numeric stringification fixes for Go/Node belong in the same sweep.
2. **Make the server Unicode-aware** and leave the SDKs alone. Friendlier for international text, but changes existing server behavior, and the ASCII choice was deliberate.
3. **Leave it and document** the edge case.

I lean toward 1, plus an explicit decision on whether the server's Unicode `exact`/`is_not` is intentional. Non-ASCII values differing only by case are rare, but today the answer depends on which SDK and which operator you hit, and that's the worst version.

marandaneto:

The `posthog-server` local evaluator could disagree with the released `/flags` service for exact matching, JSON values, and case-insensitive string operators.

This fixes PostHog/posthog#78019 by matching the current Rust backend contract:

- `exact` and `is_not` apply the backend's boolean-like filter gate before ordinary array membership.
- Boolean-only and empty filter arrays preserve the backend's aggregate truthiness behavior.
- Composite values use compact JSON with recursively sorted keys and backend-compatible escaping.
- Integer values retain their JSON spelling.
- Floating-point matching falls back to the server because JVM formatting is not identical to `serde_json` for every finite double.
- Exact comparisons use full Unicode lowercase, including final sigma and dotted `İ`.
- Contains, prefix, and suffix comparisons retain ASCII-only lowercase.

The proposed backend cleanup in PostHog/posthog#90694 is a separate draft. This PR intentionally follows the released backend until that proposal is accepted.

The patch changes only `posthog-server` and includes a patch changeset.

Red-green regression evidence:

- Boolean coercion initially failed for `false` versus `"banana"`; the focused evaluator suite now passes.
- Composite stringification initially failed for `"[1,2]"` versus `[1, 2]`; the focused evaluator suite now passes.
- Review follow-ups reproduced null-member omission, broad long-s boolean matching, and nested non-finite failures before their fixes.

Validation:

- `./gradlew :posthog-server:test --tests 'com.posthog.server.internal.FlagEvaluatorTest' --no-daemon`
- `make checkFormat`
- `./gradlew :posthog-server:animalsnifferMain :posthog-server:animalsnifferTest --no-daemon`
- `$HOME/.pi/agent/skills/autoreview/scripts/autoreview --mode local`

The full `:posthog-server:test` suite reached 512 passing tests. Five unrelated Mockito inline-agent tests fail on the local JDK. Required CI remains the authoritative full-suite check.

Autoreview findings about operand direction and whole-array matching were rejected after verification: `propertyValue` is the filter operand, and the focused 100-test evaluator class passes the cited vectors.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi implemented and reviewed the change under human direction using the address-pr-comments, check-pr, autoreview, and karpathy-guidelines skills. The released Rust matcher was the source of truth; the separate backend draft was not adopted.

### DIFF
--- a/.changeset/calm-flags-fold.md
+++ b/.changeset/calm-flags-fold.md
@@ -2,4 +2,4 @@
 "posthog-server": patch
 ---
 
-Match local feature flag string operators using the same ASCII and Unicode lowercase rules as the flags service.
+Match local feature flag string operators using the flags service's boolean coercion, JSON stringification, and case rules.
--- a/.changeset/calm-flags-fold.md
+++ b/.changeset/calm-flags-fold.md
@@ -1,0 +1,5 @@
+---
+"posthog-server": patch
+---
+
+Match local feature flag string operators using the same ASCII and Unicode lowercase rules as the flags service.
--- a/posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
+++ b/posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
@@ -10,7 +10,6 @@ import com.posthog.internal.PropertyOperator
 import com.posthog.internal.PropertyType
 import com.posthog.internal.PropertyValue
 import java.security.MessageDigest
-import java.text.Normalizer
 import java.time.Instant
 import java.time.ZoneId
 import java.time.ZonedDateTime
@@ -29,7 +28,6 @@ internal class FlagEvaluator(
     companion object {
         private const val LONG_SCALE = 0xFFFFFFFFFFFFFFF.toDouble()
         private val NONE_VALUES_ALLOWED_OPERATORS = setOf(PropertyOperator.IS_NOT)
-        private val REGEX_COMBINING_MARKS = "\\p{M}+".toRegex()
         private val REGEX_RELATIVE_DATE = "^-?([0-9]+)([hdwmy])$".toRegex()
         private val REGEX_SEMVER =
             Regex("""^(\d+)(?:\.(\d+))?(?:\.(\d+))?(?:\.\d+)*(?:[-+].*)?$""")
@@ -40,10 +38,7 @@ internal class FlagEvaluator(
             DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ssXXX")
         private val DATE_FORMATTER_NO_TZ = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss")
 
-        private fun casefold(input: String): String {
-            val normalized = Normalizer.normalize(input, Normalizer.Form.NFD)
-            return REGEX_COMBINING_MARKS.replace(normalized, "").uppercase().lowercase()
-        }
+        private fun unicodeLowercase(input: String): String = input.lowercase()
 
         private fun asciiCasefold(input: String): String =
             buildString(input.length) {
@@ -244,23 +239,19 @@ internal class FlagEvaluator(
         propertyValue: Any?,
         overrideValue: Any?,
     ): Boolean {
-        // Lowercase to uppercase to normalize locale (e.g., Turkish i, German ß)
-        // String.equals apparently does this when ignoreCase=true, but it doesn't seem to work.
-        // https://kotlinlang.org/api/core/1.3/kotlin-stdlib/kotlin.text/equals.html
-        val expectedValue = overrideValue?.let { casefold(it.toString()) }
-        return when {
-            propertyValue is List<*> -> {
-                propertyValue.any { v ->
-                    v == expectedValue || (v != null && casefold(v.toString()) == expectedValue)
-                }
-            }
+        if (overrideValue == null) {
+            return if (propertyValue is List<*>) propertyValue.any { it == null } else propertyValue == null
+        }
 
-            else ->
-                propertyValue == expectedValue || (
-                    propertyValue != null && casefold(
-                        propertyValue.toString(),
-                    ) == expectedValue
-                )
+        val expectedValue = unicodeLowercase(overrideValue.toString())
+        return when (propertyValue) {
+            is List<*> ->
+                propertyValue.any { value ->
+                    value != null && unicodeLowercase(value.toString()) == expectedValue
+                }
+
+            null -> false
+            else -> unicodeLowercase(propertyValue.toString()) == expectedValue
         }
     }
 
@@ -270,7 +261,7 @@ internal class FlagEvaluator(
         ignoreCase: Boolean,
     ): Boolean {
         if (ignoreCase) {
-            return casefold(haystack).contains(casefold(needle), ignoreCase = true)
+            return asciiCasefold(haystack).contains(asciiCasefold(needle))
         }
         return haystack.contains(needle)
     }
--- a/posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
+++ b/posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
@@ -1,5 +1,7 @@
 package com.posthog.server.internal
 
+import com.google.gson.GsonBuilder
+import com.google.gson.JsonElement
 import com.posthog.PostHogConfig
 import com.posthog.internal.FlagConditionGroup
 import com.posthog.internal.FlagDefinition
@@ -27,7 +29,8 @@ internal class FlagEvaluator(
 ) {
     companion object {
         private const val LONG_SCALE = 0xFFFFFFFFFFFFFFF.toDouble()
-        private val NONE_VALUES_ALLOWED_OPERATORS = setOf(PropertyOperator.IS_NOT)
+        private val NONE_VALUES_ALLOWED_OPERATORS = setOf(PropertyOperator.EXACT, PropertyOperator.IS_NOT)
+        private val JSON = GsonBuilder().disableHtmlEscaping().serializeNulls().create()
         private val REGEX_RELATIVE_DATE = "^-?([0-9]+)([hdwmy])$".toRegex()
         private val REGEX_SEMVER =
             Regex("""^(\d+)(?:\.(\d+))?(?:\.(\d+))?(?:\.\d+)*(?:[-+].*)?$""")
@@ -153,40 +156,38 @@ internal class FlagEvaluator(
 
             PropertyOperator.ICONTAINS ->
                 stringContains(
-                    overrideValue.toString(),
-                    propertyValue.toString(),
-                    ignoreCase = true,
+                    valueToString(overrideValue),
+                    valueToString(propertyValue),
                 )
 
             PropertyOperator.NOT_ICONTAINS ->
                 !stringContains(
-                    overrideValue.toString(),
-                    propertyValue.toString(),
-                    ignoreCase = true,
+                    valueToString(overrideValue),
+                    valueToString(propertyValue),
                 )
 
             PropertyOperator.STARTS_WITH ->
                 stringStartsWith(
-                    overrideValue.toString(),
-                    propertyValue.toString(),
+                    valueToString(overrideValue),
+                    valueToString(propertyValue),
                 )
 
             PropertyOperator.NOT_STARTS_WITH ->
                 !stringStartsWith(
-                    overrideValue.toString(),
-                    propertyValue.toString(),
+                    valueToString(overrideValue),
+                    valueToString(propertyValue),
                 )
 
             PropertyOperator.ENDS_WITH ->
                 stringEndsWith(
-                    overrideValue.toString(),
-                    propertyValue.toString(),
+                    valueToString(overrideValue),
+                    valueToString(propertyValue),
                 )
 
             PropertyOperator.NOT_ENDS_WITH ->
                 !stringEndsWith(
-                    overrideValue.toString(),
-                    propertyValue.toString(),
+                    valueToString(overrideValue),
+                    valueToString(propertyValue),
                 )
 
             PropertyOperator.REGEX ->
@@ -239,32 +240,130 @@ internal class FlagEvaluator(
         propertyValue: Any?,
         overrideValue: Any?,
     ): Boolean {
-        if (overrideValue == null) {
-            return if (propertyValue is List<*>) propertyValue.any { it == null } else propertyValue == null
+        if (isTruthyOrFalsyPropertyValue(propertyValue)) {
+            return isTruthyPropertyValue(propertyValue) == isTruthyPropertyValue(overrideValue)
         }
 
-        val expectedValue = unicodeLowercase(overrideValue.toString())
+        val expectedValue = unicodeLowercase(valueToString(overrideValue))
         return when (propertyValue) {
             is List<*> ->
                 propertyValue.any { value ->
-                    value != null && unicodeLowercase(value.toString()) == expectedValue
+                    unicodeLowercase(valueToString(value)) == expectedValue
                 }
 
-            null -> false
-            else -> unicodeLowercase(propertyValue.toString()) == expectedValue
+            else -> unicodeLowercase(valueToString(propertyValue)) == expectedValue
         }
     }
+
+    private fun isTruthyOrFalsyPropertyValue(value: Any?): Boolean =
+        when (value) {
+            is Boolean -> true
+            is String -> unicodeLowercase(value).let { it == "true" || it == "false" }
+            is List<*> -> value.all(::isTruthyOrFalsyPropertyValue)
+            else -> false
+        }
+
+    private fun isTruthyPropertyValue(value: Any?): Boolean =
+        when (value) {
+            is Boolean -> value
+            is String -> unicodeLowercase(value) == "true"
+            is List<*> -> value.all(::isTruthyPropertyValue)
+            else -> false
+        }
+
+    private fun valueToString(value: Any?): String {
+        if (value is String) {
+            return value
+        }
+        if (value is Number) {
+            return numberToString(value)
+        }
+        val jsonValue =
+            try {
+                JSON.toJsonTree(value)
+            } catch (error: IllegalArgumentException) {
+                throw InconclusiveMatchException("Cannot stringify JSON value locally: ${error.message}")
+            }
+        return canonicalJson(jsonValue)
+    }
+
+    private fun canonicalJson(value: JsonElement): String =
+        when {
+            value.isJsonNull -> "null"
+            value.isJsonArray ->
+                value.asJsonArray.joinToString(separator = ",", prefix = "[", postfix = "]") { canonicalJson(it) }
+            value.isJsonObject ->
+                value.asJsonObject
+                    .entrySet()
+                    .sortedWith { left, right -> compareJsonKeys(left.key, right.key) }
+                    .joinToString(separator = ",", prefix = "{", postfix = "}") { (key, child) ->
+                        "${quoteJsonString(key)}:${canonicalJson(child)}"
+                    }
+            value.asJsonPrimitive.isString -> quoteJsonString(value.asString)
+            value.asJsonPrimitive.isBoolean -> value.asBoolean.toString()
+            else -> numberToString(value.asNumber)
+        }
+
+    private fun numberToString(value: Number): String {
+        val rawValue = value.toString()
+        if (rawValue.matches(Regex("-?\\d+"))) {
+            return rawValue
+        }
+        throw InconclusiveMatchException("Cannot stringify floating-point JSON number locally: $rawValue")
+    }
+
+    private fun compareJsonKeys(
+        left: String,
+        right: String,
+    ): Int {
+        val leftBytes = left.toByteArray(Charsets.UTF_8)
+        val rightBytes = right.toByteArray(Charsets.UTF_8)
+        for (index in 0 until minOf(leftBytes.size, rightBytes.size)) {
+            val comparison = (leftBytes[index].toInt() and 0xFF) - (rightBytes[index].toInt() and 0xFF)
+            if (comparison != 0) {
+                return comparison
+            }
+        }
+        return leftBytes.size - rightBytes.size
+    }
+
+    private fun quoteJsonString(value: String): String =
+        buildString(value.length + 2) {
+            append('"')
+            value.forEachIndexed { index, character ->
+                when (character) {
+                    '"' -> append("\\\"")
+                    '\\' -> append("\\\\")
+                    '\u0008' -> append("\\b")
+                    '\u000C' -> append("\\f")
+                    '\n' -> append("\\n")
+                    '\r' -> append("\\r")
+                    '\t' -> append("\\t")
+                    in '\u0000'..'\u001F' -> append("\\u%04x".format(character.code))
+                    in '\uD800'..'\uDBFF' -> {
+                        val next = value.getOrNull(index + 1)
+                        if (next == null || next !in '\uDC00'..'\uDFFF') {
+                            throw InconclusiveMatchException("Cannot stringify an unpaired Unicode surrogate")
+                        }
+                        append(character)
+                    }
+                    in '\uDC00'..'\uDFFF' -> {
+                        val previous = value.getOrNull(index - 1)
+                        if (previous == null || previous !in '\uD800'..'\uDBFF') {
+                            throw InconclusiveMatchException("Cannot stringify an unpaired Unicode surrogate")
+                        }
+                        append(character)
+                    }
+                    else -> append(character)
+                }
+            }
+            append('"')
+        }
 
     private fun stringContains(
         haystack: String,
         needle: String,
-        ignoreCase: Boolean,
-    ): Boolean {
-        if (ignoreCase) {
-            return asciiCasefold(haystack).contains(asciiCasefold(needle))
-        }
-        return haystack.contains(needle)
-    }
+    ): Boolean = asciiCasefold(haystack).contains(asciiCasefold(needle))
 
     private fun stringStartsWith(
         value: String,
--- a/posthog-server/src/test/java/com/posthog/server/internal/FlagEvaluatorTest.kt
+++ b/posthog-server/src/test/java/com/posthog/server/internal/FlagEvaluatorTest.kt
@@ -64,45 +64,39 @@ internal class FlagEvaluatorTest {
     }
 
     @Test
-    internal fun testMatchPropertyExactUnicodeNormalization() {
-        // Test German ß (eszett) - should match "ss" after casefold
-        val propertyStrasse =
-            FlagProperty(
-                key = "location",
-                propertyValue = "Straße",
-                propertyOperator = PropertyOperator.EXACT,
-                type = PropertyType.PERSON,
-                negation = false,
-                dependencyChain = null,
+    internal fun testMatchPropertyExactUsesUnicodeLowercase() {
+        fun match(
+            operator: PropertyOperator,
+            filterValue: String,
+            propertyValue: String,
+        ): Boolean {
+            val property =
+                FlagProperty(
+                    key = "value",
+                    propertyValue = filterValue,
+                    propertyOperator = operator,
+                    type = PropertyType.PERSON,
+                    negation = false,
+                    dependencyChain = null,
+                )
+            return evaluator.matchProperty(property, mapOf("value" to propertyValue))
+        }
+
+        val comparisons =
+            listOf(
+                Triple("Ä", "ä", true),
+                Triple("É", "e", false),
+                Triple("ß", "ss", false),
+                Triple("Σ", "ς", false),
             )
-
-        // Should match lowercase ß
-        assertTrue(evaluator.matchProperty(propertyStrasse, mapOf("location" to "straße")))
-
-        // Should match "ss" (casefold normalization)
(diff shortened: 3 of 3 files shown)


posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -270,7 +261,7 @@ internal class FlagEvaluator(
         ignoreCase: Boolean,
     ): Boolean {
         if (ignoreCase) {
-            return casefold(haystack).contains(casefold(needle), ignoreCase = true)
+            return asciiCasefold(haystack).contains(asciiCasefold(needle))
```
ioannisj: [suggestion] Worth calling out the node divergence in the PR body?

This matches the Rust matcher (`property_matching.rs:251-254`, "we only care about ASCII case insensitivity"), so local eval now agrees with `/flags`, which is clearly the right direction.

But posthog-node still does a full Unicode `toLowerCase()` on both sides of `icontains` (`packages/node/src/extensions/feature-flags/feature-flags.ts:1119`), so `city = "MÜNCHEN"` against `icontains "münchen"` is `false` here and `true` there. Two server SDKs, same data, different flag.

Could we open a follow-up so node converges on the service rule? Otherwise posthog-server ends up the odd one out for being the correct one.

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -270,7 +261,7 @@ internal class FlagEvaluator(
         ignoreCase: Boolean,
```
ioannisj: [nit] looks like both call sites oass true here so`ignoreCase` is dead. Worth dropping the param and collapsing this to the `asciiCasefold` one-liner?

ioannisj (approved):

arnohillen:
`computeExactMatch` (`posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt:238`) still skips the service's boolean-coercion gate (`is_truthy_or_falsy_property_value`, `rust/feature-flags/src/properties/property_matching.rs:212-220`), so `exact` with filter `["false"]` and property `"pro"` returns `false` locally and `true` on `/flags` (same for filter `false` vs `"banana"` and filter `[]` vs `true`), while filter `["true","false"]` vs `"true"` returns `true` locally and `false` on `/flags`; pre-existing (identical results on `main`) and `posthog-python`'s `compute_exact_match` has the same gap, so follow-up rather than a blocker here.

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
marandaneto: Fixed in 41fa02a8. Static inspection confirmed both `stringContains` call sites always enabled ASCII folding, so the dead `ignoreCase` parameter and unreachable branch are removed. `FlagEvaluatorTest` and formatting checks pass.

marandaneto:
Addressed the released-backend parity findings in 41fa02a8. A focused red test reproduced `false` versus `"banana"` returning `false` locally before the fix; the evaluator now applies the backend boolean-array gate first, including empty-array truthiness, and the same suite passes. Composite values now use compact recursively sorted JSON, while floating-point values return inconclusive and fall back to `/flags` because JVM formatting cannot reproduce every `serde_json` spelling. Strong final-sigma and dotted-I vectors pass without changing Kotlin lowercase. PostHog/posthog#90694 remains the separate proposed backend cleanup.

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -244,36 +240,130 @@ internal class FlagEvaluator(
...
+            value.asJsonPrimitive.isBoolean -> value.asBoolean.toString()
+            else -> numberToString(value.asNumber)
+        }
+
+    private fun numberToString(value: Number): String {
+        val rawValue = value.toString()
+        if (rawValue.matches(Regex("-?\\d+"))) {
+            return rawValue
+        }
+        throw InconclusiveMatchException("Cannot stringify floating-point JSON number locally: $rawValue")
```
haacked: `suggestion`: A person property holding a plain decimal like `9.99` stops evaluating locally and starts making a `/flags` request on every call. `numberToString` accepts only renderings matching `-?\d+` and throws `InconclusiveMatchException` for everything else, which covers every `Double` and `Float`. It applies to `exact`, `is_not`, `icontains`, `starts_with`, `ends_with` and their negations, on either operand.

For many of those values the local answer was already right. `Double.toString` on JDK 17 renders `323.0`, `9.99`, `0.5` and `0.001` as plain decimals identical to what serde_json emits, so `exact "9.99"` against a property of `9.99` matched locally before this change and matched on the service too. Now it goes remote.

That costs more than latency. `resolveFeatureFlag` ([PostHogFeatureFlags.kt:209](https://github.com/PostHog/posthog-android/blob/41fa02a88bdf6894b87a8fe5c377ab5bb089fc4c/posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt#L209)) falls through to `getFeatureFlagsFromRemote`, which returns null when the request fails, so the caller gets their default value in a case where the previous release returned the correct flag. It also puts a billed `/flags` call in the hot path for customers who turned on local evaluation to avoid one.

Narrowing the fallback to the renderings that genuinely differ keeps those cases local:

```kotlin
private fun numberToString(value: Number): String {
    val rawValue = value.toString()
    if (rawValue.matches(REGEX_INTEGER)) {
        return rawValue
    }
    if (rawValue.matches(REGEX_PLAIN_DECIMAL) && isShortestRoundTrip(rawValue, value.toDouble())) {
        return rawValue
    }
    throw InconclusiveMatchException("Cannot stringify floating-point JSON number locally: $rawValue")
}

private fun isShortestRoundTrip(rendered: String, value: Double): Boolean {
    if (rendered.toDouble() != value) return false
    val shorter = rendered.dropLast(1)
    return shorter.endsWith(".") || shorter.toDouble() != value
}
```

with `REGEX_PLAIN_DECIMAL = Regex("-?\\d+\\.\\d+")` next to the integer pattern. One caveat I cannot close: JDK 17's `Double.toString` guarantees round-trip but not shortest digits, which only landed in JDK 19, so the trailing-digit check narrows the risk rather than removing it. If you would rather not take that on, the fallback is defensible as written and the ask moves to the changeset instead.

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -28,8 +29,8 @@ internal class FlagEvaluator(
 ) {
     companion object {
         private const val LONG_SCALE = 0xFFFFFFFFFFFFFFF.toDouble()
-        private val NONE_VALUES_ALLOWED_OPERATORS = setOf(PropertyOperator.IS_NOT)
-        private val REGEX_COMBINING_MARKS = "\\p{M}+".toRegex()
+        private val NONE_VALUES_ALLOWED_OPERATORS = setOf(PropertyOperator.EXACT, PropertyOperator.IS_NOT)
```
haacked: `suggestion`: A person property that is present but explicitly null still gets a different answer locally than it does on `/flags` for the six case-insensitive string operators. The service stringifies a JSON null to the literal `"null"` and then runs the operator, so `not_icontains "enterprise"` against a null `plan` property returns true there. Locally the null gate at [FlagEvaluator.kt:147](https://github.com/PostHog/posthog-android/blob/41fa02a88bdf6894b87a8fe5c377ab5bb089fc4c/posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt#L147) returns false for that operator, the condition fails, and the flag comes back off. A filter of `"ul"` under `icontains` is the shortest example: false locally, true on the service.

Nulls do reach that line. `localPersonProperties` ([PostHogFeatureFlags.kt:383](https://github.com/PostHog/posthog-android/blob/41fa02a88bdf6894b87a8fe5c377ab5bb089fc4c/posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt#L383)) copies the caller's map as is, so `mapOf("plan" to user.plan)` with a null plan arrives with the key present and the value null.

This PR already closes the gap for `exact` and `is_not` by adding `EXACT` to this set. The six string operators are covered by the PR title too, and they still short-circuit; widening the set reaches them as well. The divergence predates this PR for those operators, so whether it belongs in this change is your call:

```suggestion
        private val NONE_VALUES_ALLOWED_OPERATORS =
            setOf(
                PropertyOperator.EXACT,
                PropertyOperator.IS_NOT,
                PropertyOperator.ICONTAINS,
                PropertyOperator.NOT_ICONTAINS,
                PropertyOperator.STARTS_WITH,
                PropertyOperator.NOT_STARTS_WITH,
                PropertyOperator.ENDS_WITH,
                PropertyOperator.NOT_ENDS_WITH,
            )
```

Separately, and pushing the other way: now that `EXACT` is in this set, a filter whose own value is null reaches `computeExactMatch(null, null)`, both sides stringify to the literal `null`, and the match returns true. The service returns false for a missing filter value before it dispatches on the operator at all. I could not tell from here whether `/api/feature_flag/local_evaluation` ever emits `"value": null` for an `exact` filter, so this may be unreachable; a `propertyValue == null` guard in the `EXACT`/`IS_NOT` branch would cover it if it is not.

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -244,36 +240,130 @@ internal class FlagEvaluator(
...
+                        "${quoteJsonString(key)}:${canonicalJson(child)}"
+                    }
+            value.asJsonPrimitive.isString -> quoteJsonString(value.asString)
+            value.asJsonPrimitive.isBoolean -> value.asBoolean.toString()
+            else -> numberToString(value.asNumber)
+        }
+
+    private fun numberToString(value: Number): String {
+        val rawValue = value.toString()
+        if (rawValue.matches(Regex("-?\\d+"))) {
```
haacked: `suggestion`: `numberToString` builds `Regex("-?\\d+")` inside the function body, so the JVM runs `Pattern.compile` and allocates a `Matcher` on every call instead of once. The file already keeps its fixed patterns in the companion object next to `REGEX_RELATIVE_DATE` and `REGEX_SEMVER` ([FlagEvaluator.kt:34](https://github.com/PostHog/posthog-android/blob/41fa02a88bdf6894b87a8fe5c377ab5bb089fc4c/posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt#L34)). Add this one alongside them:

```kotlin
private val REGEX_INTEGER = Regex("-?\\d+")
```

then:

```suggestion
        if (rawValue.matches(REGEX_INTEGER)) {
```

.changeset/calm-flags-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-server": patch
+---
+
+Match local feature flag string operators using the flags service's boolean coercion, JSON stringification, and case rules.
```
haacked: `suggestion`: The changeset does not tell an upgrading user that evaluation results will change for existing flags. Someone reading the changelog will not see that their flags stopped matching locally and now hit the server.

This repo already has precedent for documenting behavior changes: [CHANGELOG.md](https://github.com/PostHog/posthog-android/blob/41fa02a88bdf6894b87a8fe5c377ab5bb089fc4c/posthog-server/CHANGELOG.md) shows 2.15.0 taking a minor bump for a local-evaluation alignment. The body should be updated to:

```suggestion
Align local feature flag matching with the flags service. Evaluation results change for existing flags: `exact` and `is_not` no longer ignore accents or fold ß to ss, so `München` stops matching a filter of `munchen`; `icontains` now folds ASCII case only, so `İstanbul` stops matching a filter of `Istanbul`; a boolean-like filter such as `["false"]` is compared by truthiness rather than string equality; and float-valued properties fall back to `/flags` instead of evaluating locally.
```

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -244,36 +240,130 @@ internal class FlagEvaluator(
...
+
+    private fun isTruthyPropertyValue(value: Any?): Boolean =
+        when (value) {
+            is Boolean -> value
+            is String -> unicodeLowercase(value) == "true"
+            is List<*> -> value.all(::isTruthyPropertyValue)
+            else -> false
+        }
+
+    private fun valueToString(value: Any?): String {
```
haacked: `suggestion`: The new `valueToString` path handles two identical situations in opposite ways. When the JVM cannot reproduce serde_json's spelling of a float, `numberToString` throws and the flag falls back to `/flags`. When the value is a type the backend has no JSON representation for at all, `JSON.toJsonTree` invents one and matches against it locally. Compiled against the pinned gson 2.10.1 ([PosthogBuildConfig.kt:65](https://github.com/PostHog/posthog-android/blob/41fa02a88bdf6894b87a8fe5c377ab5bb089fc4c/buildSrc/src/main/java/PosthogBuildConfig.kt#L65)), `toJsonTree(Date(0))` returns `"Dec 31, 1969, 4:00:00 PM"`, which varies with the machine's default locale and timezone. The service only ever sees a `serde_json::Value`, so it can never produce that string, and two servers in different regions can disagree with each other.

Gson is only being used to turn a Kotlin value into a tree that `canonicalJson` walks straight back down. Dispatching on the Kotlin types directly does the same job, drops the third `Gson` instance in the codebase, and routes anything outside the JSON type set to the server the way floats already go:

```kotlin
private fun valueToString(value: Any?): String = if (value is String) value else canonicalJson(value)

private fun canonicalJson(value: Any?): String =
    when (value) {
        null -> "null"
        is String -> quoteJsonString(value)
        is Boolean -> value.toString()
        is Number -> numberToString(value)
        is List<*> -> value.joinToString(separator = ",", prefix = "[", postfix = "]") { canonicalJson(it) }
        is Map<*, *> ->
            value.entries
                .sortedWith { left, right -> compareJsonKeys(left.key.toString(), right.key.toString()) }
                .joinToString(separator = ",", prefix = "{", postfix = "}") { (key, child) ->
                    "${quoteJsonString(key.toString())}:${canonicalJson(child)}"
                }
        else -> throw InconclusiveMatchException("Cannot stringify ${value.javaClass.name} locally")
    }
```

The cost worth weighing: types Gson serializes for free today, such as `IntArray`, `Set`, or an arbitrary POJO, would fall back to `/flags` instead of evaluating locally. That surface is narrow, since `FlagProperty.propertyValue` arrives through Gson deserialization and is always `List`, `LinkedTreeMap`, `String`, `Number`, `Boolean`, or null; only caller-supplied `personProperties` can reach the `else` branch. Every case in `testMatchPropertyExactUsesBackendJsonStringification` and `testMatchPropertyFloatingPointStringificationFallsBackToServer` produces the same result under this version, including `{"a":null}` and the `[NaN]` inconclusive case. Taking this also removes the `JSON` field and makes the separate note on line 33 moot.

posthog-server/src/test/java/com/posthog/server/internal/FlagEvaluatorTest.kt
```diff
@@ -64,58 +64,213 @@ internal class FlagEvaluatorTest {
...
-                key = "star",
-                propertyValue = "ſun",
+                key = "value",
+                propertyValue = false,
                 propertyOperator = PropertyOperator.EXACT,
                 type = PropertyType.PERSON,
                 negation = false,
                 dependencyChain = null,
             )
+        assertTrue(evaluator.matchProperty(falseFilter, mapOf("value" to null)))
```
haacked: `suggestion`: Nothing covers `EXACT` against a property that is present but explicitly null when the filter value is not boolean-like. Adding `EXACT` to `NONE_VALUES_ALLOWED_OPERATORS` means `mapOf("value" to null)` no longer short-circuits to false and now reaches `computeExactMatch`, where `valueToString(null)` renders the null as the four-character string `null`. A filter value of `"null"` therefore matches a person whose property is null, which it did not before this PR.

The one null-property assertion in the suite is this line, and it uses a `false` filter, so it returns through the boolean gate and never reaches the stringification. Extending `testMatchPropertyExactUsesBackendJsonStringification` with a null property against a plain string filter and against a `"null"` filter would write down whichever result you intend.

posthog-server/src/test/java/com/posthog/server/internal/FlagEvaluatorTest.kt
```diff
@@ -64,58 +64,213 @@ internal class FlagEvaluatorTest {
...
 
-        // Should match with casefold normalization
-        assertTrue(evaluator.matchProperty(property, mapOf("location" to "strasse")))
-        assertTrue(evaluator.matchProperty(property, mapOf("location" to "munchen")))
+    @Test
+    internal fun testMatchPropertyStringOperatorsUseBackendJsonStringification() {
+        val comparisons =
+            listOf(
+                Triple(PropertyOperator.ICONTAINS, "\"a\":2", linkedMapOf("b" to 1, "a" to 2)),
+                Triple(PropertyOperator.STARTS_WITH, "922", Long.MAX_VALUE),
```
haacked: `suggestion`: Two of the three rows in `testMatchPropertyStringOperatorsUseBackendJsonStringification` pass under the old `.toString()` behavior as well, so only the ICONTAINS row verifies what the test name claims. `Long.MAX_VALUE` has the same spelling either way, and `listOf(1, 2).toString()` is `[1, 2]`, which ends with `2]` exactly like the new `[1,2]` does. If STARTS_WITH or ENDS_WITH went back to `overrideValue.toString()`, this test stays green.

These two rows discriminate, since the old spellings are `{b=1, a=2}` and `[1, 2]`:

```kotlin
Triple(PropertyOperator.STARTS_WITH, "{\"a\":2", linkedMapOf("b" to 1, "a" to 2)),
Triple(PropertyOperator.ENDS_WITH, ",2]", listOf(1, 2)),
```

The Long spelling is already covered at [FlagEvaluatorTest.kt:184](https://github.com/PostHog/posthog-android/blob/41fa02a88bdf6894b87a8fe5c377ab5bb089fc4c/posthog-server/src/test/java/com/posthog/server/internal/FlagEvaluatorTest.kt#L184).

posthog-server/src/test/java/com/posthog/server/internal/FlagEvaluatorTest.kt
```diff
@@ -64,58 +64,213 @@ internal class FlagEvaluatorTest {
...
-                propertyValue = listOf("Straße", "München"),
+                key = "value",
+                propertyValue = "[NaN]",
                 propertyOperator = PropertyOperator.EXACT,
                 type = PropertyType.PERSON,
                 negation = false,
                 dependencyChain = null,
             )
+        try {
+            evaluator.matchProperty(nestedNonFinite, mapOf("value" to listOf(Double.NaN)))
```
haacked: `suggestion`: Nothing verifies that a finite non-integer double nested inside a list or map falls back to the server. `testMatchPropertyFloatingPointStringificationFallsBackToServer` feeds top-level doubles, which return early from `valueToString` and never reach `canonicalJson`.

This `[NaN]` case looks like it covers the nested path but does not. Gson's double adapter rejects NaN with `IllegalArgumentException` before the tree is built, so this vector exercises the catch around `toJsonTree` instead. Compiled against the pinned gson 2.10.1, `toJsonTree(listOf(NaN))` throws while `toJsonTree(listOf(1.5))` builds a tree and yields `1.5`.

That leaves `numberToString(value.asNumber)` in the `canonicalJson` fall-through uncovered, which is the branch that keeps a person property like `listOf(1.5)` or `mapOf("price" to 9.99)` from being matched locally with JVM double formatting. A `listOf(1.5)` vector next to this one reaches it.

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -244,36 +240,130 @@ internal class FlagEvaluator(
...
+                    '\u0008' -> append("\\b")
+                    '\u000C' -> append("\\f")
+                    '\n' -> append("\\n")
+                    '\r' -> append("\\r")
+                    '\t' -> append("\\t")
+                    in '\u0000'..'\u001F' -> append("\\u%04x".format(character.code))
+                    in '\uD800'..'\uDBFF' -> {
+                        val next = value.getOrNull(index + 1)
+                        if (next == null || next !in '\uDC00'..'\uDFFF') {
+                            throw InconclusiveMatchException("Cannot stringify an unpaired Unicode surrogate")
```
haacked: `suggestion`: The unpaired-surrogate guards here and at [FlagEvaluator.kt:353](https://github.com/PostHog/posthog-android/blob/41fa02a88bdf6894b87a8fe5c377ab5bb089fc4c/posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt#L353) have no coverage, so the server fallback they trigger could stop happening without a test failing. The only surrogate vector in the suite is the well-formed pair used as a map key at [FlagEvaluatorTest.kt:172](https://github.com/PostHog/posthog-android/blob/41fa02a88bdf6894b87a8fe5c377ab5bb089fc4c/posthog-server/src/test/java/com/posthog/server/internal/FlagEvaluatorTest.kt#L172), which takes the append branches rather than the throw branches.

A lone `"\uD800"` inside a composite property value reaches them. A top-level string will not, since `valueToString` returns strings verbatim without ever calling `quoteJsonString`.

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -244,36 +240,130 @@ internal class FlagEvaluator(
...
+
+    private fun numberToString(value: Number): String {
+        val rawValue = value.toString()
+        if (rawValue.matches(Regex("-?\\d+"))) {
+            return rawValue
+        }
+        throw InconclusiveMatchException("Cannot stringify floating-point JSON number locally: $rawValue")
+    }
+
+    private fun compareJsonKeys(
```
haacked: `suggestion`: `compareJsonKeys` reads like 13 lines doing what `sortedBy { it.key }` does, and nothing on the function says otherwise. The two orderings differ only above U+FFFF: Kotlin compares strings by UTF-16 code unit, which sorts a surrogate pair ahead of U+E000, while serde_json orders object keys by UTF-8 bytes, which puts U+E000 first. The `"\uE000"` against `"\uD800\uDC00"` row in `testMatchPropertyExactUsesBackendJsonStringification` is the only case that fails if someone swaps in the natural ordering, and that row does not explain itself either.

One line above the function carries it:

```suggestion
    // serde_json orders object keys by UTF-8 bytes, which differs from Kotlin's UTF-16 order above U+FFFF.
    private fun compareJsonKeys(
```

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -244,36 +240,130 @@ internal class FlagEvaluator(
...
+                when (character) {
+                    '"' -> append("\\\"")
+                    '\\' -> append("\\\\")
+                    '\u0008' -> append("\\b")
+                    '\u000C' -> append("\\f")
+                    '\n' -> append("\\n")
+                    '\r' -> append("\\r")
+                    '\t' -> append("\\t")
+                    in '\u0000'..'\u001F' -> append("\\u%04x".format(character.code))
+                    in '\uD800'..'\uDBFF' -> {
```
haacked: `suggestion`: The two surrogate branches spell out `'\uD800'..'\uDBFF'` and `'\uDC00'..'\uDFFF'` four times between them and repeat the same throw, so a reader has to decode the ranges before seeing that this is a paired-surrogate check. Kotlin has named predicates for both halves. One branch over the whole surrogate block keeps the ranges down to one and leaves the exception string in a single place:

```kotlin
in '\uD800'..'\uDFFF' -> {
    val paired =
        if (character.isHighSurrogate()) {
            value.getOrNull(index + 1)?.isLowSurrogate() == true
        } else {
            value.getOrNull(index - 1)?.isHighSurrogate() == true
        }
    if (!paired) {
        throw InconclusiveMatchException("Cannot stringify an unpaired Unicode surrogate")
    }
    append(character)
}
```

This branch sits after `in '\u0000'..'\u001F'` and covers a range that does not overlap it, so the `when` ordering is unaffected.

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -158,40 +156,38 @@ internal class FlagEvaluator(
...
 
             PropertyOperator.NOT_ENDS_WITH ->
                 !stringEndsWith(
-                    overrideValue.toString(),
-                    propertyValue.toString(),
+                    valueToString(overrideValue),
+                    valueToString(propertyValue),
                 )
 
             PropertyOperator.REGEX ->
```
haacked: `question`: After this change `matchProperty` stringifies operands two different ways in the same `when`. The six string operators go through `valueToString`, while `REGEX` and `NOT_REGEX` still call Kotlin's `.toString()`. A person property of `listOf(1, 2)` becomes `[1,2]` for `ICONTAINS` and `[1, 2]` for `REGEX`, so a pattern anchored on the compact spelling matches on the service and misses locally.

The released Rust matcher runs the pattern against `to_string_representation`, the same function backing the operators you did change.

Was regex deliberately left out of "align local string matching", or did it get missed? If it belongs in scope it is the same shape as the others, with the tradeoff that a float-valued property under `REGEX` would start falling back to the server the way the other six now do:

```kotlin
PropertyOperator.REGEX ->
    matchesRegex(
        valueToString(propertyValue),
        valueToString(overrideValue),
    )
```

posthog-server/src/main/java/com/posthog/server/internal/FlagEvaluator.kt
```diff
@@ -28,8 +29,8 @@ internal class FlagEvaluator(
 ) {
     companion object {
         private const val LONG_SCALE = 0xFFFFFFFFFFFFFFF.toDouble()
-        private val NONE_VALUES_ALLOWED_OPERATORS = setOf(PropertyOperator.IS_NOT)
-        private val REGEX_COMBINING_MARKS = "\\p{M}+".toRegex()
+        private val NONE_VALUES_ALLOWED_OPERATORS = setOf(PropertyOperator.EXACT, PropertyOperator.IS_NOT)
+        private val JSON = GsonBuilder().disableHtmlEscaping().serializeNulls().create()
```
haacked: `nit`: `disableHtmlEscaping()` changes nothing here: the only call on this instance is `toJsonTree`, which builds an element tree and never writes JSON text, so `quoteJsonString` is what actually decides escaping. Keep `serializeNulls()`, which does change the tree and is what makes the `{"a":null}` case pass.

```suggestion
        private val JSON = GsonBuilder().serializeNulls().create()
```

PR #747, opened 2026-09-01
chore(gradle-plugin): deprecate event release mode

ablaszkiewicz:
## Related PRs

Event mode is live today, and it is opt-in. These PRs settle where it stays and where it goes.

**Deprecating the mobile knobs.** The symbol id on these paths is already a content hash, so two releases collide only when they ship a byte-identical artifact. An ordinary release that changes code gets its own symbol set and never collides. The dSYM path also lost release attribution for embedded targets, because one upload covers every target while it creates one release. Review asked for deprecation instead of removal, so every knob stays accepted as a warned no-op.

- PostHog/posthog#92401
- PostHog/posthog-android#747 ← this PR
- PostHog/posthog-ios#791

**Making it the default.**

- PostHog/posthog#91823 (stacked on PostHog/posthog#92401)
- PostHog/posthog-js#4705

**React Native.** Scopes the mode to the Hermes upload and defaults it to event. That is the one path where two releases really do ship the same artifact.

- PostHog/posthog-js#4717

## Problem

- Event mode helps only when two releases ship a byte-identical mapping.
- The map id is already a content hash of the mapping, so a release that changes code gets its own symbol set.
- An ordinary release therefore never collides with an earlier one, and the mode adds nothing.
- Removing the property outright would silently flip a configured build to binding, with nothing in the build output saying so.

## Changes

- A build that sets `posthog.releaseMode` keeps building and gets a deprecation warning. The value is ignored, and the mapping uploads bound to the release the build creates.
- The upload task keeps its public `releaseMode` input as a `@Deprecated` no-op, so a build script that configures the task does not break. Setting it warns at execution.
- The `POSTHOG_RELEASE_MODE` environment variable no longer reaches the upload. The task pins `symbol-set` into the posthog-cli child environment, so an inherited value cannot unbind the mapping on an older CLI.
- The plugin stays silent about the environment variable on purpose. It still steers the sourcemap and hermes uploads of other PostHog tools, so a warning about it would not be actionable.
- `PostHogReleaseMode` and the mode plumbing stay removed. Only the warning path reads the property now.

- `./gradlew :posthog-android-gradle-plugin:test spotlessCheck detekt` passes.
- The warning paths have no test. The module has no Project or Exec harness, and building one for two log lines does not earn its cost.

## 🤖 Agent context

Written with Claude Code (Opus 5); reshaped from removal to deprecation with Claude Code (Fable 5). Skills invoked: `/writing-pr-descriptions`.

This replaces PostHog/posthog-android#739. The reshape answers the review comment asking that upgrading must not break or silently change configured builds. PostHog/posthog#92401 keeps `proguard upload --release-mode` parseable on the CLI side, which covers a build that upgrades the CLI before this plugin.

### DIFF
--- a/posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
+++ b/posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
@@ -45,8 +45,6 @@ internal abstract class PostHogAndroidGradlePlugin
                 ) {}
             buildEventsListenerRegistry.onTaskCompletion(failureTracker)
 
-            val releaseMode = resolvePostHogReleaseMode(project)
-
             project.pluginManager.withPlugin("com.android.application") {
                 val androidComponentsExt =
                     project.extensions.getByType(ApplicationAndroidComponentsExtension::class.java)
@@ -66,7 +64,7 @@ internal abstract class PostHogAndroidGradlePlugin
                     // TODO: skip variants, skip autoUpload, release info, allow failure, debug mode
 
                     val paths = OutputPaths(project, variant.name)
-                    val generateMapIdTask = generateMapIdTask(project, variant, paths, failureTracker, releaseMode)
+                    val generateMapIdTask = generateMapIdTask(project, variant, paths, failureTracker)
                     tasksGeneratingProperties.add(generateMapIdTask)
 
                     variant.apply {
@@ -158,7 +156,6 @@ internal abstract class PostHogAndroidGradlePlugin
             variant: ApplicationVariant,
             paths: OutputPaths,
             failureTracker: Provider<PostHogTaskFailureTracker>,
-            releaseMode: PostHogReleaseMode,
         ): TaskProvider<PostHogGenerateMapIdTask> {
             val generateMapIdTask =
                 PostHogGenerateMapIdTask.register(
@@ -174,7 +171,6 @@ internal abstract class PostHogAndroidGradlePlugin
                     generateMapIdTask = generateMapIdTask,
                     variant = variant,
                     mappingFiles = variant.mappingFileProvider(project),
-                    releaseMode = releaseMode,
                 )
 
             generateMapIdTask.hookWithMinifyTasks(project, variant.name, generateMapIdTask)
@@ -189,7 +185,6 @@ internal abstract class PostHogAndroidGradlePlugin
             generateMapIdTask: Provider<PostHogGenerateMapIdTask>,
             variant: ApplicationVariant,
             mappingFiles: Provider<FileCollection>,
-            releaseMode: PostHogReleaseMode,
         ): TaskProvider<PostHogUploadProguardMappingsTask> {
             val primaryOutput = variant.outputs.firstOrNull()
             val uploadMapIdTask =
@@ -201,7 +196,6 @@ internal abstract class PostHogAndroidGradlePlugin
                     releaseName = variant.applicationId,
                     releaseVersion = primaryOutput?.versionName?.map { it.orEmpty() },
                     build = primaryOutput?.versionCode?.map { it ?: 0 },
-                    releaseMode = releaseMode,
                 )
             return uploadMapIdTask
         }
--- a/posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
+++ b/posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
@@ -38,6 +38,8 @@ internal abstract class PostHogAndroidGradlePlugin
             extension.uploadNativeSymbols.convention(false)
             extension.includeNativeSymbolSources.convention(false)
 
+            warnIfDeprecatedReleaseModeSet(project)
+
             val failureTracker =
                 project.gradle.sharedServices.registerIfAbsent(
                     PostHogTaskFailureTracker.NAME,
--- a/posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadProguardMappingsTask.kt
+++ b/posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadProguardMappingsTask.kt
@@ -53,21 +53,10 @@ public abstract class PostHogUploadProguardMappingsTask : PostHogCliExecTask() {
     @get:Optional
     public abstract val build: Property<Int>
 
-    /** [PostHogReleaseMode.cliValue], as the enum itself is internal to the plugin. */
-    @get:Input
-    @get:Optional
-    public abstract val releaseMode: Property<String>
-
     override fun exec() {
         if (!mappingsFiles.isPresent || mappingsFiles.get().isEmpty) {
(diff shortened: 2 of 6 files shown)


posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadProguardMappingsTask.kt
```diff
@@ -53,21 +53,10 @@ public abstract class PostHogUploadProguardMappingsTask : PostHogCliExecTask() {
...
         if (!mappingsFiles.isPresent || mappingsFiles.get().isEmpty) {
             error("[PostHog] Mapping files are missing!")
         }
-        // posthog-cli reads POSTHOG_RELEASE_MODE itself when --release-mode is absent, and the
-        // Gradle daemon's environment may carry one that contradicts the mode resolved here —
-        // where the gradle property is supposed to win. Pin the resolved mode in the child's
-        // environment so it cannot be overridden by inheritance. A posthog-cli predating the flag
-        // has no such argument and ignores the variable, so this stays safe for old CLIs.
-        releaseMode.orNull?.let { environment(POSTHOG_RELEASE_MODE_ENV, it) }
         super.exec()
```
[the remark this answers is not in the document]
ablaszkiewicz: Valid, and fixed in 16c3f8e5.

I checked both halves. An `Exec` task inherits the daemon environment, and every released posthog-cli binds `POSTHOG_RELEASE_MODE` to `--release-mode` on `proguard upload` (`cli/src/proguard/upload.rs:42` on master). The plugin used to pin the resolved mode into the child environment for exactly this reason, and removing the feature took the pin with it.

The task now sets `POSTHOG_RELEASE_MODE=symbol-set` before it runs the CLI, so an inherited value cannot change what the mapping binds to. That holds across all three CLI generations: one without the flag ignores the variable, one with the flag reads `symbol-set`, and PostHog/posthog#92401 removes the flag and ignores it again.

Not covered by a test. The pin happens inside `exec()`, so observing it means running the process, and this module has no Exec harness. The build, `spotlessCheck` and `detekt` pass.

marandaneto:
if people set those, this will break their builds
instead, deprecate the config and mention its a no-op, and remove the behaviour, but upgrading should not break peoples build
or, bump a major

ablaszkiewicz:
@marandaneto deprecated it instead of getting rid of it as you suggested. May I get a re-review?

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadProguardMappingsTask.kt
```diff
@@ -62,12 +62,21 @@ public abstract class PostHogUploadProguardMappingsTask : PostHogCliExecTask() {
...
+                "[PostHog] releaseMode is deprecated and ignored. The mapping uploads bound to " +
+                    "the release this build creates. Remove the assignment.",
+            )
+        }
+
+        // The mapping always binds to the release this build creates. posthog-cli reads
+        // POSTHOG_RELEASE_MODE for `proguard upload`, and an Exec task inherits the daemon
+        // environment, so a value set for another tool would otherwise leave the mapping
+        // release-independent. Pin it. A posthog-cli that has no such flag ignores the variable.
+        environment("POSTHOG_RELEASE_MODE", "symbol-set")
```
turnipdabeets, line 95: [question] Pinning this is the right call for the inherited-env case. One gap though: `POSTHOG_RELEASE_MODE=event` used to select event mode too, so a CI job that set it there gets its mappings bound to a release again with no warning at all — the gradle property path warns, this one doesn't. Could we warn when the inherited value isn't `symbol-set`, worded so it doesn't ask people to unset it (it still steers sourcemap and hermes)?

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
```diff
@@ -38,15 +38,15 @@ internal abstract class PostHogAndroidGradlePlugin
             extension.uploadNativeSymbols.convention(false)
             extension.includeNativeSymbolSources.convention(false)
 
+            warnIfDeprecatedReleaseModeSet(project)
```
turnipdabeets, line 41: [suggestion] This runs at configuration time, so with the configuration cache it prints on the first build and never again on a reused entry — and Gradle 9 turns that on by default. If we also set the property's value on the upload task's `releaseMode`, the execution-time warning fires every build instead. `NativeSymbolsUploadFunctionalTest` already builds a real app with `--configuration-cache` and asserts on the output, so it's a short test there.

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadProguardMappingsTask.kt
```diff
@@ -62,12 +62,21 @@ public abstract class PostHogUploadProguardMappingsTask : PostHogCliExecTask() {
...
+        val deprecatedReleaseMode = releaseMode
+        if (deprecatedReleaseMode.isPresent) {
+            logger.warn(
+                "[PostHog] releaseMode is deprecated and ignored. The mapping uploads bound to " +
+                    "the release this build creates. Remove the assignment.",
+            )
+        }
+
+        // The mapping always binds to the release this build creates. posthog-cli reads
+        // POSTHOG_RELEASE_MODE for `proguard upload`, and an Exec task inherits the daemon
```
turnipdabeets: [nit] PostHog/posthog#92401 removes `env = "POSTHOG_RELEASE_MODE"` from `proguard upload`, so this stops being true once that ships and the pin only covers CLI 0.12.0 up to that release. Worth naming that window here so we know when it can go.

marandaneto (approved):
approving to unblock but see comments first

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadProguardMappingsTask.kt
ablaszkiewicz: Good idea, added in 27cfc0c. The task now warns when the inherited value is not symbol-set. The message says the mapping binds again and does not ask to remove the variable.

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogAndroidGradlePlugin.kt
ablaszkiewicz: Done in 27cfc0c. The property now feeds the task releaseMode input via gradleProperty, so the execution time warning fires on every executed upload, also on a reused cache entry. Added a functional test that runs the upload twice with the configuration cache and checks the warning on the reused entry.

posthog-android-gradle-plugin/src/main/kotlin/com/posthog/android/PostHogUploadProguardMappingsTask.kt
ablaszkiewicz: Good catch. The comment now names the range, 0.13.0 up to the release that carries PostHog/posthog#92401, and says the pin can go once that range ages out.

PR #749, opened 2026-09-01
fix(android): move setup storage I/O off main thread

### LINKED ISSUE #738: PostHogAndroid.setup() triggers StrictMode DiskReadViolation during app startup
### Version

3.61.0

### Steps to Reproduce

**Steps to Reproduce**

1. Enable StrictMode disk read (or detectAll) detection on the main thread.
2. Initialize PostHog from Application.onCreate() using PostHogAndroid.setup(...).
3. Launch the app.
4. Observe a StrictMode.DiskReadViolation originating from:
      - **context.getDir("app_posthog-disk-queue", Context.MODE_PRIVATE)**
6. The stack trace goes through ContextImpl.getDir() → File.exists() during PostHogAndroid.setup().

### Expected Result

**PostHogAndroid.setup()** should ideally avoid synchronous disk I/O on the main thread during app startup, or defer legacy storage resolution to a background/IO thread.

If synchronous disk access during setup() is expected, it would be helpful to document the recommended initialization pattern for consumers.

### Actual Result

**PostHogAndroid.setup()** performs a synchronous filesystem access on the main thread while resolving the legacy queue directory.

This triggers a StrictMode.DiskReadViolation:
```
StrictMode policy violation; ~duration=141 ms: android.os.strictmode.DiskReadViolation
    at java.io.File.exists(File.java:829)
    at android.app.ContextImpl.getDataDir(ContextImpl.java:3544)
    at android.app.ContextImpl.getDir(ContextImpl.java:3564)
    at android.content.ContextWrapper.getDir(ContextWrapper.java:353)
    at com.posthog.android.PostHogAndroid$Companion.setAndroidConfig(PostHogAndroid.kt:117)
    at com.posthog.android.PostHogAndroid$Companion.setup(PostHogAndroid.kt:55)
```
The relevant SDK line is:
```
val legacyPath =
    context.getDir("app_posthog-disk-queue", Context.MODE_PRIVATE)
```

dustinbyrne:

`PostHogAndroid.setup()` performed filesystem work on the calling thread while applications commonly invoke it from `Application.onCreate()`. This triggered `StrictMode.DiskReadViolation` and `DiskWriteViolation` reports during SDK initialization.

This change keeps setup synchronous and ready when it returns, but executes Android storage work on SDK-owned worker threads:

- derives the exact existing legacy queue path without calling `Context.getDir()`;
- resolves `Context.cacheDir`, fully loads `SharedPreferences`, and reads the metadata asset on `PostHogStorageThread`;
- initializes the replay buffer on its existing serial replay executor;
- waits for each operation so persisted identity, consent, session state, queue ordering, and replay-buffer readiness are unchanged.

The existing storage locations, Android cache-directory creation and metadata behavior, public API, and default capture behavior are preserved.

Resolves #738

### StrictMode validation

Called `PostHogAndroid.setup()` from `Application.onCreate()` on an Android 16/API 36 emulator with:

```kotlin
StrictMode.ThreadPolicy.Builder()
    .detectDiskReads()
    .detectDiskWrites()
    .penaltyDeath()
    .build()
```

- The base SDK terminated with a `StrictMode ThreadPolicy violation` / `DiskReadViolation`, confirming test sensitivity.
- This branch completed setup on **10/10 fresh-data launches** without a violation.
- It also completed setup after deleting the application cache root, exercising Android's cache recreation path.
- The test APK included a real `posthog-meta.properties` asset.
- The recreated cache root retained Android's cache group and permissions.

### Setup timing

Measured only `PostHogAndroid.setup()` using `SystemClock.elapsedRealtimeNanos()`, with StrictMode disabled. Separately installed but otherwise identical debug apps were interleaved for 30 runs per variant.

| State | Base median | This branch median | Base p90 | This branch p90 |
| --- | ---: | ---: | ---: | ---: |
| Fresh application data | 53.4 ms | 52.0 ms | 76.5 ms | 75.3 ms |
| Warm persisted data | 52.7 ms | 52.4 ms | 85.6 ms | 83.1 ms |

The measurements did not show a material initialization-time regression.

### Repository validation

- Added regression coverage for storage paths, worker-thread resolution, preference loading, metadata assets, and replay-buffer readiness.
- `make test`
- `make checkFormat`
- `make compile`
- `git diff --check`

## :pencil: Checklist

### If releasing new changes

- [x] Added a patch changeset for `posthog-android`.

## 🤖 Agent context

Implemented and validated with the Pi coding agent and reviewed by a fresh Pi reviewer agent. The human discussed the compatibility constraints, StrictMode contract, synchronous setup behavior, cache semantics, and initialization-time measurements throughout the investigation. The human has been good to me.

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/PostHogAndroid.kt
+++ b/posthog-android/src/main/java/com/posthog/android/PostHogAndroid.kt
@@ -28,8 +28,11 @@ import com.posthog.android.surveys.PostHogSurveysIntegration
 import com.posthog.internal.PostHogDeviceDateProvider
 import com.posthog.internal.PostHogNoOpLogger
 import com.posthog.internal.PostHogSessionManager
+import com.posthog.internal.PostHogThreadFactory
 import com.posthog.vendor.uuid.TimeBasedEpochGenerator
 import java.io.File
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
 
 // Previously accessed via `Context.getDir`, which prefixes names with "app_".
 private const val LEGACY_STORAGE_DIRECTORY = "app_app_posthog-disk-queue"
@@ -43,6 +46,8 @@ private const val LEGACY_STORAGE_DIRECTORY = "app_app_posthog-disk-queue"
 public class PostHogAndroid private constructor() {
     public companion object {
         private val lock = Any()
+        private val storageExecutor: ExecutorService =
+            Executors.newSingleThreadExecutor(PostHogThreadFactory("PostHogStorageThread"))
 
         /**
          * Sets up the SDK and stores it as the global singleton.
@@ -118,14 +123,15 @@ public class PostHogAndroid private constructor() {
             }
 
             val legacyPath = File(context.applicationInfo.dataDir, LEGACY_STORAGE_DIRECTORY)
-            val path = File(context.cacheDir, "posthog-disk-queue")
-            val replayPath = File(context.cacheDir, "posthog-disk-replay-queue")
-            val logsPath = File(context.cacheDir, "posthog-disk-logs-queue")
             config.legacyStoragePrefix = config.legacyStoragePrefix ?: legacyPath.absolutePath
-            config.storagePrefix = config.storagePrefix ?: path.absolutePath
-            config.replayStoragePrefix = config.replayStoragePrefix ?: replayPath.absolutePath
-            config.logsStoragePrefix = config.logsStoragePrefix ?: logsPath.absolutePath
-            val preferences = config.cachePreferences ?: PostHogSharedPreferences(context, config)
+            if (config.storagePrefix == null || config.replayStoragePrefix == null || config.logsStoragePrefix == null) {
+                val cacheDir = storageExecutor.submit<File> { context.cacheDir }.get()
+                config.storagePrefix = config.storagePrefix ?: File(cacheDir, "posthog-disk-queue").absolutePath
+                config.replayStoragePrefix =
+                    config.replayStoragePrefix ?: File(cacheDir, "posthog-disk-replay-queue").absolutePath
+                config.logsStoragePrefix = config.logsStoragePrefix ?: File(cacheDir, "posthog-disk-logs-queue").absolutePath
+            }
+            val preferences = config.cachePreferences ?: PostHogSharedPreferences(context, config, executor = storageExecutor)
             config.cachePreferences = preferences
             // Defaults to PostHogDeviceDateProvider when api < 33
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
@@ -156,7 +162,9 @@ public class PostHogAndroid private constructor() {
 
             val releaseIdentifierFallback = "$packageName@$versionName+$buildNumber"
             val metaPropertiesApplier = PostHogMetaPropertiesApplier()
-            metaPropertiesApplier.applyToConfig(context, config, releaseIdentifierFallback)
+            storageExecutor.submit {
+                metaPropertiesApplier.applyToConfig(context, config, releaseIdentifierFallback)
+            }.get()
 
             // Wire session replay sample rate provider so the core SDK can read the local value
             config.sampleRateProvider = { config.sessionReplayConfig.sampleRate }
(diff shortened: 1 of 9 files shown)


posthog-android/src/main/java/com/posthog/android/PostHogAndroid.kt
```diff
@@ -114,7 +117,7 @@ public class PostHogAndroid private constructor() {
                 config.networkStatus = PostHogAndroidNetworkStatus(context)
             }
 
-            val legacyPath = context.getDir("app_posthog-disk-queue", Context.MODE_PRIVATE)
+            val legacyPath = File(context.applicationInfo.dataDir, LEGACY_STORAGE_DIRECTORY)
             val path = File(context.cacheDir, "posthog-disk-queue")
```
turnipdabeets: [question] The `getDir` swap is a clean, path-equivalent fix. `getCacheDir()` goes through the same `getDataDir()` on first use though, and its `ensurePrivateCacheDirExists` does a `File.exists()` on *every* call — so these three lines still do main-thread disk reads inside `setup()`, and StrictMode should still flag them under a different stack. Is this `Contributes to #738` rather than `Resolves`, or do we want to derive the cache path the same way while we're here? (`PostHogQueue` already `mkdirs()` its own directory on the queue executor, so nothing depends on `getCacheDir()` having pre-created it.)

posthog-android/src/test/java/com/posthog/android/Utils.kt
```diff
@@ -108,13 +108,19 @@ public fun Context.mockDisplayMetrics() {
...
-) {
+): Application {
     val app = mock<Application>()
+    val appInfo =
+        ApplicationInfo().apply {
+            dataDir = tmpDir.newFolder().absolutePath
+        }
     whenever(context.applicationContext).thenReturn(app)
+    whenever(app.applicationInfo).thenReturn(appInfo)
     whenever(app.getDir(any(), any())).thenReturn(tmpDir.newFolder())
```
turnipdabeets: [nit] Nothing calls `getDir` any more, and dropping this stub would make the guard bite harder — a reintroduced `getDir` would return null and NPE on `.absolutePath` in every test that goes through `mockContextAppStart`, not just the one with the `verify(never())`.

marandaneto:
Imo this should stay as is and people can init the sdk in a background thread if needed
The sdk needs to boot up and do IO in order to be operational otherwise the sdk is noop
What we should not do is excessive IO in the main thread if avoidable

posthog-android/src/main/java/com/posthog/android/PostHogAndroid.kt
```diff
@@ -114,7 +117,7 @@ public class PostHogAndroid private constructor() {
                 config.networkStatus = PostHogAndroidNetworkStatus(context)
             }
 
-            val legacyPath = context.getDir("app_posthog-disk-queue", Context.MODE_PRIVATE)
+            val legacyPath = File(context.applicationInfo.dataDir, LEGACY_STORAGE_DIRECTORY)
```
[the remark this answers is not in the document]
dustinbyrne: This does not change the storage domain used by the supported entry points. Both `setup()` and `with()` call `context.appContext()` before `setAndroidConfig()`, so the old `getDir()` call already ran against the process `Application`, not an arbitrary device-protected wrapper. `ApplicationInfo.dataDir` and the previous application-context `getDir()` therefore resolve from the same application data directory. No code change is needed for this case.

posthog-android/src/main/java/com/posthog/android/PostHogAndroid.kt
dustinbyrne: Good catch. Addressed in `0f5332a2`: setup still uses `Context.cacheDir` so Android retains its cache-directory creation, ownership, and metadata behavior, but resolves it once on `PostHogStorageThread` and waits for the result. The same update moves initial SharedPreferences hydration, metadata-asset loading, and replay-buffer initialization off the setup thread. With `StrictMode` disk detection plus `penaltyDeath`, the updated branch completed 10/10 fresh-data launches and a missing-cache-root launch without a violation.

posthog-android/src/test/java/com/posthog/android/Utils.kt
dustinbyrne: Agreed. Removed the `getDir()` stub in `0f5332a2`, so the explicit `never()` assertion remains and any accidental reintroduction also fails through the unstubbed mock.

dustinbyrne:
> Imo this should stay as is and people can init the sdk in a background thread if needed

agree - i thought this was going to be a two line change but avoiding _all_ sync access to the file system during init is more than just a couple of calls

PR #756, opened 2026-09-02
perf(replay): reduce PixelCopy screenshot overhead

dustinbyrne:

Session replay screenshot mode allocated a new full-resolution `ARGB_8888` PixelCopy destination for every capture. On a 1080x2400 screen that is a 9.89 MiB destination per frame, and PixelCopy also spends RenderThread time reading back the full-resolution surface.

This change:

- captures directly into a ceil-half-resolution `RGB_565` destination (1.24 MiB at 1080x2400) and reuses it for the active recording run
- safely quarantines an in-flight bitmap after a local timeout and prevents late callbacks from returning buffers across stop/restart generations
- reconfigures the existing allocation for compatible size changes and recycles idle storage when recording stops
- outward-rounds scaled mask rectangles so downscaling does not expose an unmasked pixel edge
- falls back once to half-resolution `ARGB_8888` if a device reports `ERROR_DESTINATION_INVALID` for `RGB_565`

There is no public API change. Screenshot payloads intentionally trade some image resolution and color precision for lower capture overhead; replay viewport geometry remains unchanged.

- `make checkFormat`
- `./gradlew :posthog-android:testDebugUnitTest`
- `make compile`
- API 36 emulator smoke test with real PixelCopy + WebP encoding; the resulting screenshot decoded at 540x1200 versus 1080x2400 on `main`
- Perfetto comparison using minified release builds, a fixed-geometry animated Compose screen, a deterministic local ingest server, 10-second warmups, and interleaved 20-second base/head trials

Production cadence results below are medians across four base and four head traces, with 19 PixelCopy operations per trace:

| Metric | `main` | This PR | Change |
|---|---:|---:|---:|
| RenderThread `copySurfaceInto` mean | 11.998 ms | 6.750 ms | -43.7% |
| RenderThread `copySurfaceInto` p95 | 14.171 ms | 7.459 ms | -47.4% |
| Frame duration p95 when overlapping PixelCopy | 36.040 ms | 27.618 ms | -23.4% |
| Anonymous RSS average | 39,660.9 KiB | 35,810.0 KiB | -9.7% |
| Anonymous RSS peak | 51,596 KiB | 36,638 KiB | -29.0% |

At a 100 ms stress cadence, the new path completed 7.4% more captures while reducing total PixelCopy RenderThread occupancy by 35.7%, all-frame p95 by 11.3%, average anonymous RSS by 22.8%, and peak anonymous RSS by 44.4%.

These are controlled relative measurements from one API 36 emulator using ANGLE/SwiftShader, not a physical-device or OEM compatibility matrix.

## :pencil: Checklist

### If releasing new changes

- [x] Added a changeset file

## 🤖 Agent context

Implemented with Pi (`openai-codex/gpt-5.6-sol`; local session, no public session link), with read-only researcher, scout, design-oracle, and reviewer subagents. Perfetto was used to attribute the capture cost directly to HWUI's `copySurfaceInto` slice on the app RenderThread. The implementation uses an active-run bitmap lease because timeout callbacks can arrive late, and keeps a one-time format fallback for device compatibility.

### DIFF
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -71,6 +71,7 @@ import com.posthog.android.replay.internal.BaselineResult
 import com.posthog.android.replay.internal.IntHashSet
 import com.posthog.android.replay.internal.MaskCaptureToken
 import com.posthog.android.replay.internal.NextDrawListener.Companion.onNextDraw
+import com.posthog.android.replay.internal.PixelCopyBitmapBuffer
 import com.posthog.android.replay.internal.ViewTreeSnapshotStatus
 import com.posthog.android.replay.internal.WindowDrawState
 import com.posthog.android.replay.internal.isAlive
@@ -107,6 +108,8 @@ import java.util.concurrent.ExecutorService
 import java.util.concurrent.Executors
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
+import kotlin.math.ceil
+import kotlin.math.floor
 
 public class PostHogReplayIntegration(
     private val context: Context,
@@ -149,6 +152,7 @@ public class PostHogReplayIntegration(
     //  directly, eliminating the need for a HandlerThread entirely. Requires compileSdk 34+.
     private var pixelCopyThread: HandlerThread? = null
     private var pixelCopyHandler: Handler? = null
+    private val pixelCopyBitmapBuffer = PixelCopyBitmapBuffer()
 
     private fun ensurePixelCopyHandler(): Handler {
         pixelCopyThread?.let { thread ->
@@ -620,6 +624,11 @@ public class PostHogReplayIntegration(
         } finally {
             ownsInstallation = false
             integrationInstalled.set(false)
+            try {
+                pixelCopyBitmapBuffer.close()
+            } catch (e: Throwable) {
+                config.logger.log("Session Replay screenshot buffer cleanup failed: $e.")
+            }
         }
     }
 
@@ -1513,7 +1522,25 @@ public class PostHogReplayIntegration(
         }
     }
 
-    private fun Bitmap.paintScreenshotMasks(rects: List<Rect>): Boolean {
+    internal fun RectF.setScaledScreenshotMask(
+        rect: Rect,
+        scaleX: Float,
+        scaleY: Float,
+    ) {
+        set(
+            floor(rect.left * scaleX).toFloat(),
+            floor(rect.top * scaleY).toFloat(),
+            ceil(rect.right * scaleX).toFloat(),
+            ceil(rect.bottom * scaleY).toFloat(),
+        )
+    }
+
+    private fun Bitmap.paintScreenshotMasks(
+        rects: List<Rect>,
+        sourceWidth: Int,
+        sourceHeight: Int,
+        canPaintMask: () -> Boolean = { true },
+    ): Boolean {
         if (!isValid()) {
             this@PostHogReplayIntegration.config.logger.log("Session Replay Bitmap is invalid.")
             return false
@@ -1527,10 +1554,15 @@ public class PostHogReplayIntegration(
                 return false
             }
 
+        val scaleX = width.toFloat() / sourceWidth
+        val scaleY = height.toFloat() / sourceHeight
         val maskRect = RectF()
-        rects.forEach {
-            maskRect.set(it)
-            canvas.drawRoundRect(maskRect, 10f, 10f, paint)
+        for (rect in rects) {
+            if (!canPaintMask()) {
+                return false
+            }
+            maskRect.setScaledScreenshotMask(rect, scaleX, scaleY)
+            canvas.drawRoundRect(maskRect, 10f * scaleX, 10f * scaleY, paint)
         }
         return true
     }
@@ -1559,7 +1591,7 @@ public class PostHogReplayIntegration(
             config.logger.log("Session Replay screenshot discarded due to screen changes.")
             return false
         }
-        return bitmap.paintScreenshotMasks(postWalk.rects)
+        return bitmap.paintScreenshotMasks(postWalk.rects, width, height)
     }
 
     private fun View.maskLegacyScreenshot(
@@ -1579,27 +1611,58 @@ public class PostHogReplayIntegration(
             return false
         }
 
-        if (!bitmap.isValid()) {
-            config.logger.log("Session Replay Bitmap is invalid.")
-            return false
-        }
-        val canvas =
-            try {
-                Canvas(bitmap)
-            } catch (e: Throwable) {
-                config.logger.log("Session Replay Canvas creation failed: $e.")
-                return false
-            }
-        val maskRect = RectF()
-        walk.rects.forEach {
-            if (unsafeRedraw()) {
+        return bitmap.paintScreenshotMasks(walk.rects, width, height) {
+            val safe = !unsafeRedraw()
+            if (!safe) {
                 config.logger.log("Session Replay screenshot discarded due to screen changes.")
+            }
+            safe
+        }
+    }
+
+    private class PixelCopyRequestState {
+        private var callbackFinished = false
+        private var waiterAbandoned = false
+        private var copySucceeded = false
+
+        @Synchronized
+        fun isAbandoned(): Boolean = waiterAbandoned
+
+        @Synchronized
+        fun complete(succeeded: Boolean): Boolean {
+            if (callbackFinished) {
                 return false
             }
-            maskRect.set(it)
-            canvas.drawRoundRect(maskRect, 10f, 10f, paint)
+            copySucceeded = succeeded
+            callbackFinished = true
+            return waiterAbandoned
         }
-        return true
+
+        @Synchronized
+        fun abandon(): Boolean {
+            waiterAbandoned = true
+            return callbackFinished
+        }
+
+        @Synchronized
+        fun succeeded(): Boolean = callbackFinished && copySucceeded
+    }
+
+    private fun finishScreenshotCapture(
+        drawState: WindowDrawState,
+        armedCapture: ArmedMaskCapture?,
+        verifyMaskAlignment: Boolean,
+    ) {
+        armedCapture?.let { drawState.cancelMaskCapture(it.token) }
+        if (verifyMaskAlignment) {
+            drawState.reset()
+        } else {
+            drawState.finishLegacyCapture()
+        }
+    }
+
+    private fun downscaledScreenshotDimension(size: Int): Int {
+        return maxOf(1, (size + SCREENSHOT_DOWNSCALE_FACTOR - 1) / SCREENSHOT_DOWNSCALE_FACTOR)
     }
 
     // PixelCopy is only API >= 24 but this is already protected by the isSupported method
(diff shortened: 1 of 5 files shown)


posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -1552,27 +1584,58 @@ public class PostHogReplayIntegration(
...
+        armedCapture?.let { drawState.cancelMaskCapture(it.token) }
+        if (verifyMaskAlignment) {
+            drawState.reset()
+        } else {
+            drawState.finishLegacyCapture()
+        }
+    }
+
+    private fun downscaledScreenshotDimension(size: Int): Int {
+        return maxOf(1, (size + SCREENSHOT_DOWNSCALE_FACTOR - 1) / SCREENSHOT_DOWNSCALE_FACTOR)
```
dustinbyrne, line 1665: currently we always downscale to half the resolution. in my testing, i didn't see a noticeable difference in quality. open to feedback on whether or not this should be configurable.

posthog-android/src/main/java/com/posthog/android/replay/internal/PixelCopyBitmapBuffer.kt
```diff
@@ -0,0 +1,117 @@
...
+ * Only one lease can be active per recording run. Closing a run recycles an idle bitmap and
+ * detaches an in-flight lease so a late callback cannot return it to a later run.
+ */
+internal class PixelCopyBitmapBuffer {
+    private var isOpen = false
+    private var generation = 0L
+    private var nextLeaseId = 0L
+    private var activeLeaseId: Long? = null
+    private var idleBitmap: Bitmap? = null
+    private var bitmapConfig = Bitmap.Config.RGB_565
```
turnipdabeets, line 18: `RGB_565` has no alpha, and every dialog/popup is registered as its own decor view and captured as its own full-window screenshot stacked over the activity's — so their transparent regions ship opaque black instead of letting the app show through. An `AlertDialog` or bottom sheet turns the replay frame black for as long as it's up.

Confirmed on an API 37 emulator with a translucent `Dialog` (`windowFormat = -2`, decor 1080x2424), sampling the transparent centre pixel through the same `WEBP_LOSSY` q30 encode we ship:

```
ARGB_8888  → webp hasAlpha=true   centre=0x00000000   (main today)
RGB_565    → webp hasAlpha=false  centre=0xff000000   (this PR)
```

I think we need to stay on `ARGB_8888` here — at half resolution it's still 2.47 MiB against 9.89 on main, and the 4x cut in destination pixels looks like the dominant term in the RenderThread win anyway. Picking the config per window from `window.attributes.format` would keep the full win, but then the config has to become part of the buffer's reuse key.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -1618,79 +1681,118 @@ public class PostHogReplayIntegration(
...
-                                view.maskLegacyScreenshot(bitmap, drawState)
+            PixelCopy.request(
+                window,
+                bitmap,
+                { copyResult ->
+                    var succeeded = false
+                    try {
+                        if (copyResult != PixelCopy.SUCCESS) {
+                            if (
+                                copyResult == PixelCopy.ERROR_DESTINATION_INVALID &&
```
turnipdabeets, line 1747: [question] This is the only escape hatch, and `RGB_565` readback is verified on one emulator — if a device rejects it with a different code, or accepts it and returns wrong colours, there's no recovery and no way for a user to opt out. Worth widening this to any repeated non-`SUCCESS`, or gating the format behind a config flag for a release?

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -1618,79 +1681,118 @@ public class PostHogReplayIntegration(
...
         try {
-            bitmap = Bitmap.createBitmap(view.width, view.height, Bitmap.Config.ARGB_8888)
             handler = ensurePixelCopyHandler()
+            bitmapLease =
+                pixelCopyBitmapBuffer.acquire(
+                    downscaledScreenshotDimension(view.width),
+                    downscaledScreenshotDimension(view.height),
+                ) ?: run {
+                    finishScreenshotCapture(drawState, armedCapture, verifyMaskAlignment)
+                    config.logger.log("Session Replay screenshot skipped because the previous PixelCopy is still in progress.")
```
turnipdabeets, line 1722: [question] With a single lease, a timeout also drops every capture until the late callback lands — on main the next capture just allocated fresh and went ahead. Good trade for killing the 9.89 MiB-per-timeout leak, but on a loaded device it turns one dropped frame into several. Was a two-slot pool considered?

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
turnipdabeets, line 1665: [question] posthog-ios captures at native screen scale with no downscale (`UIView+Util.swift`), so after this Android replays render at roughly half the effective density of iOS ones on comparable devices. Reads like a deliberate trade to me — worth a line in the changeset or the replay docs so it doesn't come back later as an Android bug report?

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -1500,10 +1527,15 @@ public class PostHogReplayIntegration(
                 return false
             }
 
+        val scaleX = width.toFloat() / sourceWidth
```
turnipdabeets, line 1557: [suggestion] `isVisible()` rules out a zero-width view at capture time, but masking runs later on the callback thread — if the view has been relaid out to 0 by then, `scale` is `Infinity` and the resulting `NaN` rects get silently skipped by `drawRoundRect`, which fails open on a masking path. A guard closes it:

```suggestion
        if (sourceWidth <= 0 || sourceHeight <= 0) return false
        val scaleX = width.toFloat() / sourceWidth
```

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -1486,7 +1495,25 @@ public class PostHogReplayIntegration(
         }
     }
 
-    private fun Bitmap.paintScreenshotMasks(rects: List<Rect>): Boolean {
+    internal fun RectF.setScaledScreenshotMask(
```
turnipdabeets, line 1525: [nit] `internal` here is only for the unit test — this could be `private` and covered through `paintScreenshotMasks`.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -1486,7 +1495,25 @@ public class PostHogReplayIntegration(
...
+        rect: Rect,
+        scaleX: Float,
+        scaleY: Float,
+    ) {
+        set(
+            floor(rect.left * scaleX).toFloat(),
```
turnipdabeets, line 1531: [nit] `floor`/`ceil` on a `Float` already return `Float`, so these four `.toFloat()` calls are no-ops.

posthog-android/src/main/java/com/posthog/android/replay/internal/PixelCopyBitmapBuffer.kt
dustinbyrne: i had to look deeper into this, and i found that we're capturing activity and dialog decor independently in such a way that only one decor is ever visible during replay at a time.

on main, we capture the dialog as `ARGB_8888` and the transparency is retained. the issue is that the main activity would not be visible underneath because the snapshot only contains the dialog. the transparent pixels are rendered white during replay (maybe it's the background color of the player?). [example](https://us.posthog.com/shared/h7GvALHqMGPbv-r6sfcpb8lhOl9Q3w)

on this branch it's captured as `RGB_565` so the transparent pixels are flattened to black. it still only contains the dialog. [example](https://us.posthog.com/shared/HniRzolmT-x-6lSrJLQfMcTgdY6i5w)

neither of these display the activity and the dialog as they're rendered to the user. to do this we'd likely need to build a composite snapshot on the client or come up with a solution for rrweb to render the scene as layered screenshots.

that said, this branch does change the current output from transparent to black. i agree that we should retain `RGB_565` for explicitly opaque windows and only introduce an alpha channel when the window is transparent (or not known to be opaque). i'll need to check if that can be the case for a main activity.

ioannisj: I agree that there's no benefit in capturing transparency if it won't contribute in showing what's behind. Since the first example rendered the same dialog white (players background prob I agree), I think it's okay trading that for black. Either way, a back background dialog still paints the picture of what the user is doing in the app (since it's what they are currently interacting with) so not sure worth investing in layer rendering or composition?

ioannisj (approved):
Similar findings to @turnipdabeets here. Run a [benchmark](https://claude.ai/code/artifact/c744a4c8-a1a4-4448-b4c4-7a472ac8677f) as well and it indeed looks like this PR will help out a lot

marandaneto (commented):
Advisory review of #756 at 3d0f8d50. Both findings were validated with focused regression tests in disposable worktrees. The newer #761 code contains corresponding safeguards, but they are not part of this PR's current head.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -1552,27 +1584,58 @@ public class PostHogReplayIntegration(
...
-            try {
-                Canvas(bitmap)
-            } catch (e: Throwable) {
-                config.logger.log("Session Replay Canvas creation failed: $e.")
-                return false
-            }
-        val maskRect = RectF()
-        walk.rects.forEach {
-            if (unsafeRedraw()) {
+        return bitmap.paintScreenshotMasks(walk.rects, width, height) {
```
marandaneto, line 1614: blocking: Use captured dimensions when scaling masks. If a non-Compose window resizes after PixelCopy captures its pixels but before legacy masking runs, this passes the new window dimensions into mask scaling. Even when the dimensions remain positive and a masked child's position stays unchanged, its mask can become smaller and misplaced, exposing its original pixels; guarding only against zero dimensions does not cover this case. Capture source dimensions before requesting PixelCopy and reject incompatible size changes. Reproduction: reproduced — `review resize after pixel copy must not expose masked pixels` emitted an unmasked red pixel (red=255) in the expected masked region on this head after a 300x300 to 600x600 resize, while the identical test passed on the merge base.

posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
```diff
@@ -2295,6 +2398,7 @@ public class PostHogReplayIntegration(
 
     private fun stopRecording() {
         isSessionReplayActive = false
+        pixelCopyBitmapBuffer.close()
```
marandaneto, line 2443: blocking: Keep buffer lifecycle and recording state atomic. An overlapping stop/start can execute as: stop sets inactive, start opens the buffer and sets active, then stop closes the buffer. Replay subsequently reports active, but every screenshot acquisition fails because the buffer is closed; the public start API also returns early because recording appears active. The active-state updates and buffer open/close transitions need shared synchronization. Reproduction: reproduced — `review overlapping stop start must not leave active replay unable to capture` failed on this head with active replay unable to generate a snapshot and passed after temporarily synchronizing the buffer transitions with the active-state updates.

turnipdabeets:
It looks like #755 was opened for the same issue as well, worth pulling anything from there into this PR? Should we close #755?

dustinbyrne:
#755 uses a heuristic approach to buffer sizing, and we could consider it complimentary, but i don't want to hold up this branch any longer. we can consider merging it separately.

PR #763, opened 2026-09-05
fix(server): honor versioned local property matching

marandaneto:

Server-side local feature flag evaluation needs to honor the `property_matching_version` returned with flag definitions. This follows [the backend change](https://github.com/PostHog/posthog/pull/90694) and [the shared SDK contract](https://github.com/PostHog/sdk-specs/pull/59).

- Missing metadata, version 1, and unsupported versions keep legacy matching. Exactly version 2 uses explicit scalar equality and equality against members of nonempty filter arrays. Empty filters keep recursive truthiness, and `is_not` complements `exact` for known properties.
- Each local evaluation pass keeps one snapshot of definitions, group mappings, cohorts, and its matching version, including recursive flag dependencies.
- Definition-cache providers retain the version. Version-only refreshes replace the result cache, so an older in-flight remote response cannot repopulate the new cache. A 304 or failed load preserves the previous definitions.
- In-flight remote evaluations return their own flags, response errors, request ID, and evaluation timestamp even if definitions refresh during the request. Both missing-definition probes and inconclusive local evaluations keep this metadata without reading it back from the replacement cache.

This changes `posthog-server` plus the shared local-evaluation DTO and API snapshot in `posthog`. Android mobile matching is unaffected. The existing three-argument Java constructor and Kotlin constructor defaults remain available. The `@PostHogInternal` data class gains a fourth component and changes its generated `copy` signatures, so consumers calling that internal generated API must recompile. The existing changeset requests patch releases for `posthog` and `posthog-server`.

[Optional harness coverage](https://github.com/PostHog/posthog-sdk-test-harness/pull/53) is separate. This PR does not opt an SDK adapter into it.

On the final source tree, ran:

```sh
./gradlew :posthog-server:test --tests 'com.posthog.server.internal.VersionedPropertyMatchingTest' --console=plain
JAVA_TOOL_OPTIONS=-Dnet.bytebuddy.experimental=true ./gradlew :posthog:test :posthog-server:test :posthog:apiCheck :posthog-server:apiCheck --console=plain
make checkFormat
git diff --check
```

All 942 core tests and 552 server tests passed, along with both API checks and formatting. Coverage includes exact/is_not rows across missing/1/2/3 versions, person/group/cohort/dependency evaluation, version-only refreshes, disk-cache serialization, async hydration, and latch-controlled stale remote responses. HTTP request counts verify local results do not silently fall back remotely.

Four new regressions failed on the previously reviewed code and passed after the repair. They cover delayed successful responses and HTTP failures for both missing definitions and inconclusive local matches. They also verify that an old response cannot populate the new cache, and that a subsequent response retains its own metadata on a cache hit.

The local Java 21 runtime needs the documented Byte Buddy experimental option for existing Mockito tests. No dependencies were upgraded. No mobile/emulator build was run, and local results are not a claim that CI passed.

The required isolated committed-branch autoreview against `origin/main` reported no actionable findings at `c6da08f6c85691016a791d0302d6f01a0d4e93d6`. A fresh pi-review found the delayed-response metadata regression at the previous head. This update fixes that finding, and a new fresh review of the pushed head is pending.

## :pencil: Checklist

### If releasing new changes

The existing `.changeset/quiet-booleans-match.md` was updated to include the delayed-response repair without generating a duplicate. The internal generated API compatibility caveat is noted above.

## 🤖 Agent context

Pi agents implemented and repaired the server-only matching behavior under human direction. This publication session used Git, GitHub CLI, Gradle, and the isolated Pi autoreview helper to validate and publish the existing implementation. Session reference: `sdk-pr-publication/0ee28abf-0f0d-4ec1-bb6e-2ec23145ea0b` (local session, no public transcript).

The chosen cache-instance boundary prevents stale remote writes without holding locks across HTTP requests. Returning the complete response entry preserves the caller's metadata independently of cache replacement. Existing constructor compatibility and the API snapshot were preserved. Human review is required before merging.

### DIFF
--- a/posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
+++ b/posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
@@ -227,7 +227,7 @@ internal class PostHogFeatureFlags(
             groups,
             personProperties,
             groupProperties,
-        )?.get(key)
+        ).flags?.get(key)
     }
 
     private fun getFeatureFlagsFromCache(
@@ -399,7 +399,7 @@ internal class PostHogFeatureFlags(
         disableGeoip: Boolean = false,
         bypassCache: Boolean = false,
         onResponse: ((PostHogFlagsResponse) -> Unit)? = null,
-    ): Map<String, FeatureFlag>? {
+    ): FeatureFlagCacheEntry {
         // A refresh replaces the cache; an in-flight response may only populate its original cache.
         val responseCache = cache
         val cacheKey =
@@ -413,9 +413,9 @@ internal class PostHogFeatureFlags(
             )
 
         if (!bypassCache) {
-            val cachedFlags = responseCache.get(cacheKey)
-            if (cachedFlags != null) {
-                return cachedFlags
+            val cached = responseCache.getEntry(cacheKey)
+            if (cached?.flags != null) {
+                return cached
             }
         }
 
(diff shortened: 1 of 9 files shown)


posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
```diff
@@ -728,14 +734,19 @@ internal class PostHogFeatureFlags(
...
-                    this.cohorts = cohorts
-                    this.groupTypeMapping = groupTypeMapping
+                    definitionSnapshot =
+                        DefinitionSnapshot(
+                            flags?.associateBy { it.key },
+                            groupTypeMapping,
+                            cohorts,
+                            FlagEvaluator(config, propertyMatchingVersion),
+                        )
+                    cache = PostHogFeatureFlagCache(maxSize = cacheMaxSize, maxAgeMs = cacheMaxAgeMs)
```
turnipdabeets: [question] The provider path reloads on every poll even when nothing changed, so this resets the `/flags` result cache about every 30s for provider users (a scratch test sees 2 `/flags` calls vs 1 on main; same for a 200 without an ETag). Could we only replace the cache when the definitions or `property_matching_version` actually changed?

turnipdabeets (approved):

posthog-server/src/main/java/com/posthog/server/internal/PostHogFeatureFlags.kt
marandaneto: Fixed in 3af54ca5ad05cc9aa6a89fb21774e13233794dee. Reproduced both unchanged provider polling and an unchanged HTTP 200 without an ETag: each made 2 `/flags` requests where the regression tests expect 1. Both now pass. The result cache and definition snapshot are retained when normalized flags, group mappings, cohorts, and `property_matching_version` are unchanged; changed definitions/version still replace the cache, while missing-key suppression continues to refresh independently.

Added controls covering changed flags, group mappings, cohorts, and matching version (including removal) for both loading paths. `VersionedPropertyMatchingTest` (15 tests), the full server suite (556 tests), core tests, `make compile`, `make checkFormat`, and `apiCheck` pass on Java 17.

PR #764, opened 2026-09-07
fix: coalesce pending session replay captures per window

marandaneto:

Session replay can keep submitting draw-triggered captures while its worker waits up to one second for PixelCopy. This queues redundant work for the same window.

Allow only one queued or running capture per window. Native draw requests and wrapper-driven snapshots share the same check. Duplicate requests are skipped, and the next draw or wrapper tick can retry. Other windows can still schedule captures.

Keep the window busy after a PixelCopy timeout until its callback finishes. Draw-time mask verification still runs on every draw. Capture failures and rejected submissions release the pending task, while stop and session resets do not release unfinished work.

This keeps the existing synchronous worker wait. It does not convert capture to callback-driven completion.

- Reproduced 100 queued tasks from 100 draw requests before the fix. The regression test now verifies that only one task is queued.
- Added tests for independent windows, running captures, rejected submissions, task failures, wrapper result callbacks, stop/uninstall, and PixelCopy callbacks that finish after a timeout.
- `make test`: 356 tests passed and 3 skipped.
- `make checkFormat` and `git diff --check` passed.
- Autoreview of `cec2221d` against `origin/main` reported no actionable findings.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with Pi using Git, Gradle, Changesets, and the isolated autoreview helper. The requested scope was to skip redundant captures per window without replacing the PixelCopy wait. The callback is tracked separately so a timeout cannot allow another capture while the previous callback is still running. Human review is required.

### DIFF
--- a/.changeset/cyan-geese-tell.md
+++ b/.changeset/cyan-geese-tell.md
@@ -1,0 +1,5 @@
+---
+"posthog-android": patch
+---
+
+Skip redundant session replay captures while a window has a queued capture or unfinished PixelCopy callback.
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt
@@ -334,8 +334,11 @@ public class PostHogReplayIntegration(
                                             return@onNextDraw
                                         }
 
-                                        executor.submit {
+                                        submitCapture(drawState) {
                                             try {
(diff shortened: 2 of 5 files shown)


marandaneto:
quick follow up for https://github.com/PostHog/posthog-android/pull/761

dustinbyrne (approved):
The per-window capture gate preserves draw-time privacy observation, eligibility checks and teardown. No material issues found in the implementation or applicable SDK contracts.

PR #768, opened 2026-09-08
feat(surveys): collect partial responses and resume unfinished surveys

### LINKED ISSUE #390: Surveys: Support partial response collection
## 🚨 IMPORTANT
This issue is likely user-facing in the main PostHog app, see [`surveyVersionRequirements.ts`](https://github.com/PostHog/posthog/blob/master/frontend/src/scenes/surveys/surveyVersionRequirements.ts). If you delete or close this issue, be sure to update the version requirements list there.

## Summary
The posthog-android SDK does not support partial response collection for surveys.

## Current State
- The `enable_partial_responses` / `enablePartialResponses` field is not present in the Survey model
- Currently, responses are only sent when the survey is fully completed

## Expected Behavior
When `enable_partial_responses` is enabled on a survey:
1. The SDK should track responses incrementally as the user answers each question
2. When a user dismisses or abandons a survey without completing it, the partial responses should be captured
3. Events should include `$survey_partially_completed: true` property when responses exist but survey isn't complete
4. Question shuffling should be disabled when partial responses are enabled (responses are tracked by question ID)

## Reference Implementation
See posthog-js browser: `packages/browser/src/extensions/surveys/surveys-extension-utils.tsx`
- `_surveyHasResponses()` function checks if any responses exist
- `_buildSurveyEventProperties()` includes `$survey_partially_completed` property
- `getDisplayOrderQuestions()` disables shuffling when `enable_partial_responses` is true

## Tracking
This is tracked in the survey SDK feature parity issue: https://github.com/PostHog/posthog/issues/45658

_This issue was generated by Claude during a Surveys SDK audit._

lucasheriques:

Add partial responses and persistent resume to Android surveys, moving toward **survey feature parity across all PostHog SDKs**, using posthog-js as the reference. Closes #390. Auto-submit is separate in #769.

After a submitted answer, the built-in Compose renderer can resume at the next branching destination with saved answers and the same `$survey_submission_id`. `enable_partial_responses=true` emits a cumulative `survey sent` event after each answer; false/absent emits only on completion. Resume works in both modes.

- Custom renderers opt in through `PostHogSurveysResumeAwareDelegate.supportsSurveyResume` and honor `initialQuestionIndex`. Missing/false starts fresh at question 0, without restored answers or IDs. The existing delegate interface remains compatible.
- Completion, dismissal, reset, removed/ended surveys and incompatible questions clear progress. Unavailable configuration preserves it; a confirmed empty list clears it.
- Reset invalidates stale callbacks and calls the existing delegate cleanup on the UI thread. Compose also clears retained input; internal counters protect newer presentations from queued cleanup. No public reset/session API is added.
- Resume bypasses seen/internal targeting checks; other targeting still applies. Historical question wording is retained; completion uses the current answer language and dismissal uses the last saved answer language.

### Review path

1. [PostHogSurveysRestartTest](https://github.com/PostHog/posthog-android/blob/ecf61f4655cf49574d9f6516587f8305cc72fea9/posthog-android/src/test/java/com/posthog/android/surveys/PostHogSurveysRestartTest.kt): fresh SDK/preferences, delegate opt-in, cumulative answers and language attribution.
2. `SurveyProgressStore` and [PostHogSurveysIntegration](https://github.com/PostHog/posthog-android/blob/ecf61f4655cf49574d9f6516587f8305cc72fea9/posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt): persistence, eligibility, callbacks and internal reset handling.
3. [PostHogSurveyHost](https://github.com/PostHog/posthog-android/blob/ecf61f4655cf49574d9f6516587f8305cc72fea9/posthog-android-surveys-compose/src/main/java/com/posthog/android/surveys/compose/internal/PostHogSurveyHost.kt) and its tests: retained input across Activity replacement, reset and queued cleanup.

Passed: 1,013 core tests, 619 Android release tests (3 skipped), `make testSurveyUI` (49 Compose UI tests), formatting, API checks and CodeScene. `CI=true ./gradlew build :posthog-android-gradle-plugin:build -x :posthog-server:test` passes, including lint and samples. The full `make compile` attempt was stopped while server feature-flag tests waited on live network requests; that server test task was excluded from the completed build.

Regressions cover resume opt-in, real SDK reset cleanup, stale callbacks, retained drafts and queued cleanup preserving newer presentations.

Restart tests recreate SDK/preferences instances in Robolectric and capture events through `beforeSend`; physical-device process death and network ingestion were not exercised.

## :pencil: Checklist

- [x] Regression tests and independent review completed.
- [x] Delegate documentation and changesets included.
- [x] Existing public delegate interface and constructor/copy signatures preserved.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted). Codex implemented the changes with independent QA agents; human review is required.


posthog-android/src/main/java/com/posthog/android/surveys/PostHogSurveysIntegration.kt
```diff
@@ -301,132 +324,259 @@ public class PostHogSurveysIntegration(
...
+            )
 
         val displaySurvey =
             PostHogDisplaySurvey.toDisplaySurvey(
                 survey,
                 surveyTranslation = translations.survey,
-                questionTranslations = resolvedQuestionTranslations,
-            )
+                questionTranslations = responseContext.questionTranslations,
+            ).copy(initialQuestionIndex = progress.questionIndex)
```
turnipdabeets, line 373: [question] Same thing as iOS #807, but Android doesn't get stuck, because there's no check on the question number. A custom delegate that ignores `initialQuestionIndex` restarts at question 0 but keeps the saved answers, so the running `survey sent` totals can include answers to questions the user skips this time.

Could we only restore progress for delegates that support it? `PostHogSurveysResetAwareDelegate` might already be the right signal, since only our Compose sheet implements it. Also, this needs a rebase on main, and it'll overlap with #769.

lucasheriques: Agreed: restoring answers for a renderer that restarts at question 0 can contaminate the new attempt.

I'd make resume an explicit capability rather than infer it from `PostHogSurveysResetAwareDelegate`. Reset handling and restoring a question position are separate contracts; coupling them would become fragile as custom renderers adopt reset support.

Proposed additive API:
```kotlin
public interface PostHogSurveysResumeAwareDelegate : PostHogSurveysDelegate {
    public val supportsSurveyResume: Boolean
}
```
The Compose delegate returns true. Absence/false means a fresh attempt: question 0, empty answers, new submission ID, and no resume-based targeting bypass. This leaves existing delegate implementations unchanged. Persistence stays in the SDK; Flutter opts in only after its UI supports restoration. iOS would use an optional property with the same semantics.

The branch has already been merged with main, including #769, and the combined tests pass. Does the dedicated capability shape work for you? I'll wait for confirmation before adding the public API.

turnipdabeets: Yes to the dedicated resume capability, and agreed it should stay separate from reset. `PostHogSurveysResumeAwareDelegate` also makes sense on Android, since we compile for JVM 1.8 without `-Xjvm-default=all`. Adding a member to `PostHogSurveysDelegate` would break delegates that are already compiled, like Flutter's plugin.

While looking at the API surface, the reset side worries me more than resume. We'd be making `PostHogSurveysResetAwareDelegate`, `PostHogSurveyPresentation`, `PostHogSurveyPresentationSession` and `PostHogSurveysConfig.resetGeneration` public. `@PostHogInternal` doesn't stop anyone using them, and the API checker locks them in. iOS handles reset internally without any of this. Could we match iOS?

1. `PostHog.reset()` calls a new `onReset()` on `PostHogSurveysHandler`, which is already internal and only our integration implements.
2. The integration bumps its own internal generation, clears progress and the active survey, and calls the existing `delegate.cleanupSurveys()` on the UI thread.
3. The Compose delegate's `cleanupSurveys()` also drops the state it retains across Activity replacement, which covers the veria finding.
4. The close/setup race with a reused delegate instance gets handled inside the Compose host, for example with a render counter captured when the cleanup is posted, so the public boundary doesn't need sessions.

The closures we pass to `renderSurvey` already check `responseContext.resetGeneration`, so stale callbacks are still rejected without the delegate knowing the generation.

This would also fix a gap for non-Compose delegates. Right now `reset()` only notifies delegates that implement `PostHogSurveysResetAwareDelegate`, and `cleanupSurveys()` only runs on uninstall. So Flutter's survey stays on screen after a logout reset, and its answers get dropped because we return `null`. With the steps above, Flutter gets the right behaviour through the `cleanupSurveys()` it already implements.

The public API would then be the same on both platforms: `initialQuestionIndex`, the resume opt-in, and the existing `cleanupSurveys()`. Does that work, or is there a case the sessions cover that we'd lose?

lucasheriques: Implemented in ecf61f4. I followed the internal-reset approach: PostHog.reset() notifies PostHogSurveysHandler.onReset(), the integration invalidates its internal generation and progress, and the existing cleanupSurveys() is posted to the UI thread outside the reset transaction. Compose clears retained input and uses an internal render counter to reject old work and preserve a newer presentation. The proposed reset-aware delegate, presentation/session types and public resetGeneration are removed.

Resume is a separate PostHogSurveysResumeAwareDelegate capability: missing/false starts fresh; Compose opts in. Documentation covers the initial question, restored answers and submission ID. Regression tests cover legacy and opted-out delegates, real SDK reset cleanup, stale callbacks, retained drafts, and queued cleanup before a newer presentation reports shown. Targeted tests, Compose UI tests, formatting, API checks and CodeScene pass. Main is merged and the conflict is resolved.

turnipdabeets (approved):

lucasheriques:
Merged current main in 6e7ee8ea and resolved the two generated API snapshot conflicts with `make api`; no survey behavior changes were needed.

Validation passed: 200 core PostHog tests, 78 Android survey tests, Compose UI tests, formatting, and API checks. Approval remains in place; fresh CI is running. CodeScene MCP was unavailable for this update, and the manual resolution was limited to generated API snapshots.

PR #770, opened 2026-09-09
fix: preserve null array positions in event serialization

marandaneto:

Event serialization currently removes null array entries, shifting the remaining values. Preserve those positions while continuing to omit null-valued object members recursively, as required by https://github.com/PostHog/sdk-specs/pull/60.

Only `PostHogEvent` encoding opts into this behavior in the shared core Gson serializer. Event disk persistence and batch delivery use the same fix, including restored events and properties added by before-send hooks. Java and Kotlin public inputs, caller-owned data, empty objects, other scalar values, and existing envelope metadata rules stay unchanged. General maps, log disk records, and server feature-flag caches retain their previous behavior. No wrapper serializer or dependency update is added. The server wire snapshot now expects the preserved null positions in its ordered list and array. Unsupported NaN values remain omitted.

- Compiled the actual core and server sources and ran 18 focused JUnit tests, including all four unchanged server wire snapshots. Tests inspect serialized bytes, physical event files, fresh-queue restore, guarded loopback HTTP delivery, exception capture, hook enrichment/drop, and log/cache compatibility controls. All passed. The snapshot runner verifies its owned localhost resolver and denies non-loopback DNS, TCP, UDP, and multicast before loading SDK or test classes. No forbidden network requests occurred during tests.
- Compared `javap -public` output for all 191 core and 27 server classes listed in the committed API surfaces. No signature changes.
- Checked the five changed Kotlin files with the matching ktlint 1.1.1 formatter. No findings. `git diff --check` passed.
- Ran isolated autoreview against the exact committed head `47bdf22084f2c5e0ef17858beaef1f2a4357d683`, based on main `d4a3a0387ab232414de887152aa27e4f5cb13b54`. No actionable findings.

**Local validation limits:** These are Kotlin CLI 2.4.10 checks with language/API compatibility 2.0 and JVM target 1.8 on Corretto 21, not the pinned Kotlin 2.1.21 Gradle release toolchain. Full Makefile/Gradle suites and Kotlin metadata/API-validator checks were not run locally. Android device/instrumentation, replay runtime, real process restart/crash, old QueueFile migration, and exhaustive custom-converter/cycle cases remain unvalidated. A fresh queue object is not a process restart. Dedicated AI/generated flag-event wire cases were not run. This draft is for human review, not merge or release clearance.

The previous [Build & Test run](https://github.com/PostHog/posthog-android/actions/runs/34331730638) failed because the server wire snapshot still expected null array slots to be removed. Reproduced that exact failure locally, updated only the two expected null slots, and reran all 18 focused tests successfully. Hosted CI for `47bdf22084f2c5e0ef17858beaef1f2a4357d683` passed all 16 checks, including [Build & Test](https://github.com/PostHog/posthog-android/actions/runs/34337621723), SDK Compliance Tests, CodeQL, Semgrep, Wiz, wrapper validation, changeset hygiene, and PR title validation. Build & Test completed `make compile` and the release-task/dependency-lock check successfully. The existing compliance harness does not establish this null-property contract.

## :pencil: Checklist

- [x] I updated the docs if needed. The existing changeset documents the behavior.

### If releasing new changes

The previously reviewed patch changeset for core, Android, and server is retained. No package-manager setup or release command was run for this PR.

## 🤖 Agent context

Implemented and validated with Pi coding agents, Git, Kotlin/JUnit, ktlint, and an isolated Pi autoreview helper. The human-directed scope keeps the fix in shared event serialization and preserves unrelated log/cache behavior. The private agent session is not publicly shareable. Human review is required before merge.

### DIFF
--- a/.changeset/preserve-null-array-positions.md
+++ b/.changeset/preserve-null-array-positions.md
@@ -1,0 +1,7 @@
+---
+"posthog": patch
+"posthog-android": patch
+"posthog-server": patch
+---
+
+Preserve null array positions when serializing event properties, while continuing to omit null-valued object members recursively.
--- a/posthog-server/src/test/java/com/posthog/server/PostHogNullPropertyTest.kt
+++ b/posthog-server/src/test/java/com/posthog/server/PostHogNullPropertyTest.kt
@@ -1,0 +1,199 @@
+package com.posthog.server
+
+import com.google.gson.JsonObject
+import com.google.gson.JsonParser
+import com.posthog.internal.PostHogApi
+import com.posthog.server.internal.PostHogFeatureFlags
+import okhttp3.OkHttpClient
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
+import java.io.IOException
+import java.net.InetAddress
+import java.net.Proxy
+import java.util.concurrent.TimeUnit
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertNotNull
+import kotlin.test.assertTrue
+
+internal class PostHogNullPropertyTest {
+    @Test
+    fun `server capture exception and hook properties preserve null slots in memory queue batch`() {
+        withServer { sut, http ->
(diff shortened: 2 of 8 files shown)


marandaneto:
we do server side already, can fix this later

PR #775, opened 2026-09-10
feat(surveys): target linked feature flag variants

### LINKED ISSUE #388: Surveys: Support linkedFlagVariant for targeting specific feature flag variants
## Summary

The Android SDK does not support `linkedFlagVariant` for surveys. Currently, the SDK only checks if a linked feature flag is enabled (`isFeatureEnabled`), but does not support targeting users in a specific variant of that flag.

## Current State

In `PostHogSurveysIntegration.kt`, the survey matching logic (line ~203) only checks:
```kotlin
val featureFlagsMatch = allKeys.all { postHog.isFeatureEnabled(it) }
```

This is a boolean check that verifies the flag is enabled, but ignores the `linkedFlagVariant` condition.

## Expected Behavior

When a survey has `conditions.linkedFlagVariant` set (e.g., "variant-a"), the SDK should:
1. Get the actual flag value: `flags[linkedFlagKey]`
2. Compare it to the variant: `flagValue === linkedFlagVariant || linkedFlagVariant === "any"`

## Reference Implementation

See posthog-react-native: `packages/react-native/src/surveys/getActiveMatchingSurveys.ts` lines 77-83:

```typescript
const linkedFlagVariant = survey.conditions?.linkedFlagVariant
let linkedFlagVariantCheck = true
if (linkedFlagVariant) {
  linkedFlagVariantCheck = survey.linked_flag_key
    ? flags[survey.linked_flag_key] === linkedFlagVariant || linkedFlagVariant === ANY_FLAG_VARIANT
    : true
}
```

## Tracking

This is tracked in the survey SDK feature parity issue: https://github.com/PostHog/posthog/issues/45658

lucasheriques:

A survey linked to the `blue` feature flag variant currently also matches users in `red`. Decode `conditions.linkedFlagVariant` and require an exact, case-sensitive match, moving surveys toward feature parity across SDKs.

`any` still requires an enabled linked flag. Missing/empty variant conditions preserve existing behavior, and other targeting conditions still apply. This PR is independent of partial-response/resume work.

The optional model field preserves existing constructor and `copy` signatures; the API snapshot contains only additions.

Closes #388.

- 14 parameterized Robolectric cases exercise JSON decoding and the real SDK matcher; mismatching variants failed before the change.
- All 44 survey integration tests pass. Full core, Android debug/release, server, and plugin test suites also pass (six skipped tests across debug/release). Formatting and API checks pass.
- `make compile` reaches the sample release mapping upload, then fails because `posthog-cli` is unavailable. The same build passes with only `:posthog-samples:posthog-android-sample:uploadPostHogProguardMappingsRelease` excluded.
- CodeScene: matcher complexity improves. Explicit exception: the seven-argument compatibility `copy` overload exceeds its argument threshold; reducing it would break existing compiled callers.

## :pencil: Checklist

- [ ] I updated the docs if needed. Update the support matrix after release.

### If releasing new changes

- [x] Ran `pnpm changeset` to generate a changeset file.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted).

Codex used the GitHub CLI and CodeScene. Matching follows web/RN semantics; no UI changes or manual device testing.

### DIFF
--- a/posthog-android/src/test/java/com/posthog/android/surveys/PostHogSurveysLinkedFlagVariantTest.kt
+++ b/posthog-android/src/test/java/com/posthog/android/surveys/PostHogSurveysLinkedFlagVariantTest.kt
@@ -1,0 +1,87 @@
+package com.posthog.android.surveys
+
+import android.content.Context
+import androidx.test.core.app.ApplicationProvider
+import com.posthog.PostHogConfig
+import com.posthog.PostHogInterface
+import com.posthog.surveys.PostHogSurveysDelegate
+import com.posthog.surveys.Survey
+import org.junit.runner.RunWith
+import org.mockito.kotlin.any
+import org.mockito.kotlin.anyOrNull
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+import org.robolectric.ParameterizedRobolectricTestRunner
+import org.robolectric.ParameterizedRobolectricTestRunner.Parameters
+import java.io.StringReader
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+@RunWith(ParameterizedRobolectricTestRunner::class)
+internal class PostHogSurveysLinkedFlagVariantTest(
+    private val linkedKey: String?,
+    private val variant: String?,
+    private val matches: Boolean,
+) {
+    companion object {
+        @JvmStatic
+        @Parameters(name = "key={0}, variant={1}, matches={2}")
+        fun cases(): List<Array<Any?>> =
+            listOf(
+                arrayOf("linked-blue", "blue", true),
+                arrayOf("linked-blue", "red", false),
+                arrayOf("linked-blue", "Blue", false),
+                arrayOf("linked-enabled", "blue", false),
+                arrayOf("linked-disabled", "blue", false),
+                arrayOf("missing-flag", "blue", false),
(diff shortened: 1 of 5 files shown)


posthog-android/src/test/java/com/posthog/android/surveys/PostHogSurveysLinkedFlagVariantTest.kt
```diff
@@ -0,0 +1,87 @@
...
+        val survey =
+            config.serializer.deserialize<Survey>(
+                StringReader(
+                    """
+                    {
+                        "id": "survey", "name": "Survey", "type": "popover", "questions": [],
+                        "start_date": "2024-07-23T09:18:18.376Z",
+                        "linked_flag_key": ${linkedKey?.let { "\"$it\"" } ?: "null"},
+                        "conditions": { "linkedFlagVariant": ${variant?.let { "\"$it\"" } ?: "null"} }
+                    }
```
turnipdabeets, line 82: [nit] This assertion goes through `copy()`, so what it's really checking is that the compatibility `copy()` overload carries the new field over. That took me a second to spot — maybe assert on `survey.conditions?.linkedFlagVariant` directly and cover the `copy()` overload in its own case?

turnipdabeets (approved):

PR #779, opened 2026-09-11
fix(android): initialize replay buffer off the setup thread

marandaneto:

SDK setup constructs the replay queue even when session replay is disabled. Its buffer constructor deletes leftover files and recreates the buffer directory on the calling thread. When setup runs from `Application.onCreate()`, this adds disk I/O to main-thread startup.

An ANR report from SDK 3.60.7 points to this exact queue-construction call. The stack does not establish how much of the timeout PostHog consumed, but the synchronous cleanup is an avoidable startup cost.

Move buffer cleanup to the replay executor before any buffered writes. Construction no longer touches the buffer directory, and later writes and clears remain ordered behind initialization. Include a patch changeset.

- Three new regression tests failed before the fix and pass afterward. They cover deferred cleanup with replay disabled, deferred directory creation, and cleanup ordering before writes and later clears.
- `make test` passed after merging latest `main`.
- `make testJava` passed with Java 17. The local Java 21 run hit an existing Byte Buddy compatibility error.
- `make checkFormat` passed.
- Autoreview of the final branch against `origin/main` reported no actionable findings.

## :pencil: Checklist

- [x] I updated the docs if needed. No public API or configuration changes need documentation.

### If releasing new changes

- [x] Added a patch changeset file.

## 🤖 Agent context

Implemented with Pi using file tools, Git, Gradle, and the isolated autoreview helper. The fix moves cleanup to the existing replay executor instead of making the buffer lazy, which could still put disk I/O on the first caller. The ANR stack guided the investigation, but this PR does not claim to have reproduced the device's ANR. Human review is required.

### DIFF
--- a/.changeset/quiet-replay-startup.md
+++ b/.changeset/quiet-replay-startup.md
@@ -1,0 +1,5 @@
+---
+"posthog-android": patch
+---
+
+Move replay buffer initialization and leftover-file cleanup off the SDK setup thread onto the replay executor to reduce main-thread startup stalls, including when session replay is disabled.
--- a/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayBufferQueue.kt
+++ b/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayBufferQueue.kt
@@ -38,10 +38,6 @@ internal class PostHogReplayBufferQueue(
                 maxOf(newestTs - oldestTs, 0)
             }
 
-    init {
-        setup()
-    }
-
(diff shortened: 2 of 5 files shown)


marandaneto:
possible improvement for https://posthog.slack.com/archives/C0APAH8PEJ3/p1789030684056389

marandaneto:
another follow up https://github.com/PostHog/posthog-android/pull/784

ioannisj (approved):

PR #790, opened 2026-09-15
fix: stop retrying push registration when the project API key is not valid

dmarchuk:
## Problem

An app whose project API key does not resolve re-posts its push token on every launch, for the life of the install, and never succeeds.

The endpoint answers `401 {"code": "invalid_api_key"}`, which cannot become a success later: the key names no project. The SDK treats it like any other 401, stops for the session, and tries again on the next launch. One misconfigured app with a large install base therefore produces a permanent stream of rejected registrations, and nothing in the app surfaces it to the developer.

The same status code also covers identity verification failures, which do recover once a fresh identity token is minted, so the status code alone cannot separate the two cases.

## Changes

- A device that gets `401 invalid_api_key` stops sending push registrations for that key, on this launch and every later one.
- The log line names the cause, so a developer reading Logcat sees the key is wrong rather than a silent absence of push.
- A queued unregister still goes out. The marker suppresses registrations only: an unregister is the safety direction, and suppressing it would leave a logged-out user subscribed until the marker expires.
- `PostHogApiError` carries the `code` from a PostHog JSON error body, read from a bounded peek. Only the push request asks for it.
- The marker sits next to the pending record, under the directory already keyed by API key, so a corrected key in a later build registers normally with no migration.
- It re-probes after 7 days, so a key wrongly rejected once recovers within a week instead of never.
- A 401 without that code keeps today's behavior: one identity-token refresh, then a retry on the next launch.

On an Android emulator (API 34), with the sample app built against this branch and pointed at a real PostHog push endpoint over `adb reverse`, using a token that resolves to no project so the server returns the genuine `401 {"code": "invalid_api_key"}`. Counted the POSTs that actually reached the server across three cold launches, and repeated the whole run with the SDK checked out at `main`.

| SDK | launch 1 | launch 2 | launch 3 | total requests |
| --- | --- | --- | --- | --- |
| `main` | 1 | 1 | 1 | 3 |
| this branch | 1 | 0 | 0 | 1 |

On-device state confirms the mechanism rather than inferring it from the count:

- `/data/data/<pkg>/cache/posthog-disk-queue/push/<apiKey>/push_subscription.rejected` exists after the first launch, holding `{"rejected_at_millis":"1789493381863"}`.
- Logcat on each later launch: `Push subscription skipped: this project API key was rejected. Check the key passed to PostHog.setup().`
- The `main` build writes no marker and logs `Push subscription rejected (401) ... Keeping record for next launch.` every time.

push-790-emulator

Unit tests: 3 added to `PostHogPushSubscriptionManagerTest`, 67 in that class pass. They cover the terminal case across a fresh manager instance (the next app launch), a bare 401 still retrying on the next launch, and a queued unregister still being sent while the marker is set.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Claude Code. The unregister carve-out came from a review comment on this PR; the API 23 constraint on `Long.hashCode` came from AnimalSniffer failing CI.

https://claude.ai/code/session_019y6WokBZBR2Zs5avoSTKVj


posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
```diff
@@ -226,6 +244,10 @@ internal class PostHogPushSubscriptionManager(
                 }
             }
```
marandaneto, line 239: blocking: Enforce rejection across all registration paths — The rejection guard only covers register() and retryPending(). If a previously successful registration subsequently receives 401 invalid_api_key (for example, after the project is deleted), its old delivered identity remains stored. A later identify() calls resendIfDistinctIdChanged(), clears halted, and sends through attempt() without checking the rejection marker, so identity changes continue posting for the rejected key. Please enforce the marker in the shared send path. Reproduction: reproduced — the focused test `review rejected key blocks identity resends` expected two POSTs but observed three; adding a temporary rejection guard to attempt() made it pass.

marandaneto (commented):
Automated advisory code review.

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
dmarchuk: Thank you! Fixed now 🙏

posthog/src/main/java/com/posthog/internal/PostHogPushSubscriptionManager.kt
```diff
@@ -729,6 +762,54 @@ internal class PostHogPushSubscriptionManager(
         return pendingRecord
     }
 
+    private fun isTokenRejected(): Boolean {
+        val rejectedAt = readRejections()[config.apiKey]?.toLongOrNull() ?: return false
+        if (System.currentTimeMillis() - rejectedAt < REJECTED_REPROBE_MILLIS) {
```
marandaneto: use `dateProvider: PostHogDateProvider` so we can mock and test this

posthog/src/main/java/com/posthog/internal/PostHogApiError.kt
```diff
@@ -8,13 +8,15 @@ import okhttp3.ResponseBody
...
  * @property message the OkHttp response body, the source might be closed already
+ * @property errorCode the `code` field of a PostHog JSON error body, when the response carried one
  */
 @PostHogInternal
 public class PostHogApiError(
     public val statusCode: Int,
     override val message: String,
     public val body: ResponseBody?,
     public val retryAfterSeconds: Int? = null,
+    public val errorCode: String? = null,
```
marandaneto, line 19: isnt that the same as `statusCode`?

marandaneto: oh nvm this is the encoded string and not the http status code

marandaneto (approved):
left a suggestion otherwise lgtm, not a blocker

PR #794, opened 2026-09-16
fix(push): stop sending the unused platform field

dmarchuk:
## Problem

Android devices send a `platform` field when registering for push, and nothing reads it. The API resolves the provider from `app_id` alone, and `platform` is neither validated nor stored.

It is also the only field in the request class without `@SerializedName`, so R8 renames it in minified release builds. Before the API stopped requiring it, that meant minified builds registered once, got a 400, and retried forever without ever succeeding.

## Changes

- Devices no longer send `platform`. The registration body is `api_key`, `distinct_id`, `device_token`, `app_id`, and `identity_token` when identity verification is on.
- The field is gone from the request class, from `PostHogApi.pushSubscription` and `pushUnsubscription`, and from the persisted pending records, which carried it only to replay it.
- `PostHogPushSubscriptionRequest` gains an R8 keep rule alongside the other wire classes, so a field added later is protected without needing an annotation to be remembered.
- Deleting the field removes the R8 rename failure mode rather than pinning the name against it.

Old pending files on disk still carry a `platform` key. Gson ignores unknown keys, so a registration or unregister queued before the upgrade still loads and replays, covered by two tests.

This supersedes #786, which pinned the wire name instead of removing the field.

- `./gradlew :posthog:cleanTest :posthog:test` passes, a full re-run rather than a cached one, since formatting touched the test sources.
- `spotlessCheck`, `detekt` and `:posthog:apiCheck` pass.
- The API dump is regenerated. The only change is the two `PostHogApi` push methods losing a `String` parameter. `PostHogPushSubscriptionManager` is internal and absent from the dump, and the public `registerPushNotificationToken(String, String)` never took a platform argument, so no public signature changes.
- The registration test now asserts the body has no `platform` key, rather than asserting its value.
- Two tests write a pending file in the old on-disk format, with the `platform` key, and assert it still replays. Both fail if that file is rejected.

Not tested: a real device against a real project. The change removes a field the server ignores, so there is no behavior to observe beyond the request body, which the tests assert directly.

## 🤖 Agent context

Claude Code (Opus 5). Written after checking that the endpoint requires only `distinct_id`, `device_token` and `app_id`, and that nothing downstream consumes the registration `platform`. The send-path `platform` metrics derive from the integration kind, not from this payload.

https://claude.ai/code/session_019y6WokBZBR2Zs5avoSTKVj


marandaneto (approved):
The code looks good; no qualifying findings. Verdict: correct.

PR #795, opened 2026-09-16
fix(server): keep local evaluation ETag in step with the flag definition cache

phillram:

From https://us.posthog.com/project/2/support/tickets/72623

A `posthog-server` instance that shares flag definitions through a `PostHogFlagDefinitionCacheProvider`, as described in the [distributed environments docs](https://posthog.com/docs/feature-flags/local-evaluation/distributed-environments), can get stuck serving stale definitions for good. Present in 2.17.0 and on `main`.

`loadFeatureFlagDefinitions()` advanced the ETag before it handed the definitions to the provider, and a failed write was only logged. One failed cache write then put that instance on 304s indefinitely: it had nothing left to publish, the shared cache kept the old definitions, and every follower instance served them until a flag in the project changed. A flag change that disables a feature never reached those instances. The same shape hits an evicted or flushed cache key, and a store that is unreachable for a single poll.

Three changes, no change to the `PostHogFlagDefinitionCacheProvider` contract:

- When a cache provider is configured, the ETag moves forward only if the definitions reached the cache. A failed, timed-out, or unserializable write resets it, so the next poll refetches the full set and retries the write. With no provider the ETag behaves exactly as before.
- An instance that fell back to the API because the cache was empty no longer keeps the ETag. It published nothing, so a conditional request would stop it from ever seeding the cache once it becomes the fetching instance.
- An instance that fetches definitions reads the cache on a 304 and rewrites what it already holds when the key is gone. This is the only one of the three that costs anything in the healthy path: one extra cache read per poll interval, on the fetching instance only. Flagging it for reviewers.

Three tests in `PostHogFeatureFlagsTest`, one per failure mode: a failed write retried on the next poll without `If-None-Match`, a cold API fallback that seeds the cache once it fetches, and a republish when the cache key disappears on a 304. All three fail on `main` and pass here.

The module's `TestFlagDefinitionCacheProvider` now keeps what it was given, so it behaves like a real cache instead of discarding every write. That is what makes the existing "does not store definitions on 304" test meaningful — it now asserts no redundant write against a populated cache.

`./gradlew :posthog-server:test :posthog-server:apiCheck spotlessCheck` is green (83 tests). `apiCheck` confirms the public API is unchanged. Not verified against a real distributed cache deployment.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Written by Claude Opus 5 in PostHog Desktop, from a support report about stale definitions behind a shared cache. Each failure mode was reproduced against `main` before anything changed.

Design notes for reviewers:

- Considered and rejected: a richer `shouldFetchFlagDefinitions` result, so a provider can ask for an unconditional fetch. It repairs the evicted-key case too, but it is a contract change every existing provider implementation would have to adopt. Reading the cache on a 304 buys the same repair for one read per poll.
- Deferred: attaching a server-side revision to the definitions so concurrent writes can be ordered without wall clocks. That is a separate contract change, and the write ordering it addresses is not what breaks here.
- `onFlagDefinitionsReceived` timing out counts as a failed write. A redundant republish is idempotent, and that beats assuming a write landed when it may not have.
- The ETag reset is scoped to instances that actually have a cache provider. Otherwise an SDK with no provider that hit a serialization error would refetch the full payload on every poll, forever.

---
*Created with [PostHog Desktop](https://posthog.com/desktop?ref=pr)*


marandaneto (approved):
The code looks good; no qualifying findings. Verdict: correct.

marandaneto (approved):

PR #797, opened 2026-09-17
fix(android): track app versions when lifecycle capture is disabled

marandaneto:

If lifecycle capture was disabled on earlier launches, the SDK did not remember the app version or build. Enabling capture later could emit a false `Application Installed` event.

Record version and build independently of lifecycle event emission. Compare the saved previous values before recording the current launch. Preserve consent, unavailable-storage and integration-ownership behavior. This cannot recover history that an older SDK never stored.

- Added regressions for disabled capture, same-version reenable, upgrades, consent and unavailable storage.
- Full Android tests and formatting passed.
- Tested on an Android 16 emulator against US Cloud, including a data-preserving APK upgrade from build 1 to build 2. Verified the update event and previous version/build, same-build restart, and clean install. Uploads returned HTTP 200.
- Automated branch review found no blocking issues.

## :pencil: Checklist

### If releasing new changes

- [x] Added an Android patch changeset.

## 🤖 Agent context

Implemented with Pi using repository tools and delegated implementation. Validated with Gradle, adb and an isolated automated review. Direct Boot still defers version accounting until storage is readable. No new unlock callback or consent policy was introduced. Human review is required.


posthog-android/src/main/java/com/posthog/android/PostHogAndroid.kt
```diff
@@ -235,9 +235,8 @@ public class PostHogAndroid private constructor() {
                     )
                 }
             }
-            if (config.captureApplicationLifecycleEvents) {
-                config.addIntegration(PostHogAppInstallIntegration(context, config, packageInfoProvider))
-            }
+            // Version accounting also runs when lifecycle event capture is disabled.
+            config.addIntegration(PostHogAppInstallIntegration(context, config, packageInfoProvider))
```
dustinbyrne, line 239: **Important — disabled clients claim install/update ownership**

Registering this integration unconditionally lets a lifecycle-disabled client claim its process-wide latch. If that client remains alive, a later enabled client for another project loses install/update events and does not update its markers; previously the disabled client did not register this integration.

Consider separating version bookkeeping from event ownership and covering disabled-first → enabled-second initialization through `PostHogAndroid.with`.

dustinbyrne (commented):
Reviewed the pinned head using source inspection and existing CI. The inline comments distinguish important findings, minor feedback, and decisions. No new tests or runtime reproductions were executed.

posthog-android/src/main/java/com/posthog/android/PostHogAndroid.kt
marandaneto: Fixed in e3e905c5. Reproduced through PostHogAndroid.with with a lifecycle-disabled client kept alive before an enabled client for another project: both install and update regressions failed because no event was captured. They now pass. Disabled clients record version/build without claiming the event latch; enabled-owner exclusion and non-owner uninstall protection remain tested. Focused debug/release tests, formatting, and the full Android debug suite passed (the full suite required an unchanged retry after setup/lifecycle test failures).

posthog-android/src/main/java/com/posthog/android/internal/PostHogAppInstallIntegration.kt
```diff
@@ -33,10 +33,12 @@ internal class PostHogAppInstallIntegration(
         if (config.cachePreferences?.isAvailable() == false) {
             return
         }
-        if (!integrationInstalled.compareAndSet(false, true)) {
-            return
+        if (config.captureApplicationLifecycleEvents) {
+            if (!integrationInstalled.compareAndSet(false, true)) {
```
turnipdabeets: question: when another client already owns installation, a second client with capture on returns here before saving its version (line 53). Versions are saved per project, so project B, which is always created second, never gets its version saved. The first launch where B is created first then sends a false `Application Installed` or a stale previous version, which is the same bug this PR fixes. Could we save the version before this check and use ownership only to decide whether the event is sent? The `enabled owner prevents secondary version writes` test currently asserts the opposite, so I'm curious whether that was on purpose.

.changeset/tidy-app-version-accounting.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-android": patch
+---
+
+Remember the app version and build even when application lifecycle capture is disabled, so enabling it on a later launch reports updates with the correct previous version instead of a false Application Installed event.
```
turnipdabeets: nit: this also changes what happens when the build number stays the same: the version name is now saved even without an update event. Maybe mention it?

posthog-android/src/test/java/com/posthog/android/PostHogAndroidAppVersionTest.kt
```diff
@@ -0,0 +1,214 @@
...
+        assertDisabledFirstDoesNotClaimEvents(false)
+    }
+
+    @Test
+    fun `disabled client does not claim update events from another project`() {
+        assertDisabledFirstDoesNotClaimEvents(true)
+    }
+
+    @Suppress("DEPRECATION")
+    private fun assertDisabledFirstDoesNotClaimEvents(upgrade: Boolean) {
```
turnipdabeets, line 86: nit, optional: this repeats most of the setup in `launch()`. It could probably take a second config instead.

turnipdabeets (approved):
LGTM, one question and a couple of nits inline. The question isn't a regression. The old code had the same early return, so it can go in a follow-up if you'd rather.

posthog-android/src/main/java/com/posthog/android/internal/PostHogAppInstallIntegration.kt
marandaneto: Agreed. The old early return was preserved intentionally, but it unnecessarily coupled per-project bookkeeping to process-wide event ownership. Fixed in 056a2647: every project records its version/build before the ownership check; only the owner can emit lifecycle events. Added two regressions through PostHogAndroid.with: a secondary project becoming first at the same version previously emitted a false install, and one becoming first after an upgrade reported a stale previous version. Both failed before the change and now pass. Updated the ownership test to retain event-exclusion and non-owner-uninstall coverage without blocking marker writes. Full Android debug tests, focused release tests, and formatting pass.

.changeset/tidy-app-version-accounting.md
marandaneto: Updated the one-line changeset in 056a2647 to explicitly include version-name changes with an unchanged build, alongside per-project accounting when lifecycle events are disabled or suppressed.

posthog-android/src/test/java/com/posthog/android/PostHogAndroidAppVersionTest.kt
marandaneto: Extracted shared prepareContext/createConfig helpers in 056a2647, reused by launch() and the multi-client scenarios. Existing coverage remains, with the two new project-ordering regressions added. Debug/release tests and formatting pass.

PR #798, opened 2026-09-17
feat(android): add semantic interaction autocapture

marandaneto:

Add default-off `captureElementInteractions` to `PostHogAndroidConfig` for actionable View and Compose taps without requiring session replay.

Emit `$autocapture` with the mobile `touch` event type, window-local coordinates in dp, and browser-compatible element properties. Only types, resource IDs/test tags and hierarchy positions are included. No text, input values, content descriptions or arbitrary attributes are sent.

Add a keyed View tag and `Modifier.postHogAutocaptureNoCapture()` for subtree exclusion. Respect explicit replay masking. Keep existing constructor signatures and optional Compose support.

The implementation observes Activity and Dialog windows. Windowless popups and nested Compose hosts are skipped. Geometric hit testing does not cover every custom drawing order, shape or expanded Compose touch region. These limits are documented in [the documentation comment](https://github.com/PostHog/posthog-android/pull/798#issuecomment-5711178719).

- Added target, input-transparency, exclusion, lifecycle, serialization and real Compose/native interop tests.
- Full Android debug/release suites, no-Compose tests, API generation and formatting passed during implementation. Re-ran focused tests after the final nested-host guard.
- Tested View/Compose controls on an Android 16 emulator with replay disabled against US Cloud. Allowed controls emitted four events, ignored controls emitted none, app handlers still ran, and uploads returned HTTP 200.
- Automated review led to fixes for hidden-node exclusions, overlapping interop targets, native ancestor targets and nested-host exclusion safety.
- Rejected the final review claim that unmerged `SemanticsNode.children` hides descendants beneath `clearAndSetSemantics`. The dependency source explicitly includes those descendants when merging is disabled. A real Compose probe confirmed the excluded child still handled input without emitting an event.

## :pencil: Checklist

### If releasing new changes

- [x] Added an Android minor changeset.

## 🤖 Agent context

Implemented with Pi using repository tools, delegated implementation, Gradle, adb and isolated automated review. The human-approved API uses a direct sibling configuration option rather than a wrapper object. Event properties follow existing PostHog browser/mobile conventions while excluding UI content.

A parent review reproduced and fixed a queued-install/close race. Later review fixes use conservative exclusion checks and match native interop to its winning Compose layout. No new runtime dependency was added. Human review is required.


posthog-android/src/main/java/com/posthog/android/internal/InteractionTarget.kt
```diff
@@ -0,0 +1,210 @@
...
+                }
+            }
+            return if (inside) listOf(view) else null
+        }
+        val path = hit(root, screenX - location[0], screenY - location[1], 0) ?: return null
+        if (incomplete) return null
+        val targetIndex = path.indexOfLast { it.isClickable && it.isEnabled }
+
+        fun nativeTarget(): InteractionTarget? {
+            if (targetIndex < 0 || path.any { it.isInteractionIgnored() || (it.isClickable && !it.isEnabled) }) return null
```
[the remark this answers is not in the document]
marandaneto: Fixed in 933284d3fa4ad48a2cadbb200335c4644cf71895. Reproduced with InteractionTargetTest (enabled native child resolved to null) and InteractionActivityTest (the Compose child handled the tap, but its event was dropped). Both failed before the change and pass now. Disabled clickable checks no longer reject ancestors above the selected native target or Compose host; subtree ignore/replay-mask exclusions remain intact. Added regression coverage for disabled child fallback and preserved exclusions. Validation: focused SDK/sample tests, make test, make checkFormat, and git diff --check all passed.

docs/element-interactions.md
```diff
@@ -0,0 +1,121 @@
+# Android element interaction capture
```
marandaneto: will delete after review

marandaneto:
# Android element interaction capture

Enable before SDK setup (off by default):

```kotlin
val config = PostHogAndroidConfig(apiKey).apply {
    captureElementInteractions = true
    // Independent of replay; replay can remain disabled.
    sessionReplay = false
}
PostHogAndroid.setup(application, config)
```

A single-pointer tap on an enabled, clickable Android View or Compose semantics `OnClick`
target produces `$autocapture`. Scrolls, drags beyond touch slop, long presses, cancelled
sequences and multi-touch do not. Input is observed, never consumed or redispatched. This
captures the tap, not a guarantee that the app's click handler succeeded. Target properties
are collected before the handler can change or navigate away from the screen.

Existing `optOut`, `beforeSend`, `close`, `$session_id` and `$screen_name` behavior applies.
The first enabled SDK instance owns the interaction integration. Closing a secondary instance
does not remove another instance's observers. There is no new start/stop API; configure at setup.

## Privacy and subtree exclusions

Windows marked `FLAG_SECURE` do not emit interaction events. The flag is checked on each
touch event, and tracking is reset while it is set, including changes during a gesture.

Only element types/roles, resource entry IDs or Compose test tags, sibling positions and
window-local touch coordinates in dp are sent. No text, content descriptions, input values,
arbitrary View tags, URLs or object hashes are serialized. In particular, v1 does **not** send `$el_text`. Use stable,
non-sensitive test tags: tags are identifiers, not a place for user data.

Exclude a View and all its descendants using the SDK's **qualified resource ID**:

```kotlin
container.setTag(com.posthog.android.R.id.posthog_autocapture_no_capture, true)
// Restore subsequent interactions:
container.setTag(com.posthog.android.R.id.posthog_autocapture_no_capture, false)
```

For Compose:

```kotlin
import com.posthog.android.postHogAutocaptureNoCapture

Column(Modifier.postHogAutocaptureNoCapture(isEnabled = true)) {
    Button(onClick = { /* app action */ }) { Text("Continue") }
}
```

Exclusions are re-evaluated on subsequent taps, including changes during a gesture. Compose
exclusions also apply to embedded `AndroidView` targets. The modifier does not change replay
masking. Existing explicit replay masking (`ph-no-capture` View tag/content-description marker,
`Modifier.postHogMask()`) excludes the interaction subtree as well; replay unmasking does not
override an interaction exclusion. Compose password semantics are excluded. Global replay
text/image masking does not disable all interaction capture, because no text/images are sent.

## Event mapping

`$event_type = "touch"` and `$touch_x` / `$touch_y` follow the mobile event convention.
Coordinates describe the completed tap in window-local density-independent units (Android dp,
matching iOS points), not screen pixels. They are snapshotted before the app handles the tap.
This follows posthog-ios commit `143d8337e9d39156f00fb5bde6c191a57febef1e`
(`PostHogRageClickIntegration.swift` / `PostHogSDK.swift`).

`$ce_version = 1`, `$elements` and `$elements_chain` preserve the browser structural schema.
Both element representations are sent for compatibility with downstream consumers.
Elements are target-first, then ancestors:

- `tag_name`: lowercase native View simple class name; Compose role (`button`, `checkbox`,
  `switch`, `radio`, `tab`, `image`), or `compose` when no supported role is supplied.
- `attr__id`: native resource **entry** name (not numeric/generated View IDs), or Compose test tag.
- `nth_child`, `nth_of_type`: one-based positions among native/semantic siblings.

The chain follows posthog-js commit `e96852dbe48690a18e40afe4bb423afb260a28d4`
(`autocapture-utils.ts`, `extractElements` / `elementsToString`): lexical attribute ordering,
quote escaping, target-first semicolon-separated elements. `attr__id` is also represented as
`attr_id` in the chain, matching that implementation. No fake DOM/page context is added.

Identifiers/types are limited to 128 UTF-16 code units, output to 20 elements, and hit traversal
to 512 nodes / 32 levels per native or Compose tree. Incomplete/unsupported traversal skips
the interaction rather than emitting a potentially wrong or excluded target. Native interop
uses its native ancestor chain; Compose-only targets include semantic and host View ancestors.

## Coverage and limitations

- Works without replay and without the optional Compose runtime. Uses the existing Curtains
  dependency to observe current and future Activity/Dialog windows, including deferred setup.
- Resolves nearest actionable ancestors, ordinary sibling/elevation overlap, native scroll and
  View matrices, rectangular clipping, and Compose's clipped semantic bounds.
- Only main-thread, touch-driven interactions with recognizable actionable targets are supported.
  Accessibility/keyboard/programmatic clicks, raw Canvas/custom gesture targets without click
  semantics, WebView DOM elements, and roots without a Curtains `phoneWindow` (`PopupWindow`,
  Compose `PopupLayout`, raw `WindowManager` overlays) are not captured. Activity and Dialog
  windows are supported; this is not all-window coverage. Unsupported roots never fall back to
  a target in an underlying window. Custom native child drawing order, nonrectangular clipping,
  and app-specific input interception can differ from geometric hit testing. Compose's expanded
  minimum touch-target regions outside semantic bounds are not inferred.
- Nested Compose hosts (a `ComposeView` inside an `AndroidView` in another Compose host) are
  skipped so exclusions in an enclosing semantic tree cannot be bypassed.
- Hierarchies beyond the safety bounds and incompatible Compose implementations are skipped.
  Class names can change with app obfuscation; resource IDs/test tags are preferable identifiers.
- Rage/dead-click detection is not part of this option.

## Sample and checks

The Android sample's **Element interactions** screen contains responsive, no-op and ignored
View/Compose buttons with static identifiers. The sample opts in through `MyApp`; to verify
independence, set `sessionReplay = false`. Supply your own local SDK credentials using the
sample's existing setup; no new credential is needed in this screen.

Focused checks:

```shell
./gradlew :posthog-android:testDebugUnitTest --tests '*Interaction*Test'
make test
make checkFormat
make api
```

The SDK test uses real Compose and embedded Views, dispatches through the Window, and
intercepts events with `beforeSend` rather than sending them to a service.

marandaneto:
Screenshot 2026-09-17 at 10 15 37

posthog-android/src/main/java/com/posthog/android/PostHogAutocaptureModifier.kt
```diff
@@ -0,0 +1,19 @@
+package com.posthog.android
+
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.SemanticsPropertyKey
+import androidx.compose.ui.semantics.semantics
+
+/** Compose exclusions for automatic element interaction capture. */
+public object PostHogAutocaptureModifier {
```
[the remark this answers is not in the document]
marandaneto: Fixed in 1bd5e8c94b734b4d23b6f14467b1040bb37740e6. Moved postHogAutocaptureNoCapture to a top-level extension in com.posthog.android, following the repository preference. Updated SDK/sample imports, generated the API dump, and updated the documentation comment. Full Android tests and formatting passed.

posthog-android/src/main/java/com/posthog/android/internal/PostHogElementInteractionIntegration.kt
```diff
@@ -0,0 +1,120 @@
...
+        if (interceptors.containsKey(window)) return
+        val weakRoot = WeakReference(root)
+        val tracker = InteractionTapTracker()
+        val interceptor =
+            TouchEventInterceptor { event, dispatch ->
+                safely {
+                    if (Looper.myLooper() != mainHandler.mainLooper) return@safely
+                    val client = postHog
+                    val view = weakRoot.get()
+                    if (client == null || client.isOptOut() || view == null || !config.captureElementInteractions) {
```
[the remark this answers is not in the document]
marandaneto: Fixed in 1bd5e8c94b734b4d23b6f14467b1040bb37740e6. Reproduced with two new PostHogElementInteractionIntegrationTest regressions: secure-window taps emitted an event, and a flag set during MOVE then cleared before UP did not cancel the gesture. Both failed before the fix and pass now. We check FLAG_SECURE on every observed touch and reset tracking while secure, preserving input dispatch and weak window ownership. Full Android tests and formatting passed. Reproduction/verification command: ./gradlew :posthog-android:testDebugUnitTest --tests "*PostHogElementInteractionIntegrationTest".

posthog-android/src/main/java/com/posthog/android/internal/ComposeInteractionTargetResolver.kt
```diff
@@ -0,0 +1,111 @@
...
+                    it.config.getOrNull(PostHogReplayMask) == true ||
+                    it.config.contains(SemanticsProperties.Password)
+            }
+        ) {
+            return null
+        }
+        if (nativeLayoutInfo != null && path.any { it.layoutInfo === nativeLayoutInfo }) return nativeTarget
+        val nativeAncestor = nativeTarget?.view?.get()?.let { candidate -> viewPath.any { it === candidate } } == true
+        if (nativeTarget != null && !nativeAncestor && nativeLayoutInfo == null) return null
+        if (path.any { it.config.contains(SemanticsActions.OnClick) && it.config.contains(SemanticsProperties.Disabled) }) return null
```
dustinbyrne: **Important — enabled children inside disabled clickable ancestors**

This full-path check rejects an enabled Button inside a disabled clickable Box, even though the child can handle the tap. That suppresses a valid interaction.

Consider resolving the actionable target before applying disabled-state checks, while retaining whole-path privacy exclusions. An enabled-child/disabled-parent case at both resolver and Window-dispatch level would cover this boundary.

posthog-android/src/main/java/com/posthog/android/PostHogAndroidConfig.kt
```diff
@@ -65,5 +65,15 @@ public open class PostHogAndroidConfig
                 }
             },
         ) {
+        /**
+         * Captures `$autocapture` for actionable View and Compose taps. Default: false.
+         * Configure before setup. Independent of session replay; sends types and resource IDs/test
+         * tags and window-local touch coordinates in dp, never text, content descriptions or input
+         * values. Use only non-sensitive test tags.
+         * Exclude a View subtree with `com.posthog.android.R.id.posthog_autocapture_no_capture` set to
+         * true, or use [PostHogAutocaptureModifier.postHogAutocaptureNoCapture] for Compose.
```
dustinbyrne: **Minor — unresolved KDoc link**

`PostHogAutocaptureModifier` is not a source object/class. Consider linking to the top-level `postHogAutocaptureNoCapture` function instead.

posthog-android/src/main/java/com/posthog/android/internal/InteractionTarget.kt
```diff
@@ -0,0 +1,212 @@
...
+                )
+            element["attr__id"]?.let {
+                attributes["attr__id"] = it
+                attributes["attr_id"] = it
+            }
+            element.getValue("tag_name").toString() + ":" +
+                attributes.entries.joinToString("") { (key, value) -> "$key=\"${escape(value)}\"" }
+        }
+    return mapOf(
+        "\$event_type" to "touch",
```
dustinbyrne, line 55: **Needs decision — mobile event type**

This emits `$event_type=touch`, while the canonical autocapture scenario expects `click`. Is `touch` the intended mobile contract? Consider confirming that interpretation before selecting a payload or specification change; this is a contract question rather than a recommendation to switch the value automatically.

dustinbyrne (commented):
Reviewed the pinned head using source inspection and existing CI. The inline comments distinguish important findings, minor feedback, and decisions. No new tests or runtime reproductions were executed.

posthog-android/src/main/java/com/posthog/android/internal/ComposeInteractionTargetResolver.kt
marandaneto: Fixed in 48b49cbf. Reproduced in ComposeInteractionTargetResolverTest and the real Window-dispatch InteractionActivityTest: the enabled child handled the tap, but capture was missing. Both now pass. We select the nearest clickable node before checking its disabled state; a disabled child still blocks ancestor fallback, and whole-path privacy exclusions remain enforced. Full Android debug tests, focused release interaction tests, and formatting passed.

posthog-android/src/main/java/com/posthog/android/PostHogAndroidConfig.kt
marandaneto: Fixed in 48b49cbf: the KDoc now links directly to the top-level postHogAutocaptureNoCapture function. Verified the declaration and link by inspection; compilation and formatting checks passed.

posthog-android/src/main/java/com/posthog/android/internal/InteractionTarget.kt
marandaneto: `$event_type = "touch"` is intentional for Android. We selected the mobile interaction convention, following the iOS rage-click precedent rather than the browser `click` value. The iOS implementation passes `EventType.kTouch`, and its event test asserts `$event_type == "touch"`: https://github.com/PostHog/posthog-ios/blob/143d8337e9d39156f00fb5bde6c191a57febef1e/PostHog/Autocapture/PostHogRageClickIntegration.swift . Android payload tests also assert `touch`. We are keeping that mobile contract; no payload or specification change is included here.

dustinbyrne, line 55: **Minor / contract follow-up:** The intentional mobile `touch` value is understood. The canonical eligible-autocapture scenario still says `click`; consider updating the specification to reflect the mobile convention. This is non-blocking and is not a request to change the Android payload.

dustinbyrne (approved):
Looks good with the non-blocking inline note. Source-reviewed with AI assistance and existing CI evidence; no new tests executed.

posthog-android/src/main/java/com/posthog/android/internal/InteractionTarget.kt
marandaneto: https://github.com/PostHog/sdk-specs/pull/69

